PR #15, opened 2019-10-09
gRPC + Pub/Sub API

### LINKED ISSUE #5: Network API
Current setup is HTTP + a libp2p adapter. However, since many payloads contain raw bytes, having a serialization layer that can handle typed bytes (protos) would be ideal.

I'm considering gRPC + libp2 transport (I have an example of this working nicely). Pubsub is another consideration. Also considering gRPC.

sanderpick:
fixes #5 

This adds a gRPC service for log orchestration between peers. The service contains a pubsub implementation as well (gossipsub) which is used to broadcast new log records to listeners. Starting this service creates a listeners for each known thread.

IPLD nodes are CBOR. We could provide a higher level interface for working with nodes, and then provide implementations for the other node types: proto and raw. For now, I've just used CBOR but kept the logic somewhat separate which should make adding the others relatively straightforward. 

Also sneaks in CI setup.


cbor/invite.go
```diff
@@ -69,7 +63,7 @@ func NewInvite(logs []thread.LogInfo, readable bool) (format.Node, error) {
 	}, mh.SHA2_256, -1)
 }
 
-func DecodeInvite(node format.Node) ([]thread.LogInfo, bool, error) {
+func InviteFromNode(node format.Node) ([]thread.LogInfo, bool, error) {
```
sanderpick: Here we attempt to decode a node into an invite.

cbor/record.go
```diff
@@ -0,0 +1,206 @@
...
+	cbornode.RegisterCborType(record{})
+}
+
+type record struct {
+	Block cid.Cid
+	Sig   []byte
+	Prev  cid.Cid `refmt:",omitempty"`
+}
+
+func NewRecord(ctx context.Context, dag format.DAGService, block format.Node, prev cid.Cid, sk ic.PrivKey, key crypto.EncryptionKey) (thread.Record, error) {
```
sanderpick: Records are the top-most object in a log. The `block` is an event in our case, but I've left it generic so any node can be wrapped.

cbor/record.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	err = dag.AddMany(ctx, []format.Node{coded})
+	if err != nil {
+		return nil, err
+	}
+
+	return &Record{
+		Node:  coded,
+		obj:   obj,
```
sanderpick: The custom node's carry a cached obj which is the de-coded (and decrypted) version.

cbor/record.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Record{
+		Node: coded,
+		obj:  obj,
+	}, nil
+}
+
+func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record) (*pb.Record, error) {
```
sanderpick: When sending records to other peers, we serialized the node and all it's children (minus any links in the body). This means we don't have to worry about handling a bunch of pins... just send the whole thing.

pb/custom.go
```diff
@@ -132,3 +98,206 @@ func (a *ProtoAddr) UnmarshalJSON(data []byte) error {
 func (a ProtoAddr) Size() int {
 	return len(a.Bytes())
 }
+
+// ProtoCid is a custom type used by gogo to serde raw CIDs into the cid.CID type, and back.
```
sanderpick: More custom types...

pb/threads.proto
```diff
@@ -0,0 +1,68 @@
...
+    bytes node = 1;
+
+    bytes eventNode = 2;
+
+    bytes headerNode = 3;
+
+    bytes bodyNode = 4;
+}
+
+message PushRequest {
```
sanderpick: This is an outgoing new record envelope.

pb/threads.proto
```diff
@@ -0,0 +1,68 @@
...
+
+        bytes readKeyLogID = 5 [(gogoproto.customtype) = "ProtoPeerID"];
+    }
+}
+
+message PushReply {
+    bool ok = 1;
+}
+
+message PullRequest {
```
sanderpick: This is a request for new records.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		go s.subscribe(id)
+	}
+
+	// @todo: ts.pubsub.RegisterTopicValidator()
+
+	return s, nil
+}
+
+// Push receives a record request.
+func (s *service) Push(ctx context.Context, req *pb.PushRequest) (*pb.PushReply, error) {
```
sanderpick: This handler is called by a remote client and also our local pubsub listeners.

service.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if x {
+		return reply, nil
+	}
+
+	// Check if this log already exists
+	logpk := s.threads.PubKey(req.ThreadID.ID, req.LogID.ID)
+	if logpk == nil {
```
sanderpick: If the log does not exist, it _could_ be an invite.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		var kid peer.ID
+		if req.Header.ReadKeyLogID != nil {
+			kid = req.Header.ReadKeyLogID.ID
```
sanderpick: If this request indicates that the read key is from a different log, this means that this is an invite from a new peer that used the invitee's log (which we should have) to encrypt _their_ invite (which is effectively a join). I'm trying to avoid message types (like invite / join) in favor of just interpreting the context. Working so far.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+		lg, err := s.handleInvite(ctx, req.ThreadID.ID, req.LogID.ID, kid, rec)
+		if err != nil {
+			return nil, err
+		}
+
+		if lg != nil {
+			// Notify existing logs of our new log
```
sanderpick: We created a log as a new writer. Notify others of the new log by sending another invite.

service.go
```diff
@@ -0,0 +1,541 @@
...
+			invite, err := cbor.NewInvite([]thread.LogInfo{*lg}, true)
+			if err != nil {
+				return nil, err
+			}
+			_, _, err = s.threads.Add(
```
sanderpick: Add the invite record locally. This will generate an outgoing request _from this peer_.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+	} else {
+		// Verify node
+		err = rec.Verify(logpk)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	err = s.threads.Put(ctx, rec, tserv.PutOpt.Thread(req.ThreadID.ID), tserv.PutOpt.Log(req.LogID.ID))
```
sanderpick: Save the record locally.

service.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return reply, nil
+}
+
+func (s *service) Pull(ctx context.Context, req *pb.PullRequest) (*pb.PullReply, error) {
```
sanderpick: This is a pull handler called by remote clients who want to get new updates in a log that has been addressed to this peer. This could be the owner or a follower.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		pbrecs.Records[i], err = cbor.RecordToProto(ctx, s.threads.dagService, r)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return pbrecs, nil
+}
+
+// push a record to log addresses and thread topic.
+func (s *service) push(ctx context.Context, rec thread.Record, id thread.ID, lid peer.ID, settings *tserv.AddSettings) error {
```
sanderpick: This is the client-side push method. Here we gather the addresses at which to deliver the request.

service.go
```diff
@@ -0,0 +1,541 @@
...
+			}
+
+			log.Debugf("reply from %s: %t", p, reply.Ok)
+
+			wg.Done()
+		}(addr)
+	}
+
+	// Finally, publish to the thread's topic
+	err = s.publish(id, req)
```
sanderpick: Publish as well.

service.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		log.Error(err)
+	}
+
+	wg.Wait()
+	return nil
+}
+
+type records struct {
```
sanderpick: This thing just records unique records _in order_ as the come in from multiple go routines.

sanderpick: I considered sync.Map, but the `interface{}` overhead felt not worth it. Plus it seems that a map + RWMutex is more performant based on my minimal research :)

service.go
```diff
@@ -0,0 +1,541 @@
...
+	if _, ok := r.m[key]; ok {
+		return
+	}
+	r.m[key] = value
+	r.s = append(r.s, value)
+	r.Unlock()
+}
+
+// pull records from log addresses.
+func (s *service) pull(ctx context.Context, id thread.ID, lid peer.ID, offset cid.Cid, settings *tserv.PullSettings) ([]thread.Record, error) {
```
sanderpick: This is the client-side pull method. Internally, we need to gather all the addresses known for the given log.

service.go
```diff
@@ -0,0 +1,541 @@
...
+				return
+			}
+
+			for _, r := range reply.Records {
+				rec, err := cbor.RecordFromProto(r, fk)
+				if err != nil {
+					log.Error(err)
+					return
+				}
+				recs.Store(rec.Cid(), rec)
```
sanderpick: Store the value in our thread safe records object.

service.go
```diff
@@ -0,0 +1,541 @@
...
+			}
+
+			log.Debugf("reply from %s: %s", p, reply.String())
+
+			wg.Done()
+		}(addr)
+	}
+
+	wg.Wait()
```
sanderpick: As an enhancement, we should probably be emitting these records _live_ via a buffered channel. However, we can revisit when the event store is ready to integrate.

service.go
```diff
@@ -0,0 +1,541 @@
...
+}
+
+// dial attempts to open a GRPC connection over libp2p to a peer.
+func (s *service) dial(ctx context.Context, peerID peer.ID, dialOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
+	opts := append([]grpc.DialOption{s.getDialOption()}, dialOpts...)
+	return grpc.DialContext(ctx, peerID.Pretty(), opts...)
+}
+
+// getDialOption returns the WithDialer option to dial via libp2p.
+func (s *service) getDialOption() grpc.DialOption {
```
sanderpick: Here we proxy through to gostream as the dialler.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("grpc tried to dial non peer-id: %s", err)
+		}
+		c, err := gostream.Dial(ctx, s.threads.host, id, IPELProtocol)
+		return c, err
+	})
+}
+
+// subscribe to a thread for updates.
+func (s *service) subscribe(id thread.ID) {
```
sanderpick: Subscribe to pubsub updates for the given thread.

service.go
```diff
@@ -0,0 +1,541 @@
...
+func (s *service) publish(id thread.ID, req *pb.PushRequest) error {
+	data, err := req.Marshal()
+	if err != nil {
+		return err
+	}
+	return s.pubsub.Publish(id.String(), data)
+}
+
+// handleInvite attempts to process a log record as an invite event.
+func (s *service) handleInvite(ctx context.Context, id thread.ID, lid peer.ID, kid peer.ID, rec thread.Record) (*thread.LogInfo, error) {
```
sanderpick: As the doc string says, we try to process the record as an invite... adding the logs it contains if valid.

service.go
```diff
@@ -0,0 +1,541 @@
...
+			}
+		}
+		err = s.threads.AddLog(id, lg)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Create an own log if this is a new thread
+	var ownLog *thread.LogInfo
```
sanderpick: If the invite contained the read key, we just go ahead and create a log for ourselves which will be sent back.

service.go
```diff
@@ -0,0 +1,541 @@
...
+		err = s.threads.AddLog(id, lg)
+		if err != nil {
+			return nil, err
+		}
+		ownLog = &lg
+	}
+
+	// Subscribe to the new thread
+	if newThread {
+		go s.subscribe(id)
```
sanderpick: Finally, subscribe to this thread if it's new.

test/threads_suite.go
```diff
@@ -166,6 +183,28 @@ func testAddInvite(ts1, ts2 tserv.Threadservice) func(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
+
+		info := ts2.ThreadInfo(tid)
+		if len(info.Logs) != 2 {
+			t.Fatalf("expected 2 logs got %d", len(info.Logs))
+		}
+
+		for _, lid := range info.Logs {
```
sanderpick: Test the pull API... larger tests are forthcoming in other PRs.

threads.go
```diff
@@ -0,0 +1,357 @@
+package threads
```
sanderpick: I renamed this from threadservice.

threads.go
```diff
@@ -0,0 +1,357 @@
...
+		ctx:         ctx,
+		cancel:      cancel,
+		Threadstore: ts,
+	}
+	t.service, err = newService(t)
+	if err != nil {
+		return nil, err
+	}
+
+	listener, err := gostream.Listen(h, IPELProtocol)
```
sanderpick: gostream returns a `net.Listener` which wraps our host. This can be used just like any other listener!

threads.go
```diff
@@ -0,0 +1,357 @@
...
+
+func (t *threads) Host() host.Host {
+	return t.host
+}
+
+func (t *threads) DAGService() format.DAGService {
+	return t.dagService
+}
+
+func (t *threads) Add(ctx context.Context, body format.Node, opts ...tserv.AddOption) (l peer.ID, r thread.Record, err error) {
```
sanderpick: Top-level Add API for writing to a log.

threads.go
```diff
@@ -0,0 +1,357 @@
...
+	// Push out the new record
+	err = t.service.push(ctx, coded, settings.Thread, lg.ID, settings)
+	if err != nil {
+		return
+	}
+
+	return lg.ID, coded, nil
+}
+
+func (t *threads) Put(ctx context.Context, rec thread.Record, opts ...tserv.PutOption) error {
```
sanderpick: Top-level API for putting existing records into a log. I think this will be reworked into an internal method.

threads.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Update head
+	t.SetHead(settings.Thread, lg.ID, rec.Cid())
+	return nil
+}
+
+func (t *threads) Pull(ctx context.Context, id thread.ID, lid peer.ID, offset cid.Cid, opts ...tserv.PullOption) ([]thread.Record, error) {
```
sanderpick: Currently, this just pulls from the given log. However, we probably want to pull from the _whole_ thread. Also, we don't really care what the result is here... any new stuff should be delivered over some kind of channel.

threads.go
```diff
@@ -0,0 +1,357 @@
...
+func (t *threads) Logs(id thread.ID) []thread.LogInfo {
+	lgs := make([]thread.LogInfo, 0)
+	for _, lid := range t.ThreadInfo(id).Logs {
+		lg := t.LogInfo(id, lid)
+		lgs = append(lgs, lg)
+	}
+	return lgs
+}
+
+func (t *threads) Delete(ctx context.Context, id thread.ID) error {
```
sanderpick: TODO

threads.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	t.SetHead(settings.Thread, lg.ID, rec.Cid())
+
+	return rec, nil
+}
+
+func (t *threads) pullLocal(ctx context.Context, id thread.ID, lid peer.ID, offset cid.Cid, limit int) ([]thread.Record, error) {
```
sanderpick: This looks for new records ahead of the offset _locally_. This is called locally when requesting from a log addressed to this peer, and remotely via the RPC service when requesting as a remote client.

threads.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var recs []thread.Record
+	if limit == 0 {
+		return recs, nil
+	}
+	cursor := lg.Heads[0]
+	for { // @todo: Max depth
```
sanderpick: This is a bit dangerous at the moment. We have to look back from HEAD in order to collect new stuff - not ideal.

tstoreds/cyclic_batch.go
```diff
@@ -1,7 +1,7 @@
 package tstoreds
 
 import (
-	"github.com/pkg/errors"
+	"fmt"
```
sanderpick: Just replacing errors with the built in.

service.go
```diff
@@ -0,0 +1,541 @@
...
+
+	rec, err := recordFromProto(req.Record, fkey)
+	if err != nil {
+		return nil, err
+	}
+	x, err := s.threads.blocks.Blockstore().Has(rec.Cid())
+	if err != nil {
+		return nil, err
+	}
+	if x {
```
jsign: `s/x/knownRecord`?

service.go
```diff
@@ -0,0 +1,541 @@
...
+				tserv.AddOpt.Thread(req.ThreadID.ID),
+				tserv.AddOpt.KeyLog(req.LogID.ID))
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else {
+		// Verify node
+		err = rec.Verify(logpk)
+		if err != nil {
```
jsign: Change to `if err = ... ; err != nil` idiom?

util/util.go
```diff
@@ -11,7 +11,8 @@ import (
 	"github.com/textileio/go-textile-core/thread"
 )
 
-func CreateLog(p peer.ID) (info thread.LogInfo, err error) {
+// CreateLog creates a new log with the given peer as host.
+func CreateLog(host peer.ID) (info thread.LogInfo, err error) {
```
jsign: `peer.IDFrompublicKey` has a `deprecated` annotation.

threads.go
```diff
@@ -0,0 +1,384 @@
...
+		return nil, err
+	}
+
+	t.SetHead(settings.Thread, lg.ID, rec.Cid())
+
+	return rec, nil
+}
+
+// pullLocal returns local records from the given thread that are ahead of offset but not farther then limit.
+func (t *threads) pullLocal(ctx context.Context, id thread.ID, lid peer.ID, offset cid.Cid, limit int) ([]thread.Record, error) {
```
jsign: Check `limit` for `<0` values, and maybe put a cap in its max value to avoid massive pulls (maybe not erroring, but `max(maxPullSize, limit)` or similar (and check from the puller side that this case might happen (less than the requested limit)).

threads.go
```diff
@@ -0,0 +1,384 @@
...
+	fk, err := crypto.ParseDecryptionKey(lg.FollowKey)
+	if err != nil {
+		return nil, err
+	}
+
+	var recs []thread.Record
+	if limit == 0 {
+		return recs, nil
+	}
+	cursor := lg.Heads[0]
```
jsign: I remember that `len(lg.Heads)>0` is a future-idea, but maybe for the moment assert that `==1`?

threads.go
jsign: Agree!, mem usage is linear with depth! Could be made const in the worst-case if the `limit` is capped (maybe a circular-buffer?). Other O(1) mem are possible too, yep.

service.go
```diff
@@ -0,0 +1,541 @@
...
+			}
+			pid, err := peer.IDB58Decode(p)
+			if err != nil {
+				log.Error(err)
+				return
+			}
+
+			log.Debugf("pushing record to %s...", p)
+
+			cctx, _ := context.WithTimeout(ctx, reqTimeout)
```
jsign: `defer cancelFunc()`?

service.go
```diff
@@ -0,0 +1,541 @@
...
+			conn, err := s.dial(cctx, pid, grpc.WithInsecure())
+			if err != nil {
+				log.Error(err)
```
jsign: Maybe add extra info about the peer who timeout.

service.go
```diff
@@ -0,0 +1,541 @@
...
+				return
+			}
+			client := pb.NewThreadsClient(conn)
+			reply, err := client.Push(cctx, req)
+			if err != nil {
+				log.Error(err)
+				return
```
jsign: Warning! (both returns statements): `wg.Done()` isn't `defer`ed, so this will deadlock the parent-goroutine.

threads.go
```diff
@@ -0,0 +1,384 @@
...
+	info, err = t.createLog()
+	if err != nil {
+		return
+	}
+	err = t.AddLog(id, info)
+	return
+}
+
+// getOwnLoad returns the log owned by the host under the given thread.
+func (t *threads) getOwnLog(id thread.ID) (info thread.LogInfo) {
```
jsign: Can I own more than one log in a thread?, if that makes sense it returns the first.

cbor/record.go
andrewxhill: just to make sure i follow, what's a block node that isn't an event? how is it used?

threads.go
andrewxhill: A Record contains the Event? If the Event contains the sequence info, how will it rework the Event's sequencing method (i.e. the clock record) for the target Log here? Am I reading it right below that it just pulls the core bits of the Event (header/body) and wraps them in a new event for the target log?

cbor/record.go
sanderpick: it could be anything... up to the user. It's a little half-baked, but the idea is that somebody could use the bulk of this lib for something else that needs append-only single writer logs that aren't of the same shape as our events.

threads.go
sanderpick: nice yeah, a circular buffer would be worth shot. you really just want to keep a moving window of size `limit`. this is a drawback of the graph direction. another option could be to just return head and let the consumer do the reversal, but that means they'll have to look at the DHT to find the older nodes between offset and head. tricky.

cbor/event.go
```diff
@@ -107,10 +127,12 @@ type Event struct {
 	body   format.Node
 }
 
+// HeaderID returns the cid of the header.
 func (e *Event) HeaderID() cid.Cid {
 	return e.obj.Header
 }
 
+// GetHeader returns the header node.
 func (e *Event) GetHeader(ctx context.Context, dag format.DAGService, key crypto.DecryptionKey) (thread.EventHeader, error) {
```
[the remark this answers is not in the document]
sanderpick: ah yes, good idea. I will add.

cbor/invite.go
```diff
@@ -86,10 +85,6 @@ func DecodeInvite(node format.Node) ([]thread.LogInfo, bool, error) {
 		if err != nil {
 			return nil, false, err
 		}
```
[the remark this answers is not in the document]
sanderpick: +1, adding

cbor/invite.go
```diff
@@ -69,7 +67,8 @@ func NewInvite(logs []thread.LogInfo, readable bool) (format.Node, error) {
 	}, mh.SHA2_256, -1)
 }
 
-func DecodeInvite(node format.Node) ([]thread.LogInfo, bool, error) {
+// InviteFromNode returns invite info from a node.
+func InviteFromNode(node format.Node) ([]thread.LogInfo, bool, error) {
```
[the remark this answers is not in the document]
sanderpick: Good point. It doesn't need to live in the node. I'll play with returning a real type here that has a Readable method that can just traverse the logs. Really, readable should only be true if all logs are readable. Currently, it's just passing through the decoded value for readable.

cbor/record.go
```diff
@@ -0,0 +1,217 @@
...
+	Prev  cid.Cid `refmt:",omitempty"`
+}
+
+// NewRecord returns a new record from the given block and log private key.
+func NewRecord(ctx context.Context, dag format.DAGService, block format.Node, prev cid.Cid, sk ic.PrivKey, key crypto.EncryptionKey) (thread.Record, error) {
+	payload := block.RawData()
+	if prev.Defined() {
+		payload = append(payload, prev.Bytes()...)
+	}
+	sig, err := sk.Sign(payload)
```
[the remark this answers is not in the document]
sanderpick: Hmm, yeah. I suppose if you're not using `Event`s which do not have any data (only links), this could get nasty. `prev` is a `Cid`, so `Bytes` is just returns the byte representation of the Cid. How about `block.Cid().Bytes()+prev.Bytes()`?

service.go
sanderpick: Oh! good catch :)

threads.go
sanderpick: You cannot - or in anycase if you did that would be unexpected.

pb/threads.proto
```diff
@@ -0,0 +1,99 @@
...
+    repeated Record records = 1;
+}
+
+// Threads is the peer-to-peer network API for thread orchestration.
+service Threads {
+    // Push a record to a thread.
+    rpc Push(PushRequest) returns (PushReply) {}
+
+    // Pull records from a thread.
+    rpc Pull(PullRequest) returns (PullReply) {}
```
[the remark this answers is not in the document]
sanderpick: the service builds the request, or in other words, this is not public to the Threads consumer.

cbor/record.go
```diff
@@ -0,0 +1,217 @@
...
+	hnode, err := cbornode.Decode(rec.HeaderNode, mh.SHA2_256, -1)
+	if err != nil {
+		return nil, err
+	}
+	body, err := cbornode.Decode(rec.BodyNode, mh.SHA2_256, -1)
+	if err != nil {
+		return nil, err
+	}
+
+	decoded, err := DecodeBlock(node, key)
```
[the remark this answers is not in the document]
sanderpick: +1, I'll add an assertion in `RecordFromProto` - seems like using a nil key in `DecodeBlock` is a user error is panic worthy

util/util.go
sanderpick: oh, good catch - the wrong `peer` package is imported. will fix

sanderpick:
Great, @jsign. Thank you for the detailed review. I'll ping with changes.

cbor/record.go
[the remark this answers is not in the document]
jsign: LGTM

threads.go
sanderpick: Yep, events are within the record, which is the top level node. Clocks will have to be handled at the model +/ event store level. Here we just put the record, which falls into whatever place is described by its prev link.

threads.go
sanderpick: I simplified the logic in the method a little. We have an upper limit, which when reached just returns all found values. I added a note: 

```
// It is possible to reach limit before offset, meaning that the caller
// will be responsible for the remaining traversal.
```

The same method is used on the receiver side.

sanderpick: As per old logic, the default limit set in textile-core is -1. I have a change to the pull API in a different branch that removes the need for pull options. So, just leaving that default for now... the called needs to set the limit > 0(as in the tests). Otherwise, they just get no results. (-1 used to mean "all")

sanderpick:
Comments addressed, @jsign.

PR #16, opened 2019-10-09
threadstore: addrbook datastore backed implementation & more tests

### LINKED ISSUE #6: Datastore backed AddrBook
Make a ipfs go-datastore implementation for AddrBook

jsign:
Fixes #6

Turns out that `AddrBook` datastore implementation was quite massive but most of the heavy-lifting was based on the existing `peerstore` implementation. Most of the complexity lies in the TTL feature I guess.

More detailed changes here:
- Added a new test in the addrbook suite to test `LogsWithAddrs` function
- Added a `README.md` at `tstoreds` level to give some superficial description about it (also, give proper credit to plabs guys since pretty much all the code is inspired in them)
- There's a GC goroutine to handle TTLs (more on that below)
- In `tstoreds/Options.go` there's a no-brainer `DefaultOpts`, maybe should be tuned.
- I decided to also have a `pb` representation of the Addr for the datastore. Before that I experimented with doing it with `gob`, which was kind-of nice to avoid the proto generating thing and worked... but since `Multiaddress` implements some binary marshalers, it wasn't as clean as I'd liked. So sticked with the same `pb` as peerstore. 
- As mentioned in `tstoreds/addrbook.go`, the namespace in the datastore is `/thread/addrs/<b32 thread id no padding>/<b32 log id no padding>`, I guess makes sense... thought about something more like `/thread/<b32 thread id no padding>/addrs/<b32 log id no padding>`, that would imply doing some extra `.ChildString`. In this last option, other books would share the `/thread/<thread id>` prefix. Any opinions?
- All addrbook test suit pass with Badger and LevelDB

Some particular notes about the GC:
- The original GC from `peerstore` had two flavors. I only included the default one (`purgeStore`) since I prefer not adding code that I don't understand quite well (just a matter of reading time). Maybe later will take a look with more time and think about if helps including it (would be basically copy paste with some tweaks)
- I read this implementation and is basically the usual _scan all keys, check for ttl expiring, and clean them up_, but...
- After reading it, I have some concerns:
   - I don't see some kind of _stop the world_ guarantee when the GC runs. GCed elements are evicted from the cache without using the corresponding lock from the entry (as used in `addr_book.go`). Feels like a race.
  - Cache is evicted instantly but flushing to the datastore is used with a `cyclicBatching` implementation on top of the datastore to commit every 20 writes (or so). [Batching doesn't have _serializable_ guarantee](https://github.com/ipfs/go-datastore/blob/master/datastore.go#L91), so maybe will override other writes being done by the addr_book user?
  - ... some other cases mostly a result of not stoping the world when gc.

These GC notes are quite surprising, maybe seconds eyes would be nice to see if this is confirmed and I missed something.


jsign:
> Very nice @jsign! I have some conflicting proto work (https://github.com/textileio/go-textile-threads/pull/15/files#diff-277fb85ce5a7d7fda561c92d758b04db). The network API needs similar serialization with the addition of other custom types (`thread.ID`, `cid.Cid`). In any case, I can rebase if this goes in first.
> 
> Regarding GC and key shape, it might be useful to have a chat.

Sure!

One extra thing that I forgot to mention, is that I tweaked the base implementation panicking in  [AddAddrs](https://github.com/textileio/go-textile-threads/pull/16/files#diff-5dca656055099aed5e4c9dbc0cddd449R116) and [SetAddrs](https://github.com/textileio/go-textile-threads/pull/16/files#diff-5dca656055099aed5e4c9dbc0cddd449R131) since those APIs don't leave room for `error`. Not sure why they fail silently, maybe makes sense?

PR #21, opened 2019-10-14
Datastore backed Key, Head, and Metadata books

### LINKED ISSUE #7: Datastore backed HeadBook
Make a ipfs go-datastore implementation for HeadBook

### LINKED ISSUE #8: Datastore backed KeyBook
Make a ipfs go-datastore implementation for Keybook

### LINKED ISSUE #9: Datastore ThreadMetadata for Threadstore
Use ipfs go-datastore implementation for ThreadMetadata in Threadstore

jsign:
Fixes #7 
Fixes #8 
Fixes #9 

This is a datastore backed implementations of Keys, Heads and Metadata books for `Threadstore`.

This already _assumes_ a merged [Draft PR in `go-textile-core`](https://github.com/textileio/go-textile-core/pull/13) which will be possibly merged after merging #15 . This consideration (changed API defs) resulted in changing the `tstoremem` implementation, and tests.

After #15 gets merged, I still need to adjust some more things from that new code that will appear when rebasing.

To recap: The expected ordered flow of merges would be: #15, [`go-textile-core` PR](https://github.com/textileio/go-textile-core/pull/13), and this PR.


jsign:
Note here: I'll take out the LevelDB datastore for tests and only leave Badger. One of every 5 o 6 tests fail, so it's kind of flaky. I wonder if this is the same reason that the peerstore has it commented out. If that's the case, we reached the same conclusion.

tstoremem/metadata.go
```diff
@@ -29,7 +29,39 @@ func NewThreadMetadata() tstore.ThreadMetadata {
...
+}
+
+func (ts *memoryThreadMetadata) GetString(t thread.ID, key string) (*string, error) {
+	val := ts.getValue(t, key).(string)
+	return &val, nil
+}
+
+func (ts *memoryThreadMetadata) PutBytes(t thread.ID, key string, val []byte) error {
+	b := make([]byte, len(val))
+	copy(b, val)
```
jsign: Changed here to copy the byte slice. Might be debatable though, but I think is safer than a further change in the slice after `PutBytes` doesn't also mean changing the stored value.

jsign:
Feels ready to review, tests pass (leaving apart a particular test that is known to not be working yet).

When the[ `go-textile-core` PR gets merged and tagged](https://github.com/textileio/go-textile-core/pull/13), I'll add an extra commit here for the `go.mod` and switch for review.

jsign:
Considering we'll tag later, I updated `go.mod` to `go-textile-core` current master.

tstoreds/headbook.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("error when getting current heads from log %v: %w", key, err)
+	}
+
+	set := make(map[cid.Cid]struct{})
+	for i := range hr.Heads {
+		set[hr.Heads[i].Cid.Cid] = struct{}{}
+	}
+	for i := range heads {
+		if !heads[i].Defined() {
+			log.Warningf("ignoring head %s is is undefined for %s", heads[i], key)
```
sanderpick: double "is"

tstoreds/keybook.go
```diff
@@ -0,0 +1,193 @@
...
+	return pk, nil
+}
+
+// AddPubKey adds the public key of peer.ID which should match accordingly.
+func (kb *dsKeyBook) AddPubKey(t thread.ID, p peer.ID, pk ic.PubKey) error {
+	if pk == nil {
+		return fmt.Errorf("public key is nil")
+	}
+
+	if !p.MatchesPublicKey(pk) {
```
sanderpick: 👍

tstoreds/keybook.go
```diff
@@ -0,0 +1,193 @@
...
+func (kb *dsKeyBook) PubKey(t thread.ID, p peer.ID) (ic.PubKey, error) {
+	key := dsKey(t, p, kbBase).Child(pubSuffix)
+
+	var pk ic.PubKey
+	if v, err := kb.ds.Get(key); err == nil {
+		if pk, err = ic.UnmarshalPublicKey(v); err != nil {
+			return nil, fmt.Errorf("store backed public key %s can't be unmarshaled: %w", key, err)
+		}
+	} else if err == ds.ErrNotFound {
+		pk, err = p.ExtractPublicKey()
```
sanderpick: This case is a little weird. I think I removed it from the mem store version. Extracting and adding _within_ a get call feels like a black box... perhaps useful in the normal peerstore, but probably not what we want with threads. Checking if pubkey is not nil is useful to see if we have the log.

tstoreds/metadata.go
```diff
@@ -0,0 +1,113 @@
...
+	k := keyMeta(t, key)
+	v, err := ts.ds.Get(k)
+	if err == ds.ErrNotFound {
+		return err
+	}
+	if err != nil {
+		return fmt.Errorf("error when getting key from meta datastore: %w", err)
+	}
+	r := bytes.NewReader(v)
+	if err := gob.NewDecoder(r).Decode(res); err != nil {
```
sanderpick: 👍

threads.go
```diff
@@ -205,11 +205,19 @@ func (t *threads) Put(ctx context.Context, rec thread.Record, opts ...tserv.PutO
 func (t *threads) Pull(ctx context.Context, id thread.ID, lid peer.ID, offset cid.Cid, opts ...tserv.PullOption) ([]thread.Record, error) {
 	settings := tserv.PullOptions(opts...)
 
-	if t.PubKey(id, lid) == nil {
+	pk, err := t.PubKey(id, lid)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't fetch public key from book: %v", err)
```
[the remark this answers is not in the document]
sanderpick: After looking at the books, it looks like you are already wrapping these errors. Do you think we need the additional wrap here? It just makes the code less readable IMO.

service.go
```diff
@@ -195,16 +213,23 @@ func (s *service) push(ctx context.Context, rec thread.Record, id thread.ID, lid
...
+	logKey, err := s.threads.ReadKey(settings.Thread, settings.KeyLog)
+	if err != nil {
+		return fmt.Errorf("couldn't fetch read key from book: %v", err)
+	}
 	if logKey != nil {
 		keyLog = &pb.ProtoPeerID{ID: settings.KeyLog}
 	}
+	followKey, err := s.threads.FollowKey(id, lid)
+	if err != nil {
+		return fmt.Errorf("couldn't fetch follow key from addr book: %v", err)
```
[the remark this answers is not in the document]
jsign: (Gonna answer below about this)

threads.go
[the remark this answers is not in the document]
jsign: Yes, you're right. I think since I already wrap them in the books, it could be returned _as is_ at this level.

Regarding errors in general, I had pending a bit more thinking about this. Go 1.13 already has a stronger opinion about error wrapping and some new primitives to make life easier (since this string returning stuff leaves no room for the caller to check anything). 

For the moment I'm going to wipe off these `fmt.Errorf`s that are wrapped in books, and then consolidate better what I'm thinking about the error-style thing to chat about it.

tstoreds/keybook.go
jsign: Yeah, it's a bit magical. Also, seems unexpected that a read might eventually affect the store (an indirect a `.Put` is done). Gonna take it out.

PR #25, opened 2019-10-16
threadstore: new metadata testsuite & bugfix inmem metadata book

### LINKED ISSUE #23: Add Metadata test to Threadstore suite
There're some tests for Metadata but in the `Threadservice` suite scope that uses `inmem` implementation. 

I'll add some tests to a `metadata_suite.go` to test both `inmem` and `tstoreds`. Making it in other PR to avoid blocking the current one since it is making lots of changes.

jsign:
Fixes #23 

When executing the test, I realized that the inmem metadata book wasn't working correctly when getting non-existent keys. So, kind of already paid off. :)

@sanderpick, these changes are parallel to existing PR, so I'll leave it to your preference when to merge it. Leaving it here if the fixed bug may be bothering some tests.

### DIFF
--- a/test/metadata_suite.go
+++ b/test/metadata_suite.go
@@ -1,0 +1,160 @@
+package test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/textileio/go-textile-core/thread"
+	tstore "github.com/textileio/go-textile-core/threadstore"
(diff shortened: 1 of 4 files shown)


sanderpick:
Merge away!

Issue #17, opened 2019-10-11
Allow “pulling” from an entire thread, not just a single log


sanderpick:
Users should not need to care about each log - they should just pull at the thread level. We can move the current Pull method internally as it will still be needed.

sanderpick:
Also, currently pulled records are just handed back to the caller, which is not very useful. Related to #18

PR #27, opened 2019-10-16
[WIP] Dispatcher

carsonfarmer:
### Issue or RFC Endorsed by Textile's Maintainers

Signed-off-by: Carson Farmer <carson.farmer@gmail.com>

### Description of the Change

This PR consolidates some early draft code into a more cohesive (simpler) dispatcher framework. It is based around the idea of a singleton dispatcher, which persists and dispatches events, one at a time. The dispatcher blocks until a given event has been completely 'handled', which includes all registered callbacks. From discussions with the team on Slack, it because clear that some of the transactional behavior that we had initially planned is less critical, and could be mostly handled with a simple mutex. Then, waiting for reducers (which are each run concurrently) was greatly simplified. Additionally, some nice 'nice to have' features (like multiple errors being returned etc) were added for completeness thanks to @jsign. We have 100% test coverage, and while the interfaces are likely to change still, this provides a good base upon which to start.

Depends on https://github.com/textileio/go-textile-core/pull/18.

### Alternate Designs

There were a lot of alternative options here, include a more pub/sub style system (see #17 for an example), but this implementation covers most of the anticipated uses we have at the moment. It doesn't really address _consumers_ at this stage, but that isn't really the point here anyway.

### Possible Drawbacks

This is a bit more complicated than just leaving all the concurrency up to a third-party database, but what we add in complexity, I think we gain in flexibility. Additionally, we leave a lot of flexibility and power up to the reducers here, so the effectiveness of this code does rely to some degree on the effectiveness of the registered reduces. It is hard to strictly enforce good behavior this way, but... developer beware!

### Verification Process

New tests were added to verify the new features are doing what they claim to be doing. We have 100% test coverage, which is a great place to start from.

### Release Notes

Added a dispatcher framework with well-defined reducer handling and flexible underlying event store requirements. Test coverage for dispatcher starts at 100%.

### DIFF
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -23,7 +23,7 @@ var ErrPersistence = errors.New("persistence failure")
 var ErrTokenNotFound = errors.New("token not found")
 
 // Events are stored under the following db key pattern:
-// /events/<event.time><event.entityid><event.type>
+// /events/<event.entityid>/<event.time>/<event.type>
 // @todo: This is up for debate! It might make more sense to have time at the end of the key.
 // @todo: We might also want to include thread, and log information in the key?
 var baseKey = datastore.NewKey("/events")
@@ -92,8 +92,9 @@ func (d *Dispatcher) Dispatch(event dispatch.Event) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	var result *multierror.Error
-	key := baseKey.ChildString(string(event.Time())).ChildString(event.EntityID()).ChildString(event.Type())
+	key := baseKey.ChildString(event.EntityID()).ChildString(string(event.Time())).ChildString(event.Type())
 	// Add an Event's body to the event store as the value
+	// @todo: Should encode + dump whole Event here? Otherwise, need to parse key to extract type, time, etc...
 	if err := d.store.Put(key, event.Body()); err != nil {
 		return ErrPersistence
 	}
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -1,6 +1,7 @@
 package dispatcher
 
 import (
+	"errors"
 	"sync"
 
 	"github.com/hashicorp/go-multierror"
@@ -15,16 +16,17 @@ type Token uint
 // lastID is the last ID used by the singleton Dispatcher.
 var lastID uint
 
-const (
-	// ErrPersistence means that a write to the underlying event store failed.
-	ErrPersistence = esError("persistance failure")
-	// ErrTokenNotFound means the given token was not found in the dispatcher's reducer map.
-	ErrTokenNotFound = esError("token not found")
-)
+// ErrPersistence means that a write to the underlying event store failed.
+var ErrPersistence = errors.New("persistence failure")
 
-type esError string
+// ErrTokenNotFound means the given token was not found in the dispatcher's reducer map.
+var ErrTokenNotFound = errors.New("token not found")
 
-func (e esError) Error() string { return string(e) }
+// Events are stored under the following db key pattern:
+// /events/<event.time><event.entityid><event.type>
+// @todo: This is up for debate! It might make more sense to have time at the end of the key.
+// @todo: We might also want to include thread, and log information in the key?
+var baseKey = datastore.NewKey("/events")
 
 // Dispatcher is used to dispatch events to registered reducers.
 //
@@ -70,8 +72,7 @@ func (d *Dispatcher) Register(reducer dispatch.Reducer) Token {
 	return id
 }
 
-// Deregister removes a reducer based on its token. If the token is invalid (i.e. no associated reducer),
-// this is a no-op.
+// Deregister removes a reducer based on its token. If the token is invalid it will return an error.
 func (d *Dispatcher) Deregister(token Token) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
@@ -83,15 +84,15 @@ func (d *Dispatcher) Deregister(token Token) error {
 }
 
 // Dispatch dispatches a payload to all registered reducers. It returns a multierror object, which may contain
-// zero (nil) or more errors. Errors from reducer callbacks can be safely ignored and/or retried (and are prefixed
-// with "warning"), whereas errors due to event persistence are "critical" and will be prefixed as such.
+// zero (nil) or more errors. Errors from reducer callbacks can be safely ignored and/or retried, whereas errors due
+// to event persistence (`ErrPersistence`) are likely critical, and should be handled by the caller.  If a given
+// reducer does fail, it does not affect the other reducers or event store state. It is up to the caller to catch
+// this, and rerun if/as needed.
 func (d *Dispatcher) Dispatch(event dispatch.Event) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 	var result *multierror.Error
-	// Key format: <timestamp>/<entity-id>/<type>
-	// @todo: This is up for debate, its a 'fake' Event struct right now anyway
-	key := datastore.NewKey(string(event.Time())).ChildString(event.EntityID()).ChildString(event.Type())
+	key := baseKey.ChildString(string(event.Time())).ChildString(event.EntityID()).ChildString(event.Type())
 	// Add an Event's body to the event store as the value
 	if err := d.store.Put(key, event.Body()); err != nil {
 		return ErrPersistence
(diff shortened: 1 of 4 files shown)


dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if result != nil {
+		return result.ErrorOrNil()
+	}
+	return nil
+}
+
+// Query searches the internal event store and returns a query result. This is a proxy to the underlying event store's
+// Query method.
+func (d *Dispatcher) Query(query query.Query) (query.Results, error) {
```
carsonfarmer: We might need to add an additional Replay method at some point, but for now I'm leaving it out to avoid over-engineering things.

carsonfarmer:
As with https://github.com/textileio/go-textile-core/pull/18 I'm getting this in early for some discussion. Much of the implementation details here could change depending on what we do with core.

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+const (
+	// ErrPersistence means that a write to the underlying event store failed.
+	ErrPersistence = esError("persistance failure")
+	// ErrTokenNotFound means the given token was not found in the dispatcher's reducer map.
+	ErrTokenNotFound = esError("token not found")
+)
+
+type esError string
+
+func (e esError) Error() string { return string(e) }
```
jsign: Are you using this style of error defining compared to `errors.New()` for some reason?

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	lastID++
+	id := Token(lastID)
+	d.reducers[id] = reducer
+	return id
+}
+
+// Deregister removes a reducer based on its token. If the token is invalid (i.e. no associated reducer),
+// this is a no-op.
```
jsign: I think it isn't a `no-op` action since it errors when the token is invalid or already removed. Maybe I understand no-op in a different way.

Not strong opinions about if should return an error or not. I mostly prefer not in these cases, but I don't have a strong argument.

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+	if _, ok := d.reducers[token]; !ok {
+		return ErrTokenNotFound
+	}
+	delete(d.reducers, token)
+	return nil
+}
+
+// Dispatch dispatches a payload to all registered reducers. It returns a multierror object, which may contain
+// zero (nil) or more errors. Errors from reducer callbacks can be safely ignored and/or retried (and are prefixed
+// with "warning"), whereas errors due to event persistence are "critical" and will be prefixed as such.
```
jsign: When you refer to `critical` prefix, you mean that the error message has a `critical: ` prefix? That should be in `ErrPersistence`? If this prefix should be used by the caller, I think maybe nicer to return a typed error to avoid parsing messages (if I understood correctly)

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+func (d *Dispatcher) Dispatch(event dispatch.Event) error {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	var result *multierror.Error
+	// Key format: <timestamp>/<entity-id>/<type>
+	// @todo: This is up for debate, its a 'fake' Event struct right now anyway
+	key := datastore.NewKey(string(event.Time())).ChildString(event.EntityID()).ChildString(event.Type())
```
jsign: I would presume the `datastore` per application can be shared, so I think is a good idea to create a prefix key for the dispatcher. Like always starting with `/dispatcher` prefix in the keys to avoid a collision.

jsign: I'm not sure about the `event.Time()` value is safe to use in a key... still don't have the full picture. Couldn't this value cause collisions?

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,128 @@
...
+			if err := r.Reduce(event); err != nil {
+				errChan <- multierror.Prefix(err, "warning")
+			}
+		}(reducer)
+	}
+	wg.Wait()
+	// Close and then read from error channel to put into multierror
+	close(errChan)
+	for err := range errChan {
+		result = multierror.Append(result, err)
```
jsign: TBH, never used multierror; seems neat.
I wonder how it would really be used by the caller. Maybe just for logging, since they can't do much about it. If it was typed-error multierror maybe we could provide more info like the token of the reducer that failed. 

Maybe later when using it we can have more insight.

dispatcher/dispatcher.go
carsonfarmer: I read about it here: https://dave.cheney.net/2016/04/07/constant-errors

dispatcher/dispatcher.go
carsonfarmer: Ah yes, leftover comment... its no longer a no-op!

carsonfarmer: I like the error here because if you were expecting to remove something and its not there, I feel like you need to know that...

dispatcher/dispatcher.go
jsign: I supposed ;), that's a kind of well-known discussion in the Go world. To be honest, I'm from the school that thinks that if you include a malicious party dependency in your code, you'll be screwed up anyway. (as per `io.EOF` isn't constant example)

dispatcher/dispatcher.go
carsonfarmer: This is a good idea. I hadn't thought that you'd share the store, but you totally could.

jsign: Here's an example I did in the Keyboof if helps: https://github.com/textileio/go-textile-threads/blob/master/tstoreds/keybook.go

jsign: But `event.Time()` is the time of event generation or when was received by the current node?

dispatcher/dispatcher.go
carsonfarmer: Yeh, initially did it this way so that if the caller didn't know about multierror, they'd still be good (its just an error afterall)... but since we're moving this into threads, we _could_ just return the typed multierror... might as well right?

dispatcher/dispatcher.go
carsonfarmer: Ah another leftover comment I think. I'll remove that. We're returning the constant errors now instead.

dispatcher/dispatcher.go
jsign: Can be an option. But never thought about much about multierror semantics really. 
Also, not sure what consequences has a failed reducer. (The event will retry? is this fatal to the application state (since other reducers might succeed)? maybe is about defining the reducer semantics and bounded-context of state between them in docs to avoid possible pitfalls)

dispatcher/dispatcher.go
carsonfarmer: Time of event generation.

dispatcher/dispatcher.go
carsonfarmer: Yeh this needs to be thought through a lot more I think. In theory, it should probably be retried, but I wondered if that should be left up to the caller, or if we should provide some retry facilities directly... or... I _think_ intuitively you're right about simply documenting things. Something along the lines of, if a given reducer fails, it does not affect the other reducers or event store state. It is up to the caller to catch this, and rerun if/as needed? But dunno for sure.

dispatcher/dispatcher.go
carsonfarmer: So then are you advocating to just use something like the following?
```
// ErrPersistence means that a write to the underlying event store failed.
ErrPersistence = errors.New("persistence failure")
```

carsonfarmer: Also noticed a spelling error there

dispatcher/dispatcher.go
carsonfarmer: We can't guarantee that a given reducer is idempotent (it probably isn't), so rerunning reducers that _didn't_ fail is bad...

dispatcher/dispatcher.go
jsign: Only saying that's a usual idiomatic way to define sentinel errors. Your implementation based on Dave's argument is totally correct from the standpoint of const vs var... so, no strong claim here.

dispatcher/dispatcher.go
jsign: I guess it depends on what guides you tell the developer. If isn't idempotent, I agree totally. But if that's the case, this means that the reducer is in a fatal state?

carsonfarmer: I'm going to leave this as a default error for now, but we can expose it more explicitly later if we want.

carsonfarmer:
Closing this in favor of ongoing work in https://github.com/textileio/go-eventstore

PR #28, opened 2019-10-16
Handle follower addresses

### LINKED ISSUE #20: Allow follower designation in an invite payload
Currently, there's no mechanism for saying, "this is an invite to be a follower, which means that your address will be advertised to others".

sanderpick:
Closes #20 

- Adds handling for orchestrating a follower's new log address within a thread.
- Adds a top-level `NewThreadstore` to the datastore implementation.
- Adds some IDE warning fixes. 

An interesting side effect of using an event body to pass log info (keys, addresses, etc.), is that followers are not able to pick up new addresses. In other words, a follower cannot pick-up addresses of other _followers_. Followers can't pull from followers. We probably want that functionality, but it means placing new address info (or maybe just all log info) at either a different or additional position when authoring messages. One option could be to include it at the message envelope layer, encrypted with follow key.


sanderpick:
I need to add some more tests here. Otherwise, looking okay.

pb/threads.proto
```diff
@@ -56,8 +56,8 @@ message PushRequest {
 
 // PushReply is the response from a PushRequest.
 message PushReply {
-    // ok means the push succeeded.
-    bool ok = 1;
+    // newAddr is a new address for the sender's log.
+    bytes newAddr = 1 [(gogoproto.customtype) = "ProtoAddr"];
```
sanderpick: The result of a sending logs could result in the recipient becoming a follower, in which case they will return a new address for the sender's log.

service.go
```diff
@@ -144,6 +149,10 @@ func (s *service) Push(ctx context.Context, req *pb.PushRequest) (*pb.PushReply,
 		if err = rec.Verify(logpk); err != nil {
 			return nil, err
 		}
+
+		if err = s.handleLogUpdate(ctx, req.ThreadID.ID, req.LogID.ID, rec); err == nil {
```
sanderpick: Attempt to process this record as a log update, meaning that the record body will have log info.

sanderpick: (This happens when a peer that has added a follower writes the new address to their own log. The other peers need to update there addresses as well.)

service.go
```diff
@@ -285,7 +294,44 @@ func (s *service) push(
...
+				go func() {
+					pk, err := s.threads.PubKey(id, lid)
+					if err != nil {
+						log.Error(err)
+						return
+					}
+					lg := thread.LogInfo{
+						ID:     lid,
+						PubKey: pk,
+						Addrs:  []ma.Multiaddr{reply.NewAddr.Multiaddr},
```
sanderpick: Send out the new address.

service.go
```diff
@@ -567,33 +621,91 @@ func (s *service) handleInvite(
...
 		if err != nil {
-			return nil, err
+			return
 		}
 		ownLog = &lg
 	}
 
+	// If not readable, return a new address for the sender's log.
+	// This peer becomes a follower.
+	if !logs.Readable() {
```
sanderpick: This is a heavy handed assumption. We can revisit if a more explicit flow is needed.

service.go
```diff
@@ -567,33 +621,91 @@ func (s *service) handleInvite(
...
+func (s *service) handleLogUpdate(ctx context.Context, id thread.ID, lid peer.ID, rec thread.Record) error {
+	event, err := cbor.EventFromRecord(ctx, s.threads.dagService, rec)
+	if err != nil {
+		return err
+	}
+	rk, err := s.threads.ReadKey(id, lid)
+	if err != nil {
+		return err
+	}
+	if rk == nil {
```
sanderpick: In this case, we already have the log and maybe the read key.

service.go
```diff
@@ -285,7 +294,44 @@ func (s *service) push(
 				return
 			}
 
-			log.Debugf("reply from %s: %t", p, reply.Ok)
+			log.Debugf("received reply from %s", p)
+
+			// Handle new log addresses
+			if reply.NewAddr != nil {
+				err = s.threads.AddAddr(id, lid, reply.NewAddr.Multiaddr, pstore.PermanentAddrTTL)
```
jsign: Replaying what happened up to this point to validate:
- We want to push a record to a log
- We send it to all known log addresses
- Every `PushReply` now has a `newAddr` 
- If `newAddr` has value, at this quoted line we add it as a new address to `id, lid` which is basically is adding (or refreshing) the _confirmed_ multiaddr for future pulls from the log for other peers (or whatever addrs are needed for this log). Addresses from owners and followers are undistinguishable for the books, right?

The `newAddr` field being not nil is valid for every possible type of record?

sanderpick: That sounds correct to me! I'm not in love with the `newAddr` response, perhaps we could specialize messages further with a `Push` and `PushToFollower` message type / API at some point.

sanderpick:
Thanks @jsign! I'll make those tweaks this evening and get this merged. I got down a rabbit hole today.

service.go
```diff
@@ -144,6 +149,10 @@ func (s *service) Push(ctx context.Context, req *pb.PushRequest) (*pb.PushReply,
 		if err = rec.Verify(logpk); err != nil {
 			return nil, err
 		}
+
+		if err = s.handleLogUpdate(ctx, req.ThreadID.ID, req.LogID.ID, rec); err == nil {
+			log.Infof("log %s updated", req.LogID.ID.String())
+		}
```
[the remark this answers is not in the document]
sanderpick: The else case would be that the record was not a log update (could be anything). I added some logging to the Add and Put methods which will handle logging the fact that something happened here.

PR #29, opened 2019-10-18
Playground Shell

sanderpick:
This should help us dog food this sooner rather than later.

The shell is a simple interactive prompt where input prepended with ":" is interpreted as a command, e.g., `:use`, `:add`.


client.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	pblgs := make([]*pb.GetRecordsRequest_LogEntry, 0)
+	for _, lg := range lgs {
+		pblgs = append(pblgs, &pb.GetRecordsRequest_LogEntry{
+			LogID:  &pb.ProtoPeerID{ID: lg.ID},
+			Offset: &pb.ProtoCid{Cid: offsets[lg.ID]},
+			Limit:  int32(limit),
+		})
+	}
```
jsign: Maybe this loop work can be done in the above loop? Since `offsets` is a map, it shouldn't have some repeated-values problem which might affect `append`. (+ `make` with `len(offsets)` cap)

client.go
```diff
@@ -0,0 +1,400 @@
...
+				return
+			}
+
+			log.Debugf("received reply from %s", p)
+		}(addr)
+	}
+
+	// Finally, publish to the thread's topic
+	err = s.publish(id, req)
+	if err != nil {
```
jsign: Sugg: Inline err

client.go
```diff
@@ -0,0 +1,400 @@
...
+	for _, l := range info.Logs {
+		laddrs, err := s.threads.store.Addrs(settings.ThreadID, l)
+		if err != nil {
+			return err
+		}
+		addrs = append(addrs, laddrs...)
+	}
+
+	// Add additional addresses
+	addrs = append(addrs, settings.Addrs...)
```
jsign: Sugg: Move to `addrs` definition with `len(settings.Addrs)` as min capacity

client.go
```diff
@@ -0,0 +1,400 @@
...
+
+// getDialOption returns the WithDialer option to dial via libp2p.
+func (s *service) getDialOption() grpc.DialOption {
+	return grpc.WithContextDialer(func(ctx context.Context, peerIDStr string) (net.Conn, error) {
+		id, err := peer.IDB58Decode(peerIDStr)
+		if err != nil {
+			return nil, fmt.Errorf("grpc tried to dial non peer-id: %s", err)
+		}
+		c, err := gostream.Dial(ctx, s.threads.host, id, ThreadProtocol)
+		return c, err
```
jsign:
```suggestion
 		return gostream.Dial(ctx, s.threads.host, id, ThreadProtocol)
```

util/util.go
```diff
@@ -29,16 +29,75 @@ func CreateLog(host peer.ID) (info thread.LogInfo, err error) {
...
+		if sk != nil {
+			li, err := t.Store().LogInfo(id, lid)
+			if err != nil {
+				return info, err
+			}
+			return li, nil
+		}
+	}
+	return info, nil
+}
```
jsign: Sugg: unname `thread.LogInfo` param? or maybe assign in the non-case error (instead of `li`). Got some seconds thinking about when `info` was assigned :smile:

util/util.go
```diff
@@ -29,16 +29,75 @@ func CreateLog(host peer.ID) (info thread.LogInfo, err error) {
...
+	}
+	if info.PubKey != nil {
+		return
+	}
+	info, err = CreateLog(t.Host().ID())
+	if err != nil {
+		return
+	}
+	err = t.Store().AddLog(id, info)
+	return
```
jsign: Sugg: direct return

client.go
sanderpick: yep, doing now

client.go
sanderpick: nice, adding

client.go
sanderpick: Adding!

util/util.go
sanderpick: cool, I cleaned up the non-error case for readability

util/util.go
sanderpick: Yep, sounds good.

sanderpick:
Thanks for the comments @jsign. I pushed some changes.

I have since,
- Added some logic that allows new logs to be picked up during a records request. This is achieved by sending a full HEAD map of all logs in the request, which is then compared by the receiver who can tag on missing logs. This is useful when a new writer joins a thread.
- Added a simple daemon that is now running on us-west-1. We can use this as a test follower.

The shell can be run with...
```
go run ./exe/shell
```
The daemon can be run with...
```
go run ./exe/daemon
```

jsign:
> Thanks for the comments @jsign. I pushed some changes.
> 
> I have since,
> 
>     * Added some logic that allows new logs to be picked up during a records request. This is achieved by sending a full HEAD map of all logs in the request, which is then compared by the receiver who can tag on missing logs. This is useful when a new writer joins a thread.
> 
>     * Added a simple daemon that is now running on us-west-1. We can use this as a test follower.
> 
> 
> The shell can be run with...
> 
> ```
> go run ./exe/shell
> ```
> 
> The daemon can be run with...
> 
> ```
> go run ./exe/daemon
> ```

Cool, good idea.
I'll spot some time to give it a try!

sanderpick:
I still need to sink some time into understanding why the peers don't see each other in CI. I have a hunch we'll have to disable MDNS for CI stuff. This should be configurable anyway as it's not needed on a server.

sanderpick:
Cool! You could try adding the remote thread I've been playing with at some point:

```
:thread team /ip4/104.210.43.77/tcp/4006/p2p/12D3KooWMY5LbHywKAdwDddpTE1BsBZFyT4oR8QCrYCsLLC72bZ4/thread/bafk44nyecg2a32nuqx5fytdscs5or7fdprg7hhkfzwtudvlegsxlqgi
```

After all the tweaks, it's working well between two computers. The only issue I see is that sometimes the remote peer is not dialable for some amount of time, then comes back. This can be particularly annoying when adding a thread. I've seen similar behavior before. We may need to dig in there... I don't see any reason it shouldn't be _always_ dialable.

PR #39, opened 2019-10-31
gRPC Web

sanderpick:
- Integrates https://github.com/improbable-eng/grpc-web which wraps the gRPC server.
- Adds the associated TS plugin to `pb/Makefile`.


pb/Makefile
```diff
@@ -1,14 +1,19 @@
 PB = $(wildcard *.proto)
 GO = $(PB:.proto=.pb.go)
+PROTOC_GEN_TS_PATH = "./node_modules/.bin/protoc-gen-ts"
 
 all: $(GO)
 
 %.pb.go: %.proto
-	protoc -I=. -I=$(GOPATH)/src --gogofaster_out=\
+	protoc -I=. -I=$(GOPATH)/src \
```
carsonfarmer: yesssss!

pb/threads_pb.d.ts
```diff
@@ -0,0 +1,480 @@
+// package: threads.pb
```
carsonfarmer: Are we sure we want to check these into version control? Maybe good for now, but might be better to just export and publish as a package at some point right?

threads.go
```diff
@@ -100,13 +110,54 @@ func NewThreads(
 	go t.rpc.Serve(listener)
 	pb.RegisterThreadsServer(t.rpc, t.service)
 
+	webrpc := grpcweb.WrapServer(t.rpc)
```
carsonfarmer: Awesome, glad this integrated so nicely.

pb/threads_pb.d.ts
sanderpick: Totally. Let's see how these custom gogo types play with JS and go from there... I can then pull over the npm make-and-publish-on-tag stuff from `go-textile`

PR #40, opened 2019-11-01
Shell nits

sanderpick:
A quickie to improve the shell UX:
- Shows a more "real" prompt that consumes the line instead of just printing it.
- Shows the header wall-clock time of messages

Also delete the unused threadserver folder which is replaced by the HTTP proxy.


threads.go
```diff
@@ -350,6 +344,12 @@ func (t *threads) AddRecord(
 		return
 	}
 
+	// Push out the new record
```
jsign: So prefer succeed sending to bus if pushRecord fails?

sanderpick: At this point, the record has been committed... network failure is a different concern

jsign: :+1:

PR #41, opened 2019-11-02
Shell: Improve thread visibility

sanderpick:
Adds some more niceties to the shell:
- Changes the command paradigm so that you can see output from all threads
- "Enter" a thread with `:enter` such that you can type messages as before
- You can always type a message to a specific thread by prefixing the name: `:team hey!`
- "Exit" a thread with exit 
- Stack commands like `:team:address` (show team's addresses)
- Use `:help` to see available commands

### DIFF
--- a/exe/shell/main.go
+++ b/exe/shell/main.go
@@ -37,17 +37,15 @@ var (
 	api      tserv.Threadservice
 	threadID thread.ID
 
-	grey   = color.New(color.FgHiBlack).SprintFunc()
-	green  = color.New(color.FgHiGreen).SprintFunc()
-	cyan   = color.New(color.FgHiCyan).SprintFunc()
-	yellow = color.New(color.FgHiYellow).SprintFunc()
-	red    = color.New(color.FgHiRed).SprintFunc()
+	grey  = color.New(color.FgHiBlack).SprintFunc()
+	green = color.New(color.FgHiGreen).SprintFunc()
+	cyan  = color.New(color.FgHiCyan).SprintFunc()
+	blue  = color.New(color.FgHiBlue).SprintFunc()
+	red   = color.New(color.FgHiRed).SprintFunc()
 
-	bootstrapPeers = []string{
-		"/ip4/104.210.43.77/tcp/4001/ipfs/12D3KooWSdGmRz5JQidqrtmiPGVHkStXpbSAMnbCcW8abq6zuiDP", // us-west
-		"/ip4/20.39.232.27/tcp/4001/ipfs/12D3KooWLnUv9MWuRM6uHirRPBM4NwRj54n4gNNnBtiFiwPiv3Up",  // eu-west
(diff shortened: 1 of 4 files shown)


exe/shell/main.go
```diff
@@ -218,88 +269,123 @@ func addressCmd() (out string, err error) {
...
-			if err != nil {
-				return "", err
-			}
-			threadID = info.ID
-		} else {
-			threadID = thread.NewIDV1(thread.Raw, 32)
-		}
-		err = ds.Put(datastore.NewKey("/names/"+name), threadID.Bytes())
+func enterCmd(name string) (out string, err error) {
+	idv, err := ds.Get(datastore.NewKey("/names/" + name))
```
carsonfarmer: Since you use this prefix a bunch here, you could probably just create a const to make it easier to reference?

exe/shell/main.go
```diff
@@ -218,88 +269,123 @@ func addressCmd() (out string, err error) {
...
 
-func threadCmd(args []string) (out string, err error) {
-	if len(args) == 0 {
-		err = fmt.Errorf("enter a thread name to use")
+func threadsCmd() (out string, err error) {
+	q, err := ds.Query(query.Query{Prefix: "/names"})
+	if err != nil {
+		return
+	}
+	all, err := q.Rest()
```
carsonfarmer: Since you're just iterating through here anyway, maybe just use the iterator (`.Next()`)?

exe/shell/main.go
```diff
@@ -179,25 +190,65 @@ func handleLine(line string) (out string, err error) {
...
-	err = sendMessage(line)
+	if threadID.Defined() {
+		err = sendMessage(threadID, line)
+	} else {
+		err = fmt.Errorf("enter a thread with `:enter` or specify thread name with :<name>")
+	}
+	return
+}
+
+func cmdCmd() (out string, err error) {
```
carsonfarmer: I like the feel of the interactive cli you have here a lot... but kinda surprised there isn't an existing 'solution' to make generating things like help and usage msgs easier?

exe/shell/main.go
```diff
@@ -179,25 +190,65 @@ func handleLine(line string) (out string, err error) {
...
-		default:
-			err = fmt.Errorf("unknown command: %s", args[0])
+		case "enter":
+			if len(parts) == 1 {
+				err = fmt.Errorf("missing thread name")
+				return
+			}
+			return enterCmd(parts[1])
+		case "exit":
+			threadID = thread.Undef
```
carsonfarmer: Nice

exe/shell/main.go
sanderpick: There are some good shell libs but this is more a of a chat interface that allows for command. So, it's a bit different in feel. Mostly just playing around here though... when the time comes for a "real" shell, we can reevaluate.

exe/shell/main.go
sanderpick: ah yeah, that's better, though it's nice to have `i` to then deal with the newlines in the same loop - just gonna leave this for now - surely will change.

PR #66, opened 2019-11-07
WebSocket Server

### LINKED ISSUE #55: Publish updates over a web socket
We'll need a web socket server and some way of authenticating, most likely with a JWT. 

re/ https://github.com/textileio/go-textile-threads/issues/54

sanderpick:
Closes #55 

- Fixes `exe/util.Build` so that it uses a disc-backed peerstore.
- Registers a "notifee" for the MDNS module in the shell.
- Push records for registered threads to web socket clients. Web clients register for a list of threads. The example `index.html` file shows how you might do this from a list in the URL query params:
```
// Subscribe to thread in URL query
let threads
let params = new URLSearchParams(window.location.search);
if (params.has('thread')) {
    threads = params.getAll('thread')
}
let msg = {
    type: 'subscribe',
    threads: threads
}
socket.send(JSON.stringify(msg))
```

### DIFF
--- a/ws/client.go
+++ b/ws/client.go
@@ -2,10 +2,12 @@ package ws
 
 import (
 	"bytes"
+	"encoding/json"
 	"net/http"
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/textileio/go-textile-core/thread"
 )
 
 const (
@@ -36,6 +38,11 @@ var (
 	space   = []byte{' '}
 )
 
+type msg struct {
+	Type    string   `json:"type"`
+	Threads []string `json:"threads"`
+}
+
 // Client is a middleman between the websocket connection and the hub.
 type Client struct {
 	hub *Hub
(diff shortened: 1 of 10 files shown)


sanderpick:
^ Updated PR description. TL;DR is I added handling for clients subscribing to specific threads.

ws/client.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+}
+
+// writePump pumps messages from the hub to the websocket connection.
+//
+// A goroutine running writePump is started for each connection. The
+// application ensures that there is at most one writer to a connection by
+// executing all writes from this goroutine.
+func (c *Client) writePump() {
```
jsign: Most `.Write` errors ignored since is a PoC? Maybe an extra `// ToDo` to keep track?

sanderpick: Added!

sanderpick:
Cheers @jsign! I realized this morning that this package should function more like the MDNS setup, where it's up to the dev to start the server / service. On mobile, for example, we won't want this thing running or even included with `go bind`. So, I pushed that change as well.

sanderpick:
Nice, tests passed... could be a fluke though.

Screenshot 2019-11-07 09 12 27

PR #67, opened 2019-11-07
tests: use the same peers for all tests

### LINKED ISSUE #42: Fix tests in CI
For some reason, the tests peers can usually not find each other when run on CircleCI.

sanderpick:
Closes #42 

Doh. Of course, the same peers should be used for all these thread tests OR they should be shutdown afterward. Opting for using the same here. Peers were starting up on the same port, which leads to the "connected to wrong peer" error message.


jsign:
I wonder how sometimes this can run correctly.

sanderpick:
yeah, no idea!

PR #70, opened 2019-11-07
Use follow key for auth, allow follower w/o read key

### LINKED ISSUE #69: GetLogs & GetRecords authentication
The idea here is to use the follow key for auth and assume that the caller already has the read key, as in, it was passed over some back channel.

sanderpick:
Closes #69 

Depends on: https://github.com/textileio/go-textile-core/pull/27.

I wasn't able to keep the "cheap" change to thread keys and make this work. So, this includes a bit more explicit work around the key change, namely, the thread info object now has the keys and not the log info object.


exe/shell/main.go
```diff
@@ -278,15 +286,17 @@ func handleLine(line string) (out string, err error) {
 
 func cmdCmd() (out string, err error) {
 	out = pink(":help  ") + grey("Show available commands.\n")
-	out += pink(":address  ") + grey("Show host or active thread addresses.\n")
+	out += pink(":address  ") + grey("Show the host or active thread addresses.\n")
 	out += pink(":threads  ") + grey("Show threads.\n")
 	out += pink(":add <name>  ") + grey("Add a new thread with name.\n")
-	out += pink(":add <name> <address>  ") + grey("Add an existing thread with name at address.\n")
+	out += pink(":add <name> <address> <follow-key> <read-key>  ") + grey("Add an existing thread with name at address using follow and read keys.\n")
```
jsign: Maybe mentioning in the description that keys should be base58 encoded

exe/shell/main.go
```diff
@@ -359,6 +369,28 @@ func addCmd(args []string) (out string, err error) {
...
+	if len(args) > 3 {
+		rkb, err := base58.Decode(args[3])
+		if err != nil {
+			return "", nil
+		}
+		rk, err = sym.NewKey(rkb)
+		if err != nil {
+			return "", nil
+		}
+	}
```
jsign: The intention isn't to fail but to ignore invalid keys? (`return "", nil` on `err != nil`)

threads.go
```diff
@@ -675,21 +707,30 @@ func (t *threads) startPulling() {
 	tick := time.NewTicker(PullInterval)
 	defer tick.Stop()
```
jsign: Opt: maybe moving this after the initial pull, to avoid overlap.

sanderpick: 👍

exe/shell/main.go
sanderpick: Typo! Fixing...

exe/shell/main.go
sanderpick: 👍

sanderpick:
Oop, I created a tangle for myself by continuing work (and fixing stuff) here last night: https://github.com/textileio/go-textile-threads/tree/sander/ws-rpc

I'm just gonna push that work here, which should fix these tests, and then continue that elsewhere.

PR #72, opened 2019-11-10
ws: migrate to an RPC arch

sanderpick:
This migrates the WS server to an RPC style where incoming requests are mapped to responses via a request ID, i.e., https://github.com/textileio/js-threads-client/blob/master/src/index.ts. 

I've stubbed out the client methods which mirror `Threadservice` but will need to do some additional testing. That being said, I'd still take some eyes on this in the meantime.


sanderpick:
This is ready for some eyes.

cbor/event.go
```diff
@@ -47,14 +46,14 @@ func NewEvent(ctx context.Context, dag format.DAGService, body format.Node, sett
 		return nil, err
 	}
 	eventHeader := &eventHeader{
-		Time: settings.Time.Unix(),
+		Time: time.Now().Unix(),
```
carsonfarmer: Should we think about getting an HCL implementation in here soon? Something like https://github.com/lafikl/hlc looks very simple?

ws/client.go
```diff
@@ -106,94 +261,49 @@ func (c *Client) readPump() {
...
+			switch caller.Method {
+			case "addThread":
+				result, err = c.addThread(ctx, caller.Args...)
+			case "pullThread":
+				err = fmt.Errorf("todo")
+			case "deleteThread":
+				err = fmt.Errorf("todo")
+			case "addFollower":
+				err = fmt.Errorf("todo")
+			case "addRecord":
```
carsonfarmer: Why not these ones yet? Specific reason, or just not there yet?

ws/hub.go
```diff
@@ -67,3 +66,11 @@ func (h *Hub) run() {
 		}
 	}
 }
+
+// encodeRecord base64 encodes a thread record.
+func encodeRecord(rec thread.Record) []byte {
```
carsonfarmer: Nice

ws/client.go
sanderpick: Oops! I forgot to enable the new methods... added.

cbor/event.go
sanderpick: Nice—issue added: https://github.com/textileio/go-textile-threads/issues/76

carsonfarmer:
Nice, re-approved 👍.

PR #74, opened 2019-11-11
Eventstore & Threadservice integration

### LINKED ISSUE #26: Ensure that a thread is not pulled concurrently
As @jsign mentioned in a review, we should avoid have multiple pull jobs running for a single thread.

> Idea: Maybe adding extra logic to avoid pulling concurrently from the same thread.

> Like, if t.Pull is slow enough and tick ticks again, maybe pulls for the same thread start queueing. If that makes sense there're could be multiple options: from discarding it since there's an already running one, or cancelling running one and retry from zero again, etc.

> This can give more guarantees that at most there are 1 live pull for the existing threads (which also in the future we can think to move it to a (capped) worker model to cap it if that's too much pressure).

At the very least we could have a `sync.Map` or equivalent that tracks live pulls.

### LINKED ISSUE #73: evenstore: integrate with threadservice

jsign:
This PR is about integrating `Store` with Threads.
The journey of doing this uncovered some bugs and questions to think about. Bugs were fixed, but many questions remain open for me.

TL;DR:
- Read all this description first, even if many things don't make sense yet
- Read the `README.md` which has visuals and high-level explanations.
- Make some coffee 
- Review, and support again on `README.md` or me directly!

---

I think a good place to start the review, is taking a look at the [`README.md`](https://github.com/textileio/go-textile-threads/blob/435720cc6cd007c0d8e9a0471c60aaf2aea92324/eventstore/README.md) I created. It's a very quickly written explanation of how things fit together, with some details about what components are for and their main responsibility. It will also help during reviewing the code to know better where you're standing at. It also has a nice visual diagram, which I'm not 100% proud of but hopes it helps!

The next step before the review is maybe taking a look at  `TestE2E` test in `store_test.go`. It should basically get you a feeling on how `Store` works with most default settings. If no `Options` are passed to `Store`, it will automatically create for you:
- A `Threadservice`
- A default repo-path, which is the base-path for all internal data (datastore, ipfslite assets, threadservice, etc)
- A default listening-port for ipfs-lite.
- A default `EventCodec` (which now can only by the json-patcher).
- Debug information turned off.

The defaults would be creating a store simply with: `.Store()`. A more custom creation would be something like: `.NewStore(WithDebug(true), WithEventCodec(ec), WithListenPort(8008), WithRepoPath(tmpDir), WithThreadservice(ts))`

--- 
The integration was mostly guided while doing in parallel a minimal app that can be found in `exe/e2e_counter`. It basically runs an app that has a single `Model` in which instances have counters (with name and value). 

The app can be run in `writer` mode, which means that it will create (or restore) the `Store`, create a new counter, and start incrementing the counter every X seconds. If run in `reader` mode, the reader will create the `Store` and _join_ the thread, which will start syncing with what the other peer is doing. Whenever the internal store changes state (that said, the counter gets updated after syncing), it prints the value in stdout.

The easiest way to test this (moving to `exe/e2e_counter`):
On terminal 1 (the writer): `go run *.go -clean -writer -repo .threads1`
On terminal 2 (the reader): `go run *.go -clean -repo .threads2`
After running the writer, you can check a generated file `.e2e_counter_writeraddr`. This file is generated to make life easier to share where the writer is listening, and the follow and read key of the underlying thread. The reader will assume this file exists to join the thread. If you recall the `TestE2E` test, this is similar to the middle code section of it.

For some reason yet to be discovered, the reader seems to recognize new events only doing pulls (every 10s so wait a bit). Most probably is a silly thing that may be missing.

---

While moving forward in the implementation, one issue that we discussed in some of the early review days came to bite me (#26). So I decided to fix it (more comments in own-review). 
Another bug that took some time to figure out was a classical go-routine accessing a loop-variable that wasn't being copied (more comments in own-review). Quite silly things, but both weren't that trivial to discover since there're are a lot of moving parts. :) 

---

A particular new feature of the `Store` is a new broadcast feature that allow `Store` users to be notified about state changes in the `Store`, so to react and do something. This was leveraged in the `e2e_counter` app by the reader so to print to stdout all existing counts and their values.

---

Here are some questions or future work. 
For some of them, I already created an issue, for others would like some confirmation about if that would be a good idea (gonna enumerate them to ease discussion):
1) Enable a Threadservice subscription *since* a particular height of logs. Since there're different logs with different heights, it doesn't seem clear how... but the main intention is having a way for end-users (of `Threadservice`) to rebuild-state from inception. That's the case for `Store`, which now relies on subscribing soon enough before the `PullThread` is done (and a quite big MaxPullLimit for history). Mostly discussed with @sanderpick  which agrees that sounds reasonable, but repeating here for competitive.
2) While debugging something, I *think* that on `client.go` in `pushRecords` the peer push records to itself. Was a flashing moment and have no confirmation, just mentioning it if could make sense (or to take a further look or issue reminder)
3) While running the `exe/e2e_counter`, if you notice the logs for the second peer, you'll realize that there's an error that the threadservice proxy can't listen on 5050 since is already used (by the first peer/writer). I was wondering, maybe in `threads.go` we could try with `5050` as default, and try to retry with port 0 to find some free port?
4) A weird thing you may notice is a two lumberjack logger setup between `Store` and `Threadservice`. I think both shouldn't really do that work, and have the opportunity to get them injected. If it isn't the case, let them create and bootstrap them. The problem with both of them doing the setup is that the last setup will mark all levels as `Error` level, overriding any setup that the first call did. I didn't think much more about it, maybe there's other possible solution. Hope this isn't confusing.
5) Leverage that `go-logging` has span supports for opentracing, to have some tracing instrumentation of the code. At this moment, is quite hard to have a good view of where the code is taking time and how things flow. *Sounds* like a cheap thing to do, and can help with discovering *real* flows, slow paths, how much time things take, etc. (all apart from the usual mem/cpu profiling in Go). Not sure, but seeing traces of complex code is always nice #82 for discussion :)

Now some harder questions that we have already mentioned before in other talks but would like to list here to make them more clear. 

Not really blocking to figure them out, but these things really limits/can backfire in the future:
- `Threadservice` will notify the `Store` about new `Events`. At this point, the notification is synchronous. The `Store` speed can bother the `Threadservice` broadcaster and thus other important listeners?
- If the above broadcast fails for whatever reason. What does that mean?
- When the `Store` will ultimately do a `AddRecord` to store a local change in the own-thread-log. What should we do if that fails/timeouts?. Now it will `log.Fatal`. But we can explore some other options? Maybe 2 o 3 retries?. Now the call to `AddRecord` is synchronous to the original `.Commit` in the `Store`, so `Commit` would fail if you can't add a new record. Could this be relaxed at what cost?. If `.Commit()` succeeds, but then the record couldn't be added to the log... we would have some nasty problem between `Store` state and the underlying persistence _transaction log_
- If a transaction modifies three instances, it will ultimately generate three events (in current json-patcher). I'm still not sure this is a wise thing, despite that the three events will be appended in order together in the peer log. But fetching from other logs can be slower or partial, which means applying events that correspond to a single transaction partially. I guess a better solution is to generate one single event per transaction, so if we receive a new remote event we know have all the changes from a transaction (have given more consistency guarantees). I hope isn't confusing!
- _Store_ to _Threads_ mapping is something to discover yet. Currently, one Store goes to one Thread. But it can be different and that should be somewhat _easy_ to change (or even as a configuration) in how things were built. The `SingleThreadAdapter` is the one that makes that kind of decision and is quite separated from all the rest.
- Still not sure about `core.Event` (the `Event` concept at the `Store` level) attributes. I think `Type` is a too generic name, since now is used to store the model name from which the json-patch should be applied (and so, which Model reducer should take care of it, and others ignore it). Also, it has a `Time` (wall-clock) attribute that I'm not sure what would be meaningful yet (up to this point, isn't used but in the `Dispatcher` key-namespace, for ordering (own ordering? not thread-wise)).

---

Somewhat immediate task after merging:
- Moving contents in `core` folder to `go-textile-core`. And maybe some other things. #80 
- Making lumber-jack logging things more clear between components. Related with 4) in the above list.

---

Closes #73
Closes #26


eventstore/store.go
```diff
@@ -1,49 +1,188 @@
...
+	}
+	adapter := newSingleThreadAdapter(s.ctx, s, id)
+	adapter.Start()
+	return nil
+}
+
+// StartFromAddr should be called immediatelly after registering all schemas
+// and before any operation on them. It pulls the current Store thread from
+// thread addr
+func (s *Store) StartFromAddr(addr ma.Multiaddr, followKey, readKey *symmetric.Key) error {
```
jsign: Similarly with `Start`, `StartFromAddr` will bootstrap a `Store` from a Thread addr.
If other peer send us a invitation link, using that info we can start syncing things. 

As both method comments mention, models should register all types before calling start. If that isn't the case, when an Event will be reducer, no model would be found and would fail. I think this can be eventually solved by registering schema data in the log itself, so to make this thing automatic (and kind of consistent too). 

Anyway, doesn't seem too bad to ask to register all things before starting things (since that's how many similar things work). But seems like a good idea for the future. (Not to say how schema updates should be handled? :headexplode: :stuck_out_tongue_closed_eyes: )

eventstore/store.go
```diff
@@ -84,6 +242,25 @@ func (s *Store) readTxn(m *Model, f func(txn *Txn) error) error {
 	return nil
 }
 
+// Close closes the store
+func (s *Store) Close() {
+	s.cancel()
+	s.localEventsBus.bus.Discard()
+	s.stateChanged.bus.Discard()
+	s.datastore.Close()
+	if s.ownThreadService {
```
jsign: Only close the Threadservice if we created a default one. If a custom one was provided, let the original owner deal with it since can being used for other things. Opinions?

eventstore/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"os"
+	"testing"
+	"time"
+
+	ds "github.com/ipfs/go-datastore"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/textileio/go-textile-threads/core"
+)
+
+func TestE2E(t *testing.T) {
```
jsign: Two peers, two parallel stores, sync, check if things went nice.

eventstore/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+	time.Sleep(time.Second * 3) // Wait a bit for sync
+
+	dummyInstance2 := &dummyModel{}
+	checkErr(t, m2.FindByID(dummyInstance.ID, dummyInstance2))
+	if dummyInstance2.Name != dummyInstance.Name || dummyInstance2.Counter != dummyInstance.Counter {
+		t.Fatalf("instances of both peers must be equal after sync")
+	}
+}
+
+func TestOptions(t *testing.T) {
```
jsign: Test options use, and re-open an existing Store. Checks if store.Close works correctly of left things locked (helped to fix some bugs!)

eventstore/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type mockEventCodec struct {
+	called bool
+}
+
+var _ core.EventCodec = (*mockEventCodec)(nil)
+
+func (dec *mockEventCodec) Reduce(e core.Event, datastore ds.Datastore, baseKey ds.Key) error {
+	dec.called = true
```
jsign: Used to check if the Store was really using our custom `EventCodec`.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
+package eventstore
```
jsign: Refer to `README.md` to understand this component responsibility (in short: glue between Store and Threads)

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/textileio/go-textile-core/thread"
+	tserv "github.com/textileio/go-textile-core/threadservice"
+	threadcbor "github.com/textileio/go-textile-threads/cbor"
+	"github.com/textileio/go-textile-threads/util"
+)
+
+const (
+	addRecordTimeout  = time.Second * 10
+	fetchEventTimeout = time.Second * 15
+)
```
jsign: Not much though about it.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+	a.ownLogID = li.ID
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go a.threadToStore(&wg)
+	go a.storeToThread(&wg)
+	wg.Wait()
+}
+
+func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
```
jsign: Every time a new Record is discovered by Threadservice, this method will be called to send it to the Store for reducing.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+	defer sub.Discard()
+	wg.Done()
+	for {
+		select {
+		case <-a.ctx.Done():
+			log.Infof("cancelling thread event dispatch to store for thread %s since ctx was cancelled", a.threadID)
+			return
+		case rec, ok := <-sub.Channel():
+			if !ok {
+				log.Info("ending dispatch of events to store since thread channel was closed") // ToDo: reconsider action
```
jsign: Huh, the threadservice closed the subscription channel. 
Opinions if this is a reasonable thing to do?
Currently, it will end processing thread->store direction of data.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+				return
+			}
+			if rec.LogID() == a.ownLogID {
+				continue // Ignore our own events since Store already dispatches to Store reducers
```
jsign: When our own Events are added to our log, it will also reach this method.. so lets discard this case since the dispatcher will already send this events to reducers.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			ctx, cancel := context.WithTimeout(context.Background(), fetchEventTimeout)
+			defer cancel()
+
+			event, err := threadcbor.EventFromRecord(ctx, a.api, rec.Value())
+			if err != nil {
+				log.Fatalf("error when getting event from record: %v", err) // ToDo: Buffer them and retry...
+				continue
+			}
```
jsign: ToDo comment. 
What if we fail to get the Event from the new Record due to some networking timeout or who knows what?
Maybe buffer it and retry? Have some queue to keep processing in order? This would make things async (more on this mentioned in the original PR description). Not super clear the risks of that.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+			if err != nil {
+				log.Fatalf("error when getting read key for thread %s: %v", a.threadID, err)
+			}
+			if readKey == nil {
+				log.Fatalf("read key not found for thread %s/%s", a.threadID, rec.LogID())
+			}
+			node, err := event.GetBody(ctx, a.api, readKey)
+			if err != nil {
+				log.Fatalf("error when getting body of event on thread %s/%s: %v", a.threadID, rec.LogID(), err)
+			}
```
jsign: Similar to above. Maybe some 2 o 3 retries and die would be a bit better without jumping to async complexity.

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+			storeEvent, err := a.store.eventFromBytes(node.RawData())
+			if err != nil {
+				log.Fatalf("error when unmarshaling event from bytes: %v", err)
+			}
+			log.Debugf("dispatching to store external new record: %s/%s", rec.ThreadID(), rec.LogID())
+			if err := a.store.dispatch(storeEvent); err != nil {
```
jsign: Finally, dispatch to Store to reduce this event!

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+				log.Fatal(err)
+			}
+		}
+	}
+}
+
+func (a *singleThreadAdapter) storeToThread(wg *sync.WaitGroup) {
```
jsign: The other direction. When a local change gets generated in the store, it will come here to decide how to map it to Threads. In this case, it is a single Thread per store... so in short means adding the event to this single thread (own log).

eventstore/storethread.go
```diff
@@ -0,0 +1,145 @@
...
+				log.Fatalf("error when generating node for own log for thread %s: %v", a.threadID, err)
+			}
+			ctx, cancel := context.WithTimeout(context.Background(), addRecordTimeout)
+			defer cancel()
+
+			log.Debugf("adding new local store event to own log from entityid: %s", event.EntityID())
+			_, err = a.api.AddRecord(ctx, a.threadID, n)
+			if err != nil {
+				log.Fatalf("error writing record: %v", err)
+			}
```
jsign: Same as with other methods!, can timeout due to network or whatever.

In all of these cases always decided the pessimistic (and more safe for state) approach until we can resolve something.

jsonpatcher/jsonpatcher.go
```diff
@@ -202,4 +227,8 @@ func (je patchEvent) Type() string {
 	return je.TypeName
 }
 
+func (je patchEvent) Node() (format.Node, error) {
+	return cbornode.WrapObject(je, mh.SHA2_256, -1)
```
jsign: the json patch struct is encoded in cbor to get the node that will be stored in the log.

eventstore/dispatcher.go
```diff
@@ -69,6 +73,7 @@ func (d *Dispatcher) Dispatch(event core.Event) error {
 	// Safe to fire off reducers now that event is persisted
 	g, _ := errgroup.WithContext(context.Background())
 	for _, reducer := range d.reducers {
+		reducer := reducer
```
jsign: This is a fix for the concurrency issue that took some time to figure out!
The below go routine depended on mutating closure. 
Instead of doing this, there's the other usual: make value a param of go-routine.

eventstore/store.go
```diff
@@ -1,49 +1,188 @@
...
+		return thread.ID{}, false, err
+	}
+	id, err := thread.Cast(v)
+	return id, true, err
+}
+
+// Start should be called immediatelly after registering all schemas and before
+// any operation on them. If the store already boostraped on a thread, it will
+// continue using that thread. In the opposite case, it will create a new thread.
+func (s *Store) Start() error {
```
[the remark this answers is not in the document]
sanderpick: So, are you not able to register models _after_ start?

eventstore/store.go
sanderpick: Ahh, I see. You answered the question above. 

`StartFromAddr`, I like that - makes it feel very easy to get going.

sanderpick: Related to your thought about the schema... we could add an additional address protocol like,
```
/thread/<threadID>/model/<entityID>
```
Then, either check the address for the model protocol or add an additional method, `StartFromEntity`.

This would fit naturally with "getting" an entity like a shared google doc, etc. and would also mean you could return the schema associated with that model. Perhaps a you would ignore events not associated with that entity, perhaps not.

eventstore/store.go
sanderpick: It might be cleaner to have a _make default thread service_ method that is available to the user. Then they just use that instead of the option. In either case, they have to close it down.

eventstore/storethread.go
sanderpick: Seems ok to me since this will only happen if the thread service closes.

eventstore/storethread.go
sanderpick: A failure here seems like a legit failure, not a network issue, since all the data should be available.

eventstore/storethread.go
sanderpick: Here though, they body of the event may be difficult to reach on the network... +1 for some retries

exe/local_eventstore/books/main.go
```diff
@@ -1,10 +1,11 @@
...
-	ds "github.com/ipfs/go-datastore"
+	"io/ioutil"
+	"os"
+
 	"github.com/textileio/go-textile-threads/core"
 	es "github.com/textileio/go-textile-threads/eventstore"
-	"github.com/textileio/go-textile-threads/jsonpatcher"
 )
 
 type book struct {
```
sanderpick: Sweet!

jsonpatcher/jsonpatcher.go
sanderpick: Niiiiice

eventstore/store.go
```diff
@@ -109,3 +286,113 @@ func isValidModel(t interface{}) bool {
...
+	return scl.c
+}
+
+// Discard indicates that no further notifications will be received
+// and ready for being garbage collected
+func (scl *StateChangeListener) Discard() {
+	scl.listener.Discard()
+}
+
+func setLogLevels(systems map[string]logger.Level, writer io.Writer, color bool) error {
```
sanderpick: Could be shared util method?

threads.go
```diff
@@ -263,54 +268,67 @@ func (t *threads) AddThread(
...
 					}
 				}
-			}
-		}(lg)
+			}(lg)
+		}
+		wg.Wait()
+		<-ptl
+	default:
+		log.Warningf("pull thread %s ignored since already being pulled", id)
```
sanderpick: Great! Thanks for nipping that issue as well.

eventstore/dispatcher.go
sanderpick: Good catch... those ones are always hard to spot.

sanderpick:
>     1. Enable a Threadservice subscription _since_ a particular height of logs. Since there're different logs with different heights, it doesn't seem clear how... but the main intention is having a way for end-users (of `Threadservice`) to rebuild-state from inception. That's the case for `Store`, which now relies on subscribing soon enough before the `PullThread` is done (and a quite big MaxPullLimit for history). Mostly discussed with @sanderpick  which agrees that sounds reasonable, but repeating here for competitive.

Yep, +1. I ran into an issue before when `PullThread` used to pull history instead of "pull for new content". The issue is that the blockchain structure is not very good at replaying, only rewinding. Not sure how to solve it without building up a sort of temporary index in-mem, and then replaying that.

>     2. While debugging something, I _think_ that on `client.go` in `pushRecords` the peer push records to itself. Was a flashing moment and have no confirmation, just mentioning it if could make sense (or to take a further look or issue reminder)

Ok, I can check on this. 

>     3. While running the `exe/e2e_counter`, if you notice the logs for the second peer, you'll realize that there's an error that the threadservice proxy can't listen on 5050 since is already used (by the first peer/writer). I was wondering, maybe in `threads.go` we could try with `5050` as default, and try to retry with port 0 to find some free port?

Another option is to make the `DefualtThreadService` (spelling?) option take a port for the proxy. If the user has a firewall configured, they may actually want the error so that actions can be taken to a) change the firewall, b) shutdown whatever else is holding the port.

>     4. A weird thing you may notice is a two lumberjack logger setup between `Store` and `Threadservice`. I think both shouldn't really do that work, and have the opportunity to get them injected. If it isn't the case, let them create and bootstrap them. The problem with both of them doing the setup is that the last setup will mark all levels as `Error` level, overriding any setup that the first call did. I didn't think much more about it, maybe there's other possible solution. Hope this isn't confusing.

Since the `Store` is the new top-level component, I'd vote for just removing lumberjack logger setup from `Threadservice`.

>     5. Leverage that `go-logging` has span supports for opentracing, to have some tracing instrumentation of the code. At this moment, is quite hard to have a good view of where the code is taking time and how things flow. _Sounds_ like a cheap thing to do, and can help with discovering _real_ flows, slow paths, how much time things take, etc. (all apart from the usual mem/cpu profiling in Go). Not sure, but seeing traces of complex code is always nice #82 for discussion :)

Sounds great.

> - `Threadservice` will notify the `Store` about new `Events`. At this point, the notification is synchronous. The `Store` speed can bother the `Threadservice` broadcaster and thus other important listeners?

Hard to say at this point!

> - If the above broadcast fails for whatever reason. What does that mean?
> - When the `Store` will ultimately do a `AddRecord` to store a local change in the own-thread-log. What should we do if that fails/timeouts?. Now it will `log.Fatal`. But we can explore some other options? Maybe 2 o 3 retries?. Now the call to `AddRecord` is synchronous to the original `.Commit` in the `Store`, so `Commit` would fail if you can't add a new record. Could this be relaxed at what cost?. If `.Commit()` succeeds, but then the record couldn't be added to the log... we would have some nasty problem between `Store` state and the underlying persistence _transaction log_

Same for these two, but I added some comments about retries in the review. 

> - If a transaction modifies three instances, it will ultimately generate three events (in current json-patcher). I'm still not sure this is a wise thing, despite that the three events will be appended in order together in the peer log. But fetching from other logs can be slower or partial, which means applying events that correspond to a single transaction partially. I guess a better solution is to generate one single event per transaction, so if we receive a new remote event we know have all the changes from a transaction (have given more consistency guarantees). I hope isn't confusing!

Nice, I like that. Seems like the codec could transform a batch of store events into a single thread event and the reverse.

> - _Store_ to _Threads_ mapping is something to discover yet. Currently, one Store goes to one Thread. But it can be different and that should be somewhat _easy_ to change (or even as a configuration) in how things were built. The `SingleThreadAdapter` is the one that makes that kind of decision and is quite separated from all the rest.

Mentioned above, but I sorta like this initial constraint, which makes a thread feel more like a "room" where many different entities can live.

> - Still not sure about `core.Event` (the `Event` concept at the `Store` level) attributes. I think `Type` is a too generic name, since now is used to store the model name from which the json-patch should be applied (and so, which Model reducer should take care of it, and others ignore it). Also, it has a `Time` (wall-clock) attribute that I'm not sure what would be meaningful yet (up to this point, isn't used but in the `Dispatcher` key-namespace, for ordering (own ordering? not thread-wise)).

Yeah, tricky. We could probably replace wall-clock time with an HLC time at some point when we run into pain there. Not sure about `Type`. Do you think it would be better as some kind of struct?

eventstore/options.go
```diff
@@ -0,0 +1,213 @@
...
+
+func newDefaultDatastore(repoPath string) (ds.Datastore, error) {
+	path := filepath.Join(repoPath, defaultDatastorePath)
+	if err := os.MkdirAll(path, os.ModePerm); err != nil {
+		return nil, err
+	}
+	return badger.NewDatastore(path, &badger.DefaultOptions)
+}
+
+func newDefaultThreadservice(ctx context.Context, logWriter io.Writer, listenPort int, repoPath string, debug bool) (threadservice.Threadservice, error) {
```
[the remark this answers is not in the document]
jsign: :+1:  #86

eventstore/store.go
jsign: > Ahh, I see. You answered the question above.
> 
> `StartFromAddr`, I like that - makes it feel very easy to get going.

Yes!, that was the idea. Hopefully will become more clear when making the more serious app on how it feel to use this. The `e2e_counter` example is quite dry. 



> Related to your thought about the schema... we could add an additional address protocol like,
> 
> ```
> /thread/<threadID>/model/<entityID>
> ```
> 
> Then, either check the address for the model protocol or add an additional method, `StartFromEntity`.
> 
> This would fit naturally with "getting" an entity like a shared google doc, etc. and would also mean you could return the schema associated with that model. Perhaps a you would ignore events not associated with that entity, perhaps not.

Interesting. Might be an interesting feature... only following some models.

eventstore/store.go
jsign: Feels better. Having the Store provide a method to build a Threadservice is quite heavy and seems out of scope. Created #87.

eventstore/storethread.go
jsign: But if getting the `Event` from the `Record` doesn't involve a _healthy_ network connection? If I recall correctly the `Record` should resolve a CID and thus use the `DAGResolver`.
Maybe I'm confused?

eventstore/storethread.go
jsign: Created #89 to keep track of this.

eventstore/storethread.go
sanderpick: Yep, that's correct. However, in practice those nodes will have been added locally by the time you request them, meaning no network required. That said, I'm not sure we can assume that will always be the case.

eventstore/options.go
```diff
@@ -0,0 +1,213 @@
...
+// WithEventCodec configure to use ec as the EventCodec
+// manager for transforming actions in events, and viceversa
+func WithEventCodec(ec core.EventCodec) StoreOption {
+	return func(sc *StoreConfig) error {
+		sc.EventCodec = ec
+		return nil
+	}
+}
+
+func loadKey(pth string) ic.PrivKey {
```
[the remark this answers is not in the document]
jsign: Yeah, also in #85 you added some more possibly convenient things.
Created #90 to keep track.

eventstore/options.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		priv, err = ic.UnmarshalPrivateKey(key)
+		if err != nil {
+			panic(err)
+		}
+	}
+	return priv
+}
+
+func parseBootstrapPeers(addrs []string) ([]peer.AddrInfo, error) {
```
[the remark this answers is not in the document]
jsign: Added in #90

eventstore/options.go
```diff
@@ -0,0 +1,213 @@
...
+		Debug:     debug,
+	})
+	if err != nil {
+		return nil, err
+	}
+	boots, err := parseBootstrapPeers(bootstrapPeers)
+	if err != nil {
+		return nil, err
+	}
+	lite.Bootstrap(boots)
```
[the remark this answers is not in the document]
jsign: Seems reasonable. I think that when I'll make the PR to move creating the default `Threadservice` away from `Store`, which is somewhat entangled with all this (custom bootstrap peer list, and bootstrap function call).

eventstore/store.go
jsign: Added to #90

jsign:
> > ```
> > 1. Enable a Threadservice subscription _since_ a particular height of logs. Since there're different logs with different heights, it doesn't seem clear how... but the main intention is having a way for end-users (of `Threadservice`) to rebuild-state from inception. That's the case for `Store`, which now relies on subscribing soon enough before the `PullThread` is done (and a quite big MaxPullLimit for history). Mostly discussed with @sanderpick  which agrees that sounds reasonable, but repeating here for competitive.
> > ```
> 
> Yep, +1. I ran into an issue before when `PullThread` used to pull history instead of "pull for new content". The issue is that the blockchain structure is not very good at replaying, only rewinding. Not sure how to solve it without building up a sort of temporary index in-mem, and then replaying that.

Yeah, not trivial. Created #91 to keep track. I might have some ideas but not gave full focus yet.

> > ```
> > 3. While running the `exe/e2e_counter`, if you notice the logs for the second peer, you'll realize that there's an error that the threadservice proxy can't listen on 5050 since is already used (by the first peer/writer). I was wondering, maybe in `threads.go` we could try with `5050` as default, and try to retry with port 0 to find some free port?
> > ```
> 
> Another option is to make the `DefualtThreadService` (spelling?) option take a port for the proxy. If the user has a firewall configured, they may actually want the error so that actions can be taken to a) change the firewall, b) shutdown whatever else is holding the port.

That seems great. I added this consideration on #87.
 
> > ```
> > 4. A weird thing you may notice is a two lumberjack logger setup between `Store` and `Threadservice`. I think both shouldn't really do that work, and have the opportunity to get them injected. If it isn't the case, let them create and bootstrap them. The problem with both of them doing the setup is that the last setup will mark all levels as `Error` level, overriding any setup that the first call did. I didn't think much more about it, maybe there's other possible solution. Hope this isn't confusing.
> > ```
> 
> Since the `Store` is the new top-level component, I'd vote for just removing lumberjack logger setup from `Threadservice`.

Sounds good to me. Created #92. I wonder if Store also should receive an optional config to provide a lumberjack too. If that isn't the case, can create one. Just to allow the caller to use the same logger for other things he would like. Not sure anyways, just a random idea.
 
 
> > * If a transaction modifies three instances, it will ultimately generate three events (in current json-patcher). I'm still not sure this is a wise thing, despite that the three events will be appended in order together in the peer log. But fetching from other logs can be slower or partial, which means applying events that correspond to a single transaction partially. I guess a better solution is to generate one single event per transaction, so if we receive a new remote event we know have all the changes from a transaction (have given more consistency guarantees). I hope isn't confusing!
> 
> Nice, I like that. Seems like the codec could transform a batch of store events into a single thread event and the reverse.

:+1:  Created #93

> > * _Store_ to _Threads_ mapping is something to discover yet. Currently, one Store goes to one Thread. But it can be different and that should be somewhat _easy_ to change (or even as a configuration) in how things were built. The `SingleThreadAdapter` is the one that makes that kind of decision and is quite separated from all the rest.


> > * Still not sure about `core.Event` (the `Event` concept at the `Store` level) attributes. I think `Type` is a too generic name, since now is used to store the model name from which the json-patch should be applied (and so, which Model reducer should take care of it, and others ignore it). Also, it has a `Time` (wall-clock) attribute that I'm not sure what would be meaningful yet (up to this point, isn't used but in the `Dispatcher` key-namespace, for ordering (own ordering? not thread-wise)).
> 
> Yeah, tricky. We could probably replace wall-clock time with an HLC time at some point when we run into pain there. Not sure about `Type`. Do you think it would be better as some kind of struct?

I think a string is just fine, since its enough to let models identify if the Event corresponds to them to reduce it. Regarding time, I think I rushed a bit when writing. `Time()` is a `[]byte`, so not necessarily `time.Time`. I think this was a wise decision considering what you mention of using logical clocks in other future `EventCodecs`. If now isn't used for `json-patch` doesn't mean won't be important in other implementations. :+1:

jsign:
Thanks, @sanderpick  for the review!
Most of the discussion decisions were transformed into issues. Most probably will merge as is, and then jump straight into a new PR with most reasonable fixes.

sanderpick:
Great, sounds good to me.

jsign:
Fixed the test which was related with #92.

Resolved all this thing as follows:
- `setLogLevel` was moved to `util` package
- `setLogLevel` was split in two functions
- one function `SetupDefaultLoggingConfig` just configures the global `logging` to use lumberjack with a defined path. This is a util function that the end-user can call if wants default logging configuration. Don't like lumberjack?, just don't call this and configure it as you want. 
- the other function `SetLogLevels` which is still used by `Store` and `Threadservice`, if their config options indicate `Debug`. The different with was done before, is that `logging.SetAllLoggers(logger.ERROR)` isn't part of this function, so it won't mess around other system levels. It will only be changed what is indicated to do.
- This allows to enable `Debug` for `Store`, and not to `Threadservice`, or whatever someone likes.
- This also means that `Store` shoudn't be aware of other dependency subsystems... every component will deal with their own system/subsystems and doesn't need to know about anything else. If they receive a `Debug` opt, they should send `Debug` opts to dependencies to let them whatever is correct for them.

sanderpick:
Great!

PR #85, opened 2019-11-15
Fix up client method responses

sanderpick:
Mainly this enables `getRecord` and other related methods to return a JSON encoded record:

```
// Create a new client and connect
window.client = new threads.Client()
client.connect('ws://localhost:8080').then(() => {
  const tid = 'bafkufevg3bqezf3mx7vmloyezupzkdx5ozsfudwky3o3c6kiqvqsldy'
  const rid = 'bafyreifok6utuqevl7uon255kvdg6edrcfhtnax6eueyuncpl2f3iqjewi'
  // Fetch a record
  client.getRecord(rid, tid).then((res) => {
    console.log(res)
  }).catch((err) => {
    console.error(err)
  })
})
```
Returns...
```
{
    "id": "bafyreifok6utuqevl7uon255kvdg6edrcfhtnax6eueyuncpl2f3iqjewi",
    "thread_id": "bafkufevg3bqezf3mx7vmloyezupzkdx5ozsfudwky3o3c6kiqvqsldy",
    "record_node": "WLTlVlcAFCOsk4NMRcAKLwqDjGNc3hVw/q7n3Ks0HMzi1HxQ+zBRYCPByM+wS5DDzbhUzzfLsGD8cF7vptezMshZevZOrlGSd3eGFD4QXncSe1AYO1cTMY5bHiCU3bwbA0osDdBSvwKXWIXfbb2wP9yv4Tci7MgAhQB7mR6iP2VoFNhwvASr43ubwBO9VAF0IvSZODHsGhCQnrjaLG+A2vovzVmcOpiMQ7ic+y9qmDm1Ymv6xpU=",
    "event_node": "omRib2R52CpYJQABcRIgovMszk/9Y2dDNea4JkTTZkyLgLkfw/GU8VmjF04z8AJmaGVhZGVy2CpYJQABcRIgSURimRMwYvozolv0wNcs6OrgWmlwzOSkGPsvEmLd6/A=",
    "body_node": "WCIICIELEYVXuTDIjV7gCwMRL3UqY7z/oM5n1E1+/lTWsEOg",
    "header_node": "WE2F79iDylvfauQPLNYvyip5CTeY37W//uLRT0HtOCoN+L+rJ3J7XTMxwVdowe+i/naasl8ptqw/O2jT9M212Iy4ZgAXKUu4kmarzLXVeQ=="
}
```

From that info, you can construct the full fledged record CBOR node (same as in the normal thread service, assuming you have the keys).

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -155,9 +155,9 @@ func DecodeKey(k string) (*sym.Key, error) {
 	return sym.NewKey(b)
 }
 
-// CapArgs returns args with new capacity cap.
-func CapArgs(args []string, cap int) []string {
-	capped := make([]string, 0, cap)
-	copy(capped, args)
-	return capped
+// PadArgs returns args with new length pad.
+func PadArgs(args []string, len int) []string {
+	padded := make([]string, len)
+	copy(padded, args)
+	return padded
 }
--- a/util/util.go
+++ b/util/util.go
@@ -2,9 +2,13 @@ package util
 
 import (
 	"crypto/rand"
+	"strings"
 
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/peerstore"
+	swarm "github.com/libp2p/go-libp2p-swarm"
+	"github.com/mr-tron/base58"
 	ma "github.com/multiformats/go-multiaddr"
 	sym "github.com/textileio/go-textile-core/crypto/symmetric"
 	"github.com/textileio/go-textile-core/thread"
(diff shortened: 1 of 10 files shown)


util/util.go
```diff
@@ -102,3 +106,58 @@ func GetOrCreateOwnLog(t tserv.Threadservice, id thread.ID) (info thread.LogInfo
...
+// CanDial returns whether or not the address is dialable.
+func CanDial(addr ma.Multiaddr, s *swarm.Swarm) bool {
+	parts := strings.Split(addr.String(), "/"+ma.ProtocolWithCode(ma.P_P2P).Name)
+	addr, _ = ma.NewMultiaddr(parts[0])
+	tr := s.TransportForDialing(addr)
+	return tr != nil && tr.CanDial(addr)
+}
+
+// DecodeKey from a string into a symmetric key.
+func DecodeKey(k string) (*sym.Key, error) {
```
jsign: This is useful. I'll use it in the other PR eventually.

util/util.go
```diff
@@ -102,3 +106,58 @@ func GetOrCreateOwnLog(t tserv.Threadservice, id thread.ID) (info thread.LogInfo
...
+	if err != nil {
+		return nil, err
+	}
+	return sym.NewKey(b)
+}
+
+// PadArgs returns args with new length pad.
+func PadArgs(args []string, len int) []string {
+	padded := make([]string, len)
+	copy(padded, args)
```
jsign: Nitpick: copy will copy min of `len` and `len(args)`. If `len(args)` greater than `len`, something bad would happen. Maybe is at the risk of the caller, just mentioning it before seeing where is used.

ws/hub.go
```diff
@@ -51,7 +62,25 @@ func (h *Hub) run() {
...
-			msg := encodeRecord(rec.Value())
+			rid := rec.Value().Cid().String()
+
+			ctx, cancel := context.WithTimeout(h.ctx, encodeRecordTimeout)
+			jrec, err := recordToJSON(ctx, h.service, rec.Value())
+			if err != nil {
+				log.Errorf("error converting record %s to JSON: %v", rid, err)
+				break
+			}
+			cancel()
```
jsign: maybe moving to a `defer cancel()` after `.WithTimeout` to cover the case of `err != nil`.

ws/hub.go
```diff
@@ -67,10 +96,73 @@ func (h *Hub) run() {
...
+type jsonThreadInfo struct {
+	ID        string   `json:"id"`
+	Logs      []string `json:"logs"`
+	FollowKey string   `json:"follow_key"`
+	ReadKey   string   `json:"read_key,omitempty"`
+}
+
+// threadInfoToJSON returns a JSON version of thread info for transport.
+func threadInfoToJSON(info thread.Info) *jsonThreadInfo {
+	var logs []string
```
jsign: nitpick: `logs := make([]string, 0, len(info.Logs)`

ws/hub.go
```diff
@@ -67,10 +96,73 @@ func (h *Hub) run() {
...
+	}, nil
+}
+
+// encodeNode returns a base64-encoded version of n.
+func encodeNode(n format.Node) string {
+	data := n.RawData()
 	encoded := make([]byte, base64.StdEncoding.EncodedLen(len(data)))
 	base64.StdEncoding.Encode(encoded, data)
-	return encoded
+	return string(encoded)
```
jsign: Nitpick: this is exactly the implementation of [.EncodeToString()](https://golang.org/src/encoding/base64/base64.go?s=4932:4986#L164) :smile:

ws/server.go
```diff
@@ -40,6 +40,8 @@ func NewServer(service tserv.Threadservice, addr string) *Server {
 	go func() {
 		for {
 			select {
+			case <-ctx.Done():
+				return
```
jsign: Nitpick: maybe some log info _shutting down server_?

ws/hub.go
sanderpick: `defer` here is in a blocking for loop - wouldn't that lead to a resource leak?

util/util.go
sanderpick: The caller _should_ have the sense that after doing something like `args = PadArgs(args, 2)`, they cannot access `args[3]`. Maybe a better name would be `TruncateArgs`? Or maybe `padded := make([]string, max(len, len(args))`

ws/hub.go
sanderpick: Oh :) changing!

jsign:
If you want to merge this PR, and then I'll merge #74. I'm totally ok to rebase again before merging mine. :+1:

ws/hub.go
jsign: Ops, yes, didn't notice the outer loop. So then just adding `cancel()`  before `break`.

util/util.go
jsign: Oh, `TruncateArgs` feels much better seeing this code. The other option changes what the function does. I'll leave it to whatever you think is the best :)

sanderpick:
Sounds good! Thanks for the review. I pushed some tweaks.

PR #94, opened 2019-11-15
lint fixes

jsign:
Lint fixes.


jsign:
Most fixes:
- use single `for range` in unique case selects
- duplicate imports
- unused struct vars
- unecessary select
- unused funcs
- unnecessary ignored multiple returns

PR #98, opened 2019-11-19
store: playing with json models

sanderpick:
Just some playing around in here with a `RegisterJSON` method on the `Store`. I didn't work past the getEntityID` panic, but seems doable. Thoughts? Worth pursuing?

### DIFF
--- a/eventstore/model_test.go
+++ b/eventstore/model_test.go
@@ -1,6 +1,7 @@
 package eventstore
 
 import (
+	"encoding/json"
 	"errors"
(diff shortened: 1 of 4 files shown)


jsign:
Didn't know about `dynamicstruct` library, pretty neat.
This idea seems to unlock the necessary to use the API as is, so I agree that it seems doable.

Internally, I mentioned another possible approach that I'll try to sketch too and see how if feels.

jsign:
Superseded by #99

PR #99, opened 2019-11-19
store: JSONMode feature

jsign:
Proposal, with comments soon.


eventstore/model.go
```diff
@@ -130,7 +134,13 @@ func (m *Model) Reduce(event core.Event) error {
 }
 
 func (m *Model) validInstance(v interface{}) (bool, error) {
-	vLoader := gojsonschema.NewGoLoader(v)
+	var vLoader gojsonschema.JSONLoader
+	if m.jsonMode {
```
jsign: If we are in json mode.. use a bytes loader instead of a goloader. Which basically means use plain json vs a go type.

eventstore/model.go
```diff
@@ -214,16 +224,25 @@ func (t *Txn) Save(updated ...interface{}) error {
...
+			// lets just say that is exactly the json representation that is currently in
+			// the datastore.
+			previous = beforeBytes
+		} else {
+			before := reflect.New(t.model.valueType.Elem()).Interface()
+			err = json.Unmarshal(beforeBytes, before)
+			if err != nil {
+				return err
+			}
+			previous = before
```
jsign: In JSONMode, `Action`  `Previous` and `Current` aren't Go Type representations of instances... they are just the json strings representations. So, here saved work for the store, no reflect magic and unmarshaling from datastore.

eventstore/model.go
```diff
@@ -284,6 +303,14 @@ func (t *Txn) FindByID(id core.EntityID, v interface{}) error {
...
 		return err
 	}
+	if t.model.jsonMode {
+		str := string(bytes)
+		rflStr := reflect.ValueOf(str)
+		reflV := reflect.ValueOf(v)
+		reflV.Elem().Set(rflStr)
+
+		return nil
+	}
```
jsign: In `FindByID` we'll store in `v` (which should be a `*string`) just the plain byte json we got from the store. In this mode we avoid the `json.Unmarshal` to return a Go Type.

eventstore/model.go
```diff
@@ -317,24 +344,67 @@ func (t *Txn) Discard() {
...
+		v := reflect.ValueOf(t)
+		if v.Type().Kind() != reflect.Ptr {
+			v = reflect.New(reflect.TypeOf(v))
+		}
+		v = v.Elem().FieldByName(idFieldName)
+		if !v.IsValid() || v.Type() != reflect.TypeOf(core.EmptyEntityID) {
+			panic("invalid instance: doesn't have EntityID attribute")
+		}
+
+		return core.EntityID(v.String())
```
jsign: `getEntityID`, in JSONMode should interpret the json string of the instance to see if has an `ID` attribute, and if not empty if has uuid format. (Some hacks here to distinguish between empty ID from ommited field... just to get the idea)

eventstore/model.go
```diff
@@ -317,24 +344,67 @@ func (t *Txn) Discard() {
...
+		strPatchedValue := string(patchedValue)
+		reflectPatchedValue := reflect.ValueOf(&strPatchedValue)
+		v.Elem().Set(reflectPatchedValue.Elem())
+	} else {
+		v := reflect.ValueOf(t)
+		if v.Type().Kind() != reflect.Ptr {
+			v = reflect.New(reflect.TypeOf(v))
+		}
+		v.Elem().FieldByName(idFieldName).Set(reflect.ValueOf(newID))
+	}
```
jsign: `setNewEntityID` which autogenerates ids...
In JSON mode, will inject the generated ID in the same `*string` instance that was called on `Create`.
leveraged a json patch to do that change

exe/local_eventstore/json-books/main.go
```diff
@@ -0,0 +1,170 @@
...
+	Author string
+	Meta   bookStats
+}
+
+type bookStats struct {
+	TotalReads int
+	Rating     float64
+}
+
+func main() {
```
jsign: An analogous books example but with jsonmode.
If `go run main.go` should work.

jsonpatcher/jsonpatcher.go
```diff
@@ -50,7 +51,7 @@ func init() {
 
 // New returns a JSON-Patcher EventCodec
 func New() core.EventCodec {
-	return &jsonPatcher{}
+	return &jsonPatcher{jsonMode: true}
```
jsign: Now the `EventCodec` jsonpatcher should be aware if is working in jsonmode or not.
Basically it tells him to know if `v instance{}` things that receives are Go Types that should marshal, or are string jsons which already can use to do its work.

eventstore/model.go
```diff
@@ -317,24 +344,67 @@ func (t *Txn) Discard() {
...
-		v = reflect.New(reflect.TypeOf(v))
-	}
 	newID := core.NewEntityID()
-	v.Elem().FieldByName(idFieldName).Set(reflect.ValueOf(newID))
+	// JSONMODE: assumes true. make `getEntity` a method of `Model` so maybe can access
+	// the attr, or keep it as a func and receive jsonMode as param
+	jsonMode := true
+	if jsonMode {
+		// Ok so here we should generate a new ID and embbed it in `t`...
+		// first idea that comes to my mind. Create a json patch and apply it :)
```
sanderpick: May as well! :)

eventstore/model.go
```diff
@@ -214,16 +224,25 @@ func (t *Txn) Save(updated ...interface{}) error {
...
-		before := reflect.New(t.model.valueType.Elem()).Interface()
-		err = json.Unmarshal(beforeBytes, before)
-		if err != nil {
-			return err
+		var previous interface{}
+		if t.model.jsonMode {
+			// Instead of saving `before` as a Go Type of the before value of this instance
+			// lets just say that is exactly the json representation that is currently in
+			// the datastore.
+			previous = beforeBytes
```
sanderpick: 👍

eventstore/model.go
```diff
@@ -284,6 +303,14 @@ func (t *Txn) FindByID(id core.EntityID, v interface{}) error {
 	if err != nil {
 		return err
 	}
+	if t.model.jsonMode {
+		str := string(bytes)
+		rflStr := reflect.ValueOf(str)
+		reflV := reflect.ValueOf(v)
+		reflV.Elem().Set(rflStr)
+
+		return nil
```
sanderpick: Not returning `bytes`?

eventstore/model.go
```diff
@@ -317,24 +344,67 @@ func (t *Txn) Discard() {
...
+	// JSONMODE: assumes true. make `getEntity` a method of `Model` so maybe can access
+	// the attr, or keep it as a func and receive jsonMode as param
+	jsonMode := true
+	if jsonMode {
+		partialStruct := &structWithID{ID: "YOUDONTHAVEID"}
+		strJson := t.(*string)
+		if err := json.Unmarshal([]byte(*strJson), partialStruct); err != nil {
+			panic(err)
+		}
+		if partialStruct.ID == "YOUDONTHAVEID" {
```
sanderpick: :)

eventstore/model.go
jsign: This method returns an `error`. It populates `v` with the searched instance by ID.
Maybe the code confuses a bit since it returns the err of `json.Marshal` for the non-JSONMode case?

jsign:
> This is great. Best of both worlds.

Yeah, I like that this new feature *avoid* work rather than add. The remaining work in other APIs I think is re-applying the same idea again. Also, avoid harcoding jsonmode and add some opts things.

Besides this hacky proposal... having this kind of `if jsonMode` may uncover that some extra abstraction can make things more elegant... maybe the Store should really be json mode on, and the Go Type is an extra wrap layer with refect.  That maybe it can be an idea to tidy things up in the future, don't know if may worth now.

eventstore/model.go
sanderpick: Oh! Yeah makes sense.

sanderpick:
SGTM.

eventstore/model.go
```diff
@@ -32,7 +34,7 @@ var (
 // Model contains instances of a schema, and provides operations
 // for creating, updating, deleting, and quering them.
 type Model struct {
-	schema       *jsonschema.Schema
+	name         string
```
jsign: Now instead of using the json schema `Ref` as the identifier of the model, we use the `name` that the user chose. Seems to simplify things and makes more sense since the client chose that name for a reason.

eventstore/model.go
```diff
@@ -317,24 +350,48 @@ func (t *Txn) Discard() {
...
-	if !v.IsValid() || v.Type() != reflect.TypeOf(core.EmptyEntityID) {
-		panic("invalid instance: doesn't have EntityID attribute")
+func getEntityID(t interface{}, jsonMode bool) core.EntityID {
+	if jsonMode {
+		partial := &struct{ ID *string }{}
+		if err := json.Unmarshal([]byte(*(t.(*string))), partial); err != nil {
+			log.Fatalf("error when unmarshaling json instance: %v", err)
+		}
+		if partial.ID == nil {
+			log.Fatal("invalid instance: doesn't have an ID attribute")
```
jsign: Much better than the fixed string :smile:

eventstore/model.go
```diff
@@ -317,24 +350,48 @@ func (t *Txn) Discard() {
...
+		}
+		if *partial.ID != "" && !core.IsValidEntityID(*partial.ID) {
```
jsign: If has a value, we do an extra check that is valid (i.e. a valid uuid)

eventstore/model.go
```diff
@@ -317,24 +350,48 @@ func (t *Txn) Discard() {
...
-func setNewEntityID(t interface{}) core.EntityID {
-	v := reflect.ValueOf(t)
-	if v.Type().Kind() != reflect.Ptr {
-		v = reflect.New(reflect.TypeOf(v))
-	}
+func setNewEntityID(t interface{}, jsonMode bool) core.EntityID {
 	newID := core.NewEntityID()
-	v.Elem().FieldByName(idFieldName).Set(reflect.ValueOf(newID))
+	if jsonMode {
+		patchedValue, err := jsonpatch.MergePatch([]byte(*(t.(*string))), []byte(fmt.Sprintf(`{"ID": %q}`, newID.String())))
```
jsign: :sunglasses:

eventstore/store.go
```diff
@@ -176,7 +184,8 @@ func (s *Store) Threadservice() threadservice.Threadservice {
 func (s *Store) Register(name string, defaultInstance interface{}) (*Model, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	if s.alreadyRegistered(defaultInstance) {
+
+	if _, ok := s.modelNames[name]; ok {
```
jsign: I now added a set-map of model registered `Name`s, so we use that for checking if the model is already registered. Before, the GoTyped map was used... but in JSON Mode no Go Types are registered so that map will be always empty.

I guess this would be the right thing to do in advance: use the `Name` that the client chose for the Model as their identity. (The Go Type isn't really an identity, just a representation).

exe/local_eventstore/json-books/main.go
```diff
@@ -0,0 +1,126 @@
...
+		if !exists {
+			panic("instance shouldn't exist")
+		}
+	}
+}
+
+func createJsonModeMemStore() (*es.Store, func()) {
+	dir, err := ioutil.TempDir("", "")
+	checkErr(err)
+	s, err := es.NewStore(es.WithRepoPath(dir), es.WithJsonMode(true))
```
jsign: New `.WithJsonMode(..)`, where default is `false`.

go.mod
```diff
@@ -54,3 +54,5 @@ require (
 	google.golang.org/grpc v1.24.0
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 )
+
+replace github.com/textileio/go-textile-core v0.0.1 => ../go-textile-core
```
jsign: Will edit when [go-textile-core #30](https://github.com/textileio/go-textile-core/pull/30) is merged

exe/local_eventstore/json-books/main.go
sanderpick: Slick!

eventstore/model.go
sanderpick: 🤯

eventstore/store.go
```diff
@@ -186,6 +195,20 @@ func (s *Store) Register(name string, defaultInstance interface{}) (*Model, erro
 
 	m := newModel(name, defaultInstance, s)
 	s.models[m.valueType] = m
+	s.modelNames[name] = struct{}{}
+	s.dispatcher.Register(m)
+	return m, nil
+}
+
+func (s *Store) RegisterSchema(name string, schema string) (*Model, error) {
```
sanderpick: 🔥

jsign:
Seems that tests are failing due to parallel tests that try to bind to 5050 port and those sorts of things.

I'll merge this to unblock you and aaron, and then jump to try to resolve some issues we generated related to this thing of extra config opts. Main goal in that PR: unit-tests should stop failing (at least for this reason)

PR #106, opened 2019-11-20
store: multiple issue fixes

### LINKED ISSUE #86: store: bootstrap peers options
Per @sanderpick  [idea](https://github.com/textileio/go-textile-threads/pull/74#discussion_r346535281), add options to add custom bootstrap peers.

### LINKED ISSUE #87: threadservice: add method to build a default instance
Currently, Store has a default unexported method to do it in order to make things simple, but as @sanderpick mentioned may be more useful to make it public and remove it from Store `Options`. 

Could also be leveraged by users that want to build their own thing on top of a easy-default-threadservice too.

Edit: add port configuration [comment](https://github.com/textileio/go-textile-threads/pull/74#issuecomment-554108570)

### LINKED ISSUE #90: store: leverage utils and/or add new methods
Motivated by [review comment](https://github.com/textileio/go-textile-threads/pull/74#discussion_r346534598), [comment](https://github.com/textileio/go-textile-threads/pull/74#discussion_r346534690) and [comment](https://github.com/textileio/go-textile-threads/pull/74#discussion_r346558948).

The package `utils` has methods that can be leveraged to actions that are repeated in multiple packages. Maybe even adding some extra reasonable things.

### LINKED ISSUE #107: store: default theadservice with configurable proxy port
To avoid error logs with unbindable 5050 port.

jsign:
The `DefaultThreadService` lives in the `eventstore` package.
Originally, I moved it to `util` but we have a circular dependency (utils, threads, threadstore, utils). Trying on `threads`, gets the same circular dependency (threads, threadstore, threads).

Some general points:
- reduced log levels to errors in tests
- Fixes #86 
- Fixes #87 
- Fixes #90
- Fixes #107
- makes tests pass

Besides implementation details, may worth taking a look again to `TestE2EWithThreads` to see how it impacted end user experience, in short:
1. Create `DefaultThreadService()` with an explicit `repoPath`, and optionals: `ListenPort` for listening IPFS Lite, `ProxyPort` if want to avoid default `5050`, and `Debug` for enabling debug lefel.
2. Create the `Store` using that `Threadservice`.
3. Eventually, client *is responsible* for closing both `Threadservice` (the created default), and `Store`.
4. The client is also responsible for calling `Boostrap(..)` on the default `Threadservice`. Note that `Boostrap()` isn't part of `Threadservice` interface, but an [augmented one](https://github.com/textileio/go-textile-threads/blob/d253be406e70687077fec794d86221ab52bb7425/eventstore/util.go#L30) (only default instances have this method).

### DIFF
--- a/eventstore/util.go
+++ b/eventstore/util.go
@@ -32,7 +32,14 @@ type ThreadserviceBoostrapper interface {
 	Bootstrap(addrs []peer.AddrInfo)
 }
 
-func DefaultThreadservice(listenPort int, repoPath string, debug bool) (ThreadserviceBoostrapper, error) {
+func DefaultThreadservice(repoPath string, opts ...Option) (ThreadserviceBoostrapper, error) {
+	config := &Config{ProxyPort: 5050}
+	for _, opt := range opts {
+		if err := opt(config); err != nil {
+			return nil, err
+		}
+	}
+
 	ipfsLitePath := filepath.Join(repoPath, defaultIpfsLitePath)
 	if err := os.MkdirAll(ipfsLitePath, os.ModePerm); err != nil {
 		return nil, err
@@ -49,7 +56,7 @@ func DefaultThreadservice(listenPort int, repoPath string, debug bool) (Threadse
 		cancel()
 		return nil, err
 	}
-	listen, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", listenPort))
+	listen, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", config.ListenPort))
 	if err != nil {
 		ds.Close()
 		cancel()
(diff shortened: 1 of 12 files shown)


threads.go
```diff
@@ -708,8 +707,12 @@ func (t *threads) startPulling() {
...
 	}
 	timer := time.NewTimer(InitialPullInterval)
-	<-timer.C
-	pull()
+	select {
+	case <-timer.C:
+		pull()
+	case <-t.ctx.Done():
+		return
+	}
```
jsign: This bug took me quite a bit to find. (wasn't related with any created fix).

The problem here was that if the client closes the `Threadservice` before `InitialPullInterval`, the `Close()` method of `Threadservice` will close the datastore, and will cause `pull()` internal work to panic strangely at the badger level (since the database was closed!)

eventstore/util.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func Debug(enabled bool) Option {
+	return func(c *Config) error {
+		c.Debug = enabled
+		return nil
+	}
+}
+
+type tservBoostraper struct {
```
sanderpick: nit: `Bootstaper`/`Bootstrapper`

threads.go
sanderpick: Ooo, sneaky one!

PR #108, opened 2019-11-20
api: initial poc with grpc

### LINKED ISSUE #112: Create a store manager
Users will need to be able to get there hands on stores that they have created over the gRPC API.
- A manager can allow all stores to use the same datastore and thread service.
- A manager can load stores into memory from the datastore using a key prefix

sanderpick:
Closes #112 

Integrates an API that serves two purposes:
1. A _remote_ API for multiple browser (and possible simple mobile) clients, aka, a client _Hub_
2. A _local_ API for command-line, shell, and native framework clients

See forthcoming comments...


api/pb/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@textile/threads-api-pb",
```
sanderpick: In order to run the browser example locally, 
```
cd api/pb
npm install
npm link
```

api/server.go
```diff
@@ -0,0 +1,125 @@
+package api
```
sanderpick: Since the API is not required to run any part of threads, it can live in its own package.

api/server.go
```diff
@@ -0,0 +1,125 @@
...
+		pb.RegisterAPIServer(s.rpc, s.service)
+		s.rpc.Serve(listener)
+	}()
+
+	webrpc := grpcweb.WrapServer(
+		s.rpc,
+		grpcweb.WithOriginFunc(func(origin string) bool {
+			return true
+		}),
+		grpcweb.WithWebsockets(true),
```
sanderpick: WebSockets works well, but possible only extra useful for client-side streaming. Luckily, you can set the transport on a per-request basis.

api/server.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	go func() {
+		pb.RegisterAPIServer(s.rpc, s.service)
+		s.rpc.Serve(listener)
+	}()
+
+	webrpc := grpcweb.WrapServer(
+		s.rpc,
+		grpcweb.WithOriginFunc(func(origin string) bool {
+			return true
```
sanderpick: At this point, anything goes!

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+	es "github.com/textileio/go-textile-threads/eventstore"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type service struct {
+	datastore ds.Datastore
+	threads   tserv.Threadservice
+
+	stores map[string]*es.Store
```
sanderpick: Didn't put much thought into this... just a dirty way to access a created store.

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (s *service) NewStore(ctx context.Context, req *pb.NewStoreRequest) (*pb.NewStoreReply, error) {
+	log.Debugf("received new store request")
+
+	store, err := es.NewStore(
+		es.WithDatastore(s.datastore),
+		es.WithThreadservice(s.threads),
+		es.WithJsonMode(true),
+		es.WithDebug(true))
```
sanderpick: These options feel great, kudos

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	id, err := uuid.NewRandom()
```
sanderpick: Silly way to enable later access to a store before it's been started.

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+
+	reply := &pb.ModelCreateReply{
+		EntityIDs: make([]string, len(values)),
+	}
+	for i, v := range values {
+		id, err := getEntityID(v)
+		if err != nil {
+			return nil, err
+		}
+		reply.EntityIDs[i] = id.String()
```
sanderpick: Maybe it makes more sense to return the entire entity

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return reply, nil
+}
+
+func getEntityID(t interface{}) (core.EntityID, error) {
```
sanderpick: Just hi-jacked this from the store for now.

eventstore/options.go
```diff
@@ -147,6 +147,13 @@ func WithDebug(enable bool) StoreOption {
 	}
 }
 
+func WithDatastore(ds ds.Datastore) StoreOption {
```
sanderpick: I needed this to leverage an existing store (the one created for the threadservice)

eventstore/store.go
```diff
@@ -56,7 +56,7 @@ type Store struct {
 
 	lock       sync.RWMutex
 	models     map[reflect.Type]*Model
-	modelNames map[string]struct{}
+	modelNames map[string]*Model
```
sanderpick: If we do this, it means a registered model can be retrieved by name. Is there any reason to keep the `models` field?

eventstore/store.go
```diff
@@ -209,10 +212,15 @@ func (s *Store) RegisterSchema(name string, schema string) (*Model, error) {
 	}
 
 	m := newModelFromSchema(name, schema, s)
+	s.modelNames[name] = m
```
sanderpick: Seems like this was missing...

exe/daemon/main.go
```diff
@@ -17,16 +19,24 @@ func main() {
...
-	_, cancel, ds, h, dht, api := util.Build(*repo, *port, *proxyAddr, true)
+	_, cancel, ds, h, dht, ts := util.Build(*repo, *port, *proxyAddr, true)
 
 	defer cancel()
 	defer dht.Close()
-	defer api.Close()
+	defer ts.Close()
 	defer ds.Close()
 
+	server, err := api.NewServer(context.Background(), ds, ts, api.Options{
```
sanderpick: Hijacking the daemon to run the API server.

api/server.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	s := &Server{
+		rpc: grpc.NewServer(),
+		service: &service{
+			datastore: ds,
+			threads:   ts,
+			stores:    make(map[string]*es.Store), // @todo how to reload stores on start?
```
jsign: Good question. Maybe inventing some `StoreManager` with a simple badger to stor store IDs and they repo paths.

api/server.go
jsign: Sounds good for a future stream-service to hear that the underlying `Store` state changed (using the `Store` listener).

api/server.go
```diff
@@ -0,0 +1,125 @@
...
+	go func() {
+		errc <- s.proxy.ListenAndServe()
+		close(errc)
+	}()
+	go func() {
+		for err := range errc {
+			if err != nil && err != http.ErrServerClosed {
+				log.Errorf("proxy error: %s", err)
+			}
+		}
```
jsign: Shouldn't break the `for` loop if `err == http.ErrServerClosed`?

api/service.go
```diff
@@ -0,0 +1,111 @@
...
+	threads   tserv.Threadservice
+
+	stores map[string]*es.Store
+}
+
+func (s *service) NewStore(ctx context.Context, req *pb.NewStoreRequest) (*pb.NewStoreReply, error) {
+	log.Debugf("received new store request")
+
+	store, err := es.NewStore(
+		es.WithDatastore(s.datastore),
```
jsign: Reminder: Datastore should be unique per store.

api/service.go
jsign: Yeah, I was thinking the same.

eventstore/store.go
jsign: Maybe not. At first I checked the model re-registration check with the Go Type, and thus using the `models` map. Now that identities are their name, isn't necessary.

exe/daemon/main.go
jsign: Same datastore for possibly multiple stores. I think this won't work correctly. (See above 'Reminder:' comment)
Maybe we could make the store ID a first class citizen and use it as a prefix for using the datastore. This would allow sharing a single datastore for multiple Stores.

jsign: Other option is using [keytransform](https://github.com/ipfs/go-datastore/tree/master/keytransform) to build a wrapper with the ID prefix, and make it transparent for the Store.

asutula:
Looks cool so far. For sure still need to dig into the details, but looking good!

asutula:
Wonder how authentication fits into this? Seems that some auth token would somehow scope a client's access to a set of `Store`s

api/server.go
sanderpick: Yep, but you can also use other transports for server-side pushing. WS are required for "client-side" streaming.

eventstore/store.go
sanderpick: Ok, cool. I'll make that change.

api/server.go
sanderpick: Oop, copy pasta from `threads`, probably should change that too.

exe/daemon/main.go
sanderpick: OK cool. I was definitely confused about that. I think we'll want to leverage the same repo across stores if at all possible. I'm imagining a client hub can be serving many stores. I can take a crack at a manager.

sanderpick:
Good question @asutula. I was thinking the same, leverage a JWT scope somehow. I like the per-store basis idea.

sanderpick:
I pushed a bunch of changes here. The biggest one is an initial store `Manager` that is used by the grpc service. I'm not sure if we want to figure out how to "persist" the models so they can be re-hydrated by the manager as well.

I can make some inline comments in the morning. Gotta go hit the sack now.

api/server.go
```diff
@@ -0,0 +1,133 @@
...
+		es.WithRepoPath(conf.RepoPath),
+		es.WithDebug(true))
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	s := &Server{
+		rpc:     grpc.NewServer(),
+		service: &service{manager: manager},
```
sanderpick: RPC service takes the store manager.

api/server.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (s *Server) Close() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	if err := s.proxy.Shutdown(ctx); err != nil {
+		log.Errorf("error shutting down proxy: %s", err)
+	}
+
+	s.rpc.GracefulStop()
+	s.service.manager.Close()
```
sanderpick: Closing the manager closes its datastore. Stores that are part of the manager do not close their datastores.

daemon/main.go
```diff
@@ -0,0 +1,52 @@
...
+	repo := flag.String("repo", ".threads", "repo location")
+	listenPort := flag.Int("port", 4006, "host port")
+	proxyPort := flag.Int("proxyPort", 5050, "grpc proxy port")
+	flag.Parse()
+
+	if err := logging.SetLogLevel("daemon", "debug"); err != nil {
+		log.Fatal(err)
+	}
+
+	ts, err := es.DefaultThreadservice(
```
sanderpick: Migrated to this util method.

eventstore/manager.go
```diff
@@ -0,0 +1,108 @@
...
+var (
+	dsStoreManagerBaseKey = ds.NewKey("/manager")
+)
+
+type Manager struct {
+	io.Closer
+
+	config *StoreConfig
+
+	threadservice threadservice.Threadservice
```
sanderpick: The manager needs a threadservice but it does _not_ deal with closing it. In practice, the daemon uses two datastores: one for the manager and one for the thread service.

eventstore/manager.go
```diff
@@ -0,0 +1,111 @@
...
+		idStr := ds.RawKey(res.Key).Parent().Parent().Parent().Parent().Name() // reaching for the stars here
+		id, err := uuid.Parse(idStr)
+		if err != nil {
+			return nil, err
+		}
+		s, err := newStore(m.threadservice, m.config)
+		if err != nil {
+			return nil, err
+		}
+		s.datastore = kt.Wrap(s.datastore, kt.PrefixTransform{
```
jsign: Maybe direct assignment of an unexported field may break some assumptions (not sure now, but may in the future).
If `New` fires a goroutine that does something before this line, would write to other datastore that would be leaked?

Maybe just do a copy of `m.config`, edit the `datastore` field, and then call `newStore(..)` with that cloned-modified config.

eventstore/manager.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if config.Datastore == nil {
+		datastore, err := newDefaultDatastore(config.RepoPath)
+		if err != nil {
+			return nil, err
+		}
+		config.Datastore = datastore
+	}
+
+	// @todo: handle debug here or in store?
```
sanderpick: The way it is currently, it won't cause any problems, but is a not so nice as the log level will be set for every new store. We could leverage the keytransform check utilized in Store.Close to determine if this is a managed store, or a standalone store.

eventstore/manager.go
```diff
@@ -0,0 +1,114 @@
...
+			return nil, err
+		}
+		s, err := newStore(m.threadservice, m.config)
+		if err != nil {
+			return nil, err
+		}
+		s.datastore = kt.Wrap(s.datastore, kt.PrefixTransform{
+			Prefix: dsStoreManagerBaseKey.ChildString(id.String()),
+		})
+		if err = s.Start(); err != nil {
```
sanderpick: Start reloaded stores. I'm not sure about this. It will cause unexpected errors if the API is restarted when the API user is min-model registration.

eventstore/manager.go
```diff
@@ -0,0 +1,114 @@
...
+func (m *Manager) NewStore() (id uuid.UUID, store *Store, err error) {
+	id, err = uuid.NewRandom()
+	if err != nil {
+		return
+	}
+	store, err = newStore(m.threadservice, m.config)
+	if err != nil {
+		return
+	}
+	store.datastore = kt.Wrap(store.datastore, kt.PrefixTransform{
```
sanderpick: Just forcing the internally chosen store to be wrapped.

eventstore/manager.go
```diff
@@ -0,0 +1,114 @@
...
+	return m, nil
+}
+
+// NewStore creates a new store and prefix its datastore with base key.
+func (m *Manager) NewStore() (id uuid.UUID, store *Store, err error) {
+	id, err = uuid.NewRandom()
+	if err != nil {
+		return
+	}
+	store, err = newStore(m.threadservice, m.config)
```
sanderpick: All managed stores are created with the same config, which contains the same datastore and other settings.

eventstore/store.go
```diff
@@ -68,8 +68,14 @@ type Store struct {
...
 	for _, opt := range opts {
-		opt(config)
+		if err := opt(config); err != nil {
+			return nil, err
+		}
 	}
+	return newStore(ts, config)
+}
+
+func newStore(ts threadservice.Threadservice, config *StoreConfig) (*Store, error) {
```
sanderpick: Helper method to allow creating a store with the same config.

eventstore/store.go
```diff
@@ -243,7 +253,9 @@ func (s *Store) Close() {
 	s.cancel()
 	s.localEventsBus.bus.Discard()
 	s.stateChanged.bus.Discard()
-	s.datastore.Close()
+	if _, ok := s.datastore.(kt.KeyTransform); !ok {
```
sanderpick: Check if datastore is wrapped...

shell/main.go
```diff
@@ -73,46 +69,37 @@ func (n *notifee) HandlePeerFound(p peer.AddrInfo) {
 
 func main() {
```
sanderpick: This "shell" can eventually run an API server like the daemon, and provide a kind of model playground. In the meantime, I moved it to the root dir (along with daemon) and left the chat app in there.

eventstore/manager.go
jsign: What do you mean by _min-model_ registration?

eventstore/manager.go
jsign: Related with same comment I did above.

eventstore/manager.go
sanderpick: oop, typo - "mid-model registration" as in, when they are building a store

jsign: Oh, yeah maybe we should think about this border cases and how they impact corruption (unrecoverable?)

sanderpick:
Thanks for the review, @jsign! Here are some general notes below regarding this work and next steps:

1. I realize we're layering on a kind of "server" use-case to the stores, and exposing the full API to clients (other than local) may not be ideal. However, I'm finding it hard to determine where the line should be, which point to just needing to build some apps with it to get a better feel. 
2. A good example of the above fuzziness is model registration: Should _any_ client be able to register models? This is analogous to creating a SQL table, so probably not.
3. I believe we'll need some way to "reload" models into each store (and register them with the dispatcher). In the meantime, the client can just force registration whenever they start up. Maybe this is how it should always be? Hard to say.
4. No auth still, but there's an [issue tracking the JWT work](https://github.com/textileio/go-textile-threads/issues/54).

Next steps: 
- @sanderpick to add state change subscription in a separate PR
- @asutula to take over flushing out the remaining APIs in a separate PR

eventstore/manager.go
sanderpick: yep, I'll create an issue for this

eventstore/manager.go
sanderpick: nice, good call.. i'll make that change

PR #110, opened 2019-11-21
Imperative Transactions

asutula:
Adds api for imperatively creating, using, and committing transactions.

I started playing with how we might implement transactions over the gRPC api, and I'm feeling pretty confident that something like this will be necessary. Thinking is that the gRPC service will need to maintain a map of active transactions (zero or one per `Store`) and if a transaction exists for a `Store`, run all subsequent requests for that `Store` against its transaction. Once `EndTransaction` is called for a `Store`, that `Store`'s transaction is removed from the map and subsequent calls return to normal (using implicit transactions).

For sure a reach with my current grasp on Go, but hopefully this gets the idea across.

### DIFF
--- a/eventstore/model.go
+++ b/eventstore/model.go
@@ -66,11 +66,31 @@ func newModelFromSchema(name string, schema string, s *Store) *Model {
 	return m
 }
 
+// StartReadTxn starts a read transation and returns it
+func (m *Model) StartReadTxn() *Txn {
+	return m.store.startReadTxn(m)
+}
+
+// EndReadTxn ends the specified read transaction
+func (m *Model) EndReadTxn(txn *Txn) {
+	m.store.endReadTxn(txn)
+}
+
 // ReadTxn creates an explicit readonly transaction. Any operation
 // that tries to mutate an instance of the model will ErrReadonlyTx.
 // Provides serializable isolation gurantees.
 func (m *Model) ReadTxn(f func(txn *Txn) error) error {
 	return m.store.readTxn(m, f)
+}
+
+// StartWriteTxn starts a write transation and returns it
+func (m *Model) StartWriteTxn() *Txn {
+	return m.store.startWriteTxn(m)
+}
+
+// EndWriteTxn ends the specified write transaction, optionally commiting its changes
+func (m *Model) EndWriteTxn(txn *Txn, commit bool) error {
+	return m.store.endWriteTxn(txn, commit)
 }
 
 // WriteTxn creates an explicit write transaction. Provides
--- a/eventstore/model_test.go
+++ b/eventstore/model_test.go
@@ -89,6 +89,15 @@ func TestCreateInstance(t *testing.T) {
 			checkErr(t, err)
 			assertPersonInModel(t, model, newPerson)
 		})
+		t.Run("WithImperativeTx", func(t *testing.T) {
+			newPerson := &Person{Name: "Foo", Age: 42}
+			txn := model.StartWriteTxn()
+			err := txn.Create(newPerson)
+			checkErr(t, err)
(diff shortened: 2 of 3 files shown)


asutula:
Idea is you can do this:
```
newPerson := &Person{Name: "Foo", Age: 42}
txn := model.StartWriteTxn()
err := txn.Create(newPerson)
shouldCommit = err == nil
_ = model.EndWriteTxn(txn, shouldCommit)
```

sanderpick:
Nice! Your Go skills are looking good :) 

It would be cool if the txn could live in the client-stream, such that you wouldn't need to maintain the state in some central map, keyed by an ID or something similar. That would be in line with your comment about bi-di streams over WS.

jsign:
If we eventually have a stream-based API, I think is possible to also use the original func-based one: like, generate one `func(txn *Txn) error` per stream, and run the message-loop of the stream-api inside that func. When the stream is closed, Discard/Commit txn. 

To be more concrete (apart from whatever txn style we like): We should tie the underlying txn (in any of their styles) with some scoped thing that depends on the client. I mean, making _impossible_ to leak an open transaction if the client closes the connection, or whatever. 

An open txn is blocking not only other local/client txn (which isn't that bad), but also allowing to sync new changes from the external world that could pile up. Anyway, that's job for the next steps... :)

asutula:
That makes sense @jsign. It's easier to wrap my head around the grpc client to server part (with the server running the transaction), but I'm having a harder time understanding what the native (obj c/java/swift) to grpc client part would look like... Somehow, we have to allow the user to say "I want to make these actions within a transaction", and somehow map each action to a grpc request object. It's not clear to me how to do that. Ideas?

asutula:
When the user wants to use an implicit transaction, they would just call an sdk method directly (`CreateModel()`, for example), but in the case of a transaction we have to somehow compile a list of methods to call on the server side (as opposed to calling a series of grpc client methods).

sanderpick:
I think in either case, the context of the transaction can be "held" by the server, while the client streams calls, right?

jsign:
Well, we definitely should think about things that run on transactions as being interactive, since batching things and sending them only works for batch things (like creating multiple things). But other more useful workflows such as open a transaction, do some query, and make an insert considering that query result.

Native to Grpc: I'd imagine
```
(On the native side)
1. txn := sdk.OpenTxn()
2. res = txn.Query...
3. txn.Create( something)
4. txn.Delete(..)
5. txn.Commit()
```
Wher in 1., the sdk created a stream-grpc call that holds alive until commit. That `t` has the underlying grpc connection which sends and receives the 2., 3., 4. calls back and forth.
On the Go side, that stream-grpc call is looping inside an explicit txn, receiving the 2., 3., 4. things and replying appropriately.

Makes sense?. I might be missing something since as you say is somewhat hard to imagine.

asutula:
I think you are both making sense, it's just really hard to imagine exactly how the implementation will work. Sounds like there will be some bi-directional streaming as well as a more robust native sdk class structure (adding a `Txn` class, for example). Will be good to start making a client and experimenting.

asutula:
I suppose we could make it work like the Go version works, where everything is done in a transaction, and we provide helper methods for implicit transactions. That would actually never use the implicit methods on the server.

asutula:
The more I think about the bi-directional streaming, receiving and sending inside the server's Txn function, the more I like it. Makes tons of sense. Gonna try to get the POC on that built. Thanks for the vision @jsign

asutula:
Will leave this open for now, but if things go as planned with the bi-directional streaming, I'll just close it.

asutula:
Not needed anymore, was able to implement transactions using bi-directional grpc streams

PR #118, opened 2019-11-21
store: persist json schemas

### LINKED ISSUE #117: Persist models in JSON Mode
When registering models in JSON mode, the models can be re-registered when the store reloads, say, no startup.

sanderpick:
Closes #117 

I've gone back and forth on whether or not this is needed. I think it will become more clear soon.

### DIFF
--- a/eventstore/store.go
+++ b/eventstore/store.go
@@ -12,6 +12,7 @@ import (
 
 	ds "github.com/ipfs/go-datastore"
 	kt "github.com/ipfs/go-datastore/keytransform"
+	"github.com/ipfs/go-datastore/query"
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/textileio/go-textile-core/broadcast"
@@ -38,6 +39,7 @@ var (
 	log             = logging.Logger("store")
 	dsStorePrefix   = ds.NewKey("/store")
 	dsStoreThreadID = dsStorePrefix.ChildString("threadid")
+	dsStoreSchemas  = dsStorePrefix.ChildString("schema")
 )
 
 // Store is the aggregate-root of events and state. External/remote events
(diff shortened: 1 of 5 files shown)


eventstore/storethread.go
```diff
@@ -82,12 +82,10 @@ func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
 				continue // Ignore our own events since Store already dispatches to Store reducers
 			}
 			ctx, cancel := context.WithTimeout(context.Background(), fetchEventTimeout)
-			defer cancel()
```
jsign: And again the same thing with `deferin` on a loop... Thanks!

eventstore/storethread.go
```diff
@@ -121,25 +120,25 @@ func (a *singleThreadAdapter) storeToThread(wg *sync.WaitGroup) {
...
-				log.Info("ending sending store local event to own thread since channel was closed for thread %s", a.threadID)
+				log.Infof("ending sending store local event to own thread since channel was closed for thread %s", a.threadID)
 				return
 			}
 			n, err := event.Node()
 			if err != nil {
 				log.Fatalf("error when generating node for own log for thread %s: %v", a.threadID, err)
 			}
 			ctx, cancel := context.WithTimeout(context.Background(), addRecordTimeout)
-			defer cancel()
```
jsign: :100:

eventstore/store.go
```diff
@@ -205,6 +233,16 @@ func (s *Store) RegisterSchema(name string, schema string) (*Model, error) {
 	}
```
jsign: Now I wonder if erroring on existing schema is the correct thing.
I guess we're not sure about how the client will know if do it or not... I guess by his own means of something similar. It can leverage `store.ThreadId()` or something that gives a clue (or maybe we can create a cheap func in the future).

Anyway, not mentioning to make something now... but just a flashing thought.

sanderpick: I was thinking the same.... hard to say!

PR #126, opened 2019-11-22
API: Listen for state changes

### LINKED ISSUE #113: Store API: Listen for state changes
Use a server-side stream method to listen for state changes.

sanderpick:
Closes #113 

Adds a server-side stream `Listen` method.

Currently, it requires that the caller provide an ```entityID``` such that the server can query for latest when the listener is triggered, and then pass it to the client. I _think_ eventually we'll want to enable the store listener to push the latest state, w/o requiring an explicitly query. This would allow the client to just listen for all entities changing.


jsign:
> I _think_ eventually we'll want to enable the store listener to push the latest state, w/o requiring an explicitly query. This would allow the client to just listen for all entities changing.

Yeah, totally. The current notification is just a _signal_ that *something* changed, which falls pretty short. I like the idea of giving different levels of granularity for notifications, like three options:
- Something changed in Store (what exists now). (Good for multi-model aggregates or views)
- Something changed in a Model. (Good for aggregates/views of models)
- Something changed in a particular instance of a Model. (Good for track particular things)

sanderpick:
Yes! That sounds great.

asutula:
Yea, that sounds perfect, can't wait for this :)

api/service.go
```diff
@@ -76,6 +77,38 @@ func (s *service) ModelCreate(ctx context.Context, req *pb.ModelCreateRequest) (
...
+	if err != nil {
+		return err
+	}
+
+	model := store.GetModel(req.ModelName)
+	if model == nil {
+		return status.Error(codes.NotFound, "model not found")
+	}
+
+	listener := store.StateChangeListen()
```
jsign: Add: `defer listener.Discard()`

sanderpick: nice, added!

PR #129, opened 2019-11-24
grpc impl

### LINKED ISSUE #122: Implement gRPC server
### RPC Methods
- [x] `NewStore`
- [x] `RegisterSchema`
- [ ] `Start`
- [ ] `StartFromAddress`
- [x] `ModelCreate`
- [ ] `ModelHas`
- [ ] `ModelSave`
- [ ] `ModelDelete`
- [ ] `ModelReadTxn`
- [ ] `ModelWriteTxn`
- [ ] `ModelFindByID`
- [ ] `ListenForStateChanges`?
- [ ] others?

### LINKED ISSUE #123: Implement gRPC client
This is useful as we're developing the server implementation and can also be used to build new `go-textile-threads` extensions such as the cli.

### LINKED ISSUE #124: Transactions over gRPC
This will be one of the more challenging parts of the gRPC api. Current thinking is that the solution will involve:

- a single gRPC method connection from client to server
- bi-directional streaming within that connection (one round trip for each transaction component)
- heavy use of proto `oneof` types so support different transaction request types
- the existing transaction api on the server should work as is

### LINKED ISSUE #125: Figure out query gRPC serialization
We have a nice and robust query api in Go, but we need to figure out how to send it over gRPC in a way that can be implemented in gRPC clients of various languages.

asutula:
This implements the event store api as a GRPC Server and also provides a Go client + tests. Still some issues to work through, but this is ready for review.

closes #122 
closes #123 
closes #124 
closes #125


api/service.go
```diff
@@ -77,6 +87,114 @@ func (s *service) ModelCreate(ctx context.Context, req *pb.ModelCreateRequest) (
...
+	}
+
+	model := store.GetModel(req.ModelName)
+	if model == nil {
+		return nil, status.Error(codes.NotFound, "model not found")
+	}
+
+	type Foo struct{}
+
+	return s.processFindByIDRequest(req, model.FindByID, &Foo{})
```
asutula: In this case, how do I create a default instance of the model type that is being queried to pass it into `model.FindByID`?

api/client/client.go
```diff
@@ -96,3 +96,12 @@ func (c *Client) ModelDelete(storeID, modelName string, entityIds []string) erro
...
 	return err
 }
+
+// ReadTransaction returns a read transaction that can be started and used and ended
+func (c *Client) ReadTransaction(storeID, modelName string) (*ReadTransaction, error) {
+	client, err := c.client.ReadTransaction(c.ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &ReadTransaction{client: client, storeID: storeID, modelName: modelName}, nil
```
asutula: the `client` here is bi-directional streaming inside of a single transaction on the server side

api/client/client.go
```diff
@@ -0,0 +1,207 @@
...
+	req := &pb.ListenRequest{
+		StoreID:   storeID,
+		ModelName: modelName,
+		EntityID:  entityID,
+	}
+	stream, err := c.client.Listen(c.ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	channel := make(chan interface{})
```
asutula: @sanderpick @jsign @carsonfarmer I don't really know how to deal with this situation since there are no generics in Go. This is the api where the user can subscribe to updates on a entity instance. You can see here that I'm making a `channel` of some unknown type (it could also be of `map[string]interface{}`), and then below I'm unmarshalling the json string into a `map[string]interface{}`. Not a great api. How can this be improved?

asutula: slight improvement passing in a dummy instance, but not sure

api/client/client.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	resp, err := c.client.ModelHas(c.ctx, req)
+	if err != nil {
+		return false, err
+	}
+	return resp.GetExists(), nil
+}
+
+// ModelFind finds records by query
+func (c *Client) ModelFind(storeID, modelName string, query es.JSONQuery) ([][]byte, error) {
```
asutula: Would be cool to not have to return `[][]byte`, but not clear to me if that is possible without generics.

api/client/client.go
```diff
@@ -0,0 +1,230 @@
...
+func (c *Client) WriteTransaction(storeID, modelName string) (*WriteTransaction, error) {
+	client, err := c.client.WriteTransaction(c.ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &WriteTransaction{client: client, storeID: storeID, modelName: modelName}, nil
+}
+
+// Listen provides an update whenever the specified model is updated
+func (c *Client) Listen(storeID, modelName, entityID string) (chan []byte, func(), error) {
```
asutula: Would be nice to have a typed channel to return here, but not sure if that is possible without generics.

api/client/client.go
```diff
@@ -0,0 +1,230 @@
...
+		for {
+			event, err := stream.Recv()
+			if err != nil {
+				break
+			}
+			bytes := []byte(event.GetEntity())
+			channel <- bytes
+		}
+	}()
+	return channel, cancel, nil
```
asutula: I'm having trouble testing this method. I'm suspicious that the code here is fine but it is actually the tear down of the grpc setup that causes the test to hang. I'll comment more on that in the test file, but if something look wrong above, speak up. 

Logic is:

1. Create a channel and cancelable context, immediately return the channel and the cancel function
2. Fork off a goroutine that starts a grpcserver streaming request
2a. If that fails, close the channel
3. Receive events on the stream from the server and forward them on the channel until an error happens
4. If the context cancel function is called, that errors out the server stream and the channel is closed

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+	_, clean := server(t)
+	defer clean()
+	client := client(t)
+
+	storeID, _ := client.NewStore()
+	err := client.RegisterSchema(storeID, modelName, schema)
+	checkErr(t, err)
+
+	// TODO: figure out how to test this
+	// client.StartFromAddress(storeId, <multiaddress>, <read key>, <follow key>)
```
asutula: Could use some help figuring out how to create the required inputs to test this.

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+		Ands: []es.JSONCriterion{
+			es.JSONCriterion{
+				FieldPath: "lastName",
+				Operation: es.Eq,
+				Value: es.JSONValue{
+					String: &person.LastName,
+				},
+			},
+		},
+	}
```
asutula: I haven't been able to find how to create a `JSONQuery` using a fluent API. Can someone point me to that?

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+	if err != nil {
+		t.Fatalf("failed to create read txn: %v", err)
+	}
+
+	defer func() {
+		err = txn.End()
+		if err != nil {
+			t.Fatalf("failed to end txn: %v", err)
+		}
+	}()
```
asutula: You _must_ always end a transaction, even if there is an error during a transaction. This seems reasonable because we don't want to assume the transaction should be ended automatically if there is an error during the transaction.

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+		t.Fatalf("failed to save in write txn: %v", err)
+	}
+
+	err = txn.Delete(existingPerson.ID)
+	if err != nil {
+		t.Fatalf("failed to delete in write txn: %v", err)
+	}
+}
+
+func TestListen(t *testing.T) {
```
asutula: This test hangs forever and times out.

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+	_, clean := server(t)
+	defer clean()
```
asutula: If I disable this call to `clean()`, it works fine.

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+	} else {
+		p := &Person{}
+		if err := json.Unmarshal(val, p); err != nil {
+			t.Fatalf("failed to unmarshal listen result: %v", err)
+		}
+		if p.Age != 40 {
+			t.Fatalf("expected listen result age = 40 but got: %v", p.Age)
+		}
+	}
+	discard()
```
asutula: We call the listener discard method here which should cancel the underlying gRPC request that is consuming the server stream. Even though I think that is working, it seems that the `defer clean()` call is hanging up for some reason, possibly related to the gRPC listen request status or something. It's hard to figure out, could use some help here.

api/client/client_test.go
```diff
@@ -0,0 +1,487 @@
...
+	checkErr(t, err)
+	return client
+}
+
+func checkErr(t *testing.T, err error) {
+	t.Helper()
+	if err != nil {
+		t.Fatal(err)
+	}
+}
```
asutula: Strange thing... I can run each of these tests individually and the pass (other than `TestListen`), but if I comment out `TestListen` and try to run the package tests, it hangs and times out. No idea what could be causing that.

api/client/client.go
jsign: This method is all for your best interests, so I can change it to `[][]bytes` if you prefer. In fact, it would be even better since returing `[]string` will allocate-extra memory.
If I have your confirmation, I can change that.

api/client/write.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil
+	default:
+		return fmt.Errorf("WriteTransactionReply.Option has unexpected type %T", x)
+	}
+}
+
+// End ends the active transaction
+func (t *WriteTransaction) End() error {
+	return t.client.CloseSend()
+}
```
asutula: Lots of similar code between the read and write transaction files, but it's just different enough that it's not clear to me how to consolidate. If you have any ideas, please let me know.

api/service.go
```diff
@@ -110,6 +336,83 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
 	return nil
 }
 
+func (s *service) processCreateRequest(req *pb.ModelCreateRequest, createFunc func(...interface{}) error) (*pb.ModelCreateReply, error) {
```
asutula: These "process..." methods are to consolidate code that is used in both the contexts of implicit transaction calls and explicit transactions. The function to call to actually execute the call is passed in (`model.Create` vs `txn.Create`, for example)

api/client/client.go
jsign: We can ask the client to provide a _default instance type_ of this model, and with some reflect magic unmarshal objects of that type. But that would be a `chan interface{}` anyways.

api/client/client.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+	return &WriteTransaction{client: client, storeID: storeID, modelName: modelName}, nil
+}
+
+// Listen provides an update whenever the specified model is updated
+func (c *Client) Listen(storeID, modelName, entityID string) (chan []byte, func(), error) {
+	channel := make(chan []byte)
+	ctx, cancel := context.WithCancel(c.ctx)
+	go func() {
```
jsign: All `err` here are dismissed. 
I see two cases:
- an `err` is returned by `.Recv` since the client canceled the client. So that's a kind of correct-workflow error.
- an `err` because something wrong happened (closed connection, the server died, etc).

I guess some type of checking can be done to detect both cases, and if in the second case:
1. do some `log.Error` to keep track that something bad happened
2. Add some extra documentation in the method like:
```
 // Listen provides an update whenever the specified model is updated
// If the returned channel is closed without the cancel func being called, 
// should be interpreted as an error in the communication-channel with 
// the server.
```
since on any kind of network connection problem, the goroutine finishes, and the returned channel is closed.

Another idiomatic option, in this case, is to return an extra `err chan error`, to allow the client to have visibility that something wrong happened (and actually, what was the error).

Not sure what you think @asutula  @sanderpick .

api/client/client_test.go
jsign: This `StarFromAddress` is to join the Thread (Store) created by another server.
So, say you have Server1 with Client1, and Server2 with Client2.

Client1/Server1 will `Start()` as usual. Then Client2 can join the underlying Thread that Client1/Server1 created, so to operate in the same `Store`. In order to do that, after creating Client1/Server` with `Start()`, we should get the multiaddr and keys from Server1, and use that info to `StartFromAddress`. After that, both clients should be operating on the same _shared Store_.

api/client/client_test.go
jsign: Sure, should be something analogous to [this](https://github.com/textileio/go-textile-threads/blob/master/eventstore/query.go#L30) and [this](https://github.com/textileio/go-textile-threads/blob/master/eventstore/criterion.go#L57)

api/client/client.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+// Client provides the client api
+type Client struct {
+	client pb.APIClient
+	ctx    context.Context
+}
+
+// NewClient starts the client
+func NewClient(host string, port int) (*Client, error) {
```
jsign: Shouldn't the `Client` type have a `Close` method that closes the underlying gRPC (and tcp connection)?
Maybe it could also err if there're open txns, or force them to close too and do that type of cleaning stuff.

api/client/client_test.go
jsign: That's a clean close. I'm wonder if we can think about good practices of using these things, since it may feel a bit verbose that kind of defers. As a client, I'd generally ignore if ending the readtx had an error since sounds quite safe.

Also, we could make `txn.End()` be a return of `ReadTransaction` to insist on the fact that this is a task that should be done explicitly. (since maybe someone doesn't notice there is an `End()` method). Just playing around with an idea, not sure.

api/client/client_test.go
jsign: I'll take a look after the review.

api/service.go
```diff
@@ -93,21 +319,108 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
...
-	for range listener.Channel() {
-		err := model.ReadTxn(func(txn *es.Txn) error {
-			var res string
-			if err := txn.FindByID(core.EntityID(req.EntityID), &res); err != nil {
-				return err
+
+	for {
+		select {
+		case _ = <-server.Context().Done():
+			return nil
```
asutula: If the context is canceled (because the client canceled the request), we need to return

api/service.go
```diff
@@ -93,21 +319,108 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
...
+		case _, ok := <-listener.Channel():
+			if !ok {
+				return nil
```
asutula: If for some reason textile event store closes the listener channel, we need to return

api/service.go
```diff
@@ -93,21 +319,108 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
...
+				}
+				return server.Send(&pb.ListenReply{
+					Entity: res,
+				})
 			})
-		})
-		if err != nil {
-			return err
+			if err != nil {
+				return err
```
asutula: If we get an error processing the event, we need to return

api/client/client_test.go
carsonfarmer: It does in JS as well. Is there an API to 'cancel' the stream/listener via the gRPC API?

jsign: To leave the note here. This problem was solved in a further commit in this PR. The server-side wasn't checking if the gRPC call context was cancelled, thus stayed in an infinite loop which made the `GracefulShutdown` of the server in the test to hang forever.

api/client/client.go
asutula: Yea I like that idea. It looks like it could close the grpc connection as well as cancel a context that would cascade down into any requests that are currently in fight (including transactions if we use contexts correctly in that scenario). I'll try to make this change.

api/client/client.go
asutula: I'm not totally following what you're saying here, but it makes me think of two things...

1. My original comment was meant to convey that it would be nice for the user to get go-typed data instead of json (`[]byte` or `string`)
2. I think maybe you're talking about making json mode return `[]byte` instead of `string`s. I've actually been thinking that could make more sense. I need to think through the repercussions in other languages, but for Go, it seems to make sense.

api/client/client.go
asutula: Yea, not a ton of help I guess if the cannel is still `interface{}`. In that case, the user could just cast to their go type. The way it is now, the unmarshal into their go type. Not much different.

api/client/client.go
asutula: Yea, it was the server implementation of `Listen` that never returns. Got that fixed and the test for this method now works.

api/client/client_test.go
asutula: nice. I guess the part I need some details on is how to "get the multiaddr and keys from Server1".

api/client/client_test.go
asutula: Hmm I think I'm just not making some connection here... Those links point to functions that operate on the non-json queries and criterions. I need a similar API for JSONQuery, right? Does it exist?

api/client/client_test.go
asutula: Yea, I think once we get a feel for using the transactions, we can make tweaks, but I think you're idea of returning an `end` function from `txn.Start` could be cool. I'll try that out.

api/client/client_test.go
asutula: Yep, fixed.

api/client/client.go
asutula: Ok I added a simple implementation of a client Close method. It closes the grpc connection and calls the root context's cancel function. That root context is used by all client requests. I need to take another pass through with fresh mind/eyes to make sure the root context is correctly used in `Listen`, `ReadTransaction`, and `WriteTransaction`.

api/client/client.go
asutula: I agree we can improve this. Mind is a little too tired right now to make changes here, so I'll come back to this another day.

api/client/client_test.go
asutula: Ok, `txn.Start` now returns the end function. I like it, check it out.

jsign: Nice :+1:

api/client/client.go
asutula: @jsign what if the returned channel was a channel of `ListenEvent` where:
```
type ListenEvent struct {
    data []byte
    err error
```
That way, we can just send the error on the same channel and close the channel.

jsign: Yes, that can be a possible option. Check each received `ListenEvent` if `err != nil` and do accordingly.

asutula: Ok, made that change and addressed the issue of inspecting the error type and just swallowing it silently if it is the error from disposing of the listener.

api/client/client.go
```diff
@@ -0,0 +1,254 @@
...
+		defer close(channel)
+		req := &pb.ListenRequest{
+			StoreID:   storeID,
+			ModelName: modelName,
+			EntityID:  entityID,
+		}
+		stream, err := c.client.Listen(ctx, req)
+		if err != nil {
+			channel <- ListenEvent{err: err}
+			return
```
asutula: If we get an error trying to start the listen request, send the error over the channel and then close the channel, we're done.

api/client/client.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		for {
+			event, err := stream.Recv()
+			if err != nil {
+				status := status.Convert(err)
+				if status == nil || (status != nil && status.Code() != codes.Canceled) {
+					channel <- ListenEvent{err: err}
```
asutula: If we get an error from the server stream, send it over the channel and close the channel as long as the error isn't the result of canceling the listener.

api/client/client_test.go
```diff
@@ -0,0 +1,502 @@
...
+		if err := json.Unmarshal(val.data, p); err != nil {
+			t.Fatalf("failed to unmarshal listen result: %v", err)
+		}
+		if p.Age != 40 {
+			t.Fatalf("expected listen result age = 40 but got: %v", p.Age)
+		}
+	}
+}
+
+func server(t *testing.T) (*api.Server, func()) {
```
asutula: @jsign now that your recent PR is merged in here, what was the change that you told me we can revert here?

asutula: Hmm, it was moving the call to `server.Close()` in the returned cleanup method to after the call to `ts.Close()`, but I tried reverting and none of the tests pass... appears to hang, probably when calling the cleanup method.

jsign: Looking again at the current state of this file, the way is now is right. Maybe on Friday I was looking at your original code (not with my suggested change from slack).

Anyway, the returned `func()` for closing seems to be correct as is.

api/client/client.go
sanderpick: +1 for `[]byte` return types in JSON mode.

api/client/client.go
```diff
@@ -0,0 +1,254 @@
...
+		Values:    values,
+	}
+
+	resp, err := c.client.ModelCreate(c.ctx, req)
+	if err != nil {
+		return err
+	}
+
+	for i, entity := range resp.GetEntities() {
+		err := json.Unmarshal([]byte(entity), items[i])
```
sanderpick: Related to the conversation below about whether the store should return bytes or strings, if here you just returned an error you wouldn't give the user the impression that they can expect the client to unmarshal model instances. Alternatively, the find methods could return `[]interface{}` and just have to perform a type assertion. Right now it's sorta both: "i'll unmarshal when i can, otherwise you do it"

api/client/client.go
```diff
@@ -0,0 +1,254 @@
...
+func (c *Client) WriteTransaction(storeID, modelName string) (*WriteTransaction, error) {
+	client, err := c.client.WriteTransaction(c.ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &WriteTransaction{client: client, storeID: storeID, modelName: modelName}, nil
+}
+
+// Listen provides an update whenever the specified model is updated
+func (c *Client) Listen(storeID, modelName, entityID string) (chan ListenEvent, func(), error) {
```
sanderpick: Since you're not ever using `error`, you can return `(<-chan ListenEvent, context.CancelFunc)`, which indicates a read-only channel, and a more specific function type.

api/client/write.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	pb "github.com/textileio/go-textile-threads/api/pb"
+	es "github.com/textileio/go-textile-threads/eventstore"
+)
+
+// WriteTransaction encapsulates a read transaction
```
sanderpick: r/read/write

api/pb/api.proto
```diff
@@ -1,6 +1,11 @@
 syntax = "proto3";
 package api.pb;
 
+option java_multiple_files = true;
+option java_package = "io.textile.grpc_textile_threads";
+option java_outer_classname = "TextileThreads";
+option objc_class_prefix = "TTE";
```
sanderpick: 👍

api/service.go
```diff
@@ -45,36 +50,257 @@ func (s *service) RegisterSchema(ctx context.Context, req *pb.RegisterSchemaRequ
...
+	}
+	return s.processCreateRequest(req, model.Create)
+}
 
-	store, err := s.getStore(req.StoreID)
+func (s *service) ModelSave(ctx context.Context, req *pb.ModelSaveRequest) (*pb.ModelSaveReply, error) {
+	model, err := s.getModel(req.StoreID, req.ModelName)
 	if err != nil {
-		return nil, err
+		return nil, status.Error(codes.NotFound, "model not found")
```
sanderpick: This `err` should already be correctly wrapped in `getModel`. It could be "store" not found as well.

api/service.go
```diff
@@ -45,36 +50,257 @@ func (s *service) RegisterSchema(ctx context.Context, req *pb.RegisterSchemaRequ
...
+	default:
+		return fmt.Errorf("ReadTransactionRequest.Option has unexpected type %T", x)
+	}
+
+	model, err := s.getModel(storeID, modelName)
+	if err != nil {
+		return status.Error(codes.NotFound, "model not found")
+	}
+
+	err = model.ReadTxn(func(txn *es.Txn) error {
```
sanderpick: nit: `return model.ReadTxn(func(txn *es.Txn) error {` ?

api/service.go
```diff
@@ -45,36 +50,257 @@ func (s *service) RegisterSchema(ctx context.Context, req *pb.RegisterSchemaRequ
...
+	default:
+		return fmt.Errorf("WriteTransactionRequest.Option has unexpected type %T", x)
+	}
+
+	model, err := s.getModel(storeID, modelName)
+	if err != nil {
+		return status.Error(codes.NotFound, "model not found")
+	}
+
+	err = model.WriteTxn(func(txn *es.Txn) error {
```
sanderpick: same nit

api/client/client_test.go
sanderpick: I think in this case (with JSON mode), the "fluent" part has to be implemtned on the client.

api/client/client_test.go
sanderpick: This is currently a manual process... some methods from the foldersync app:
- create the full multiaddress: https://github.com/textileio/go-foldersync/blob/master/client.go#L167
- parse it on the other end and use with `StartFromAddress`: https://github.com/textileio/go-foldersync/blob/master/client.go#L141

api/client/client.go
asutula: Think we need to talk through this in a bit more. I think it's easy to cross wires in this conversation. There are two issues:

1. Should the back end event store code return `[]byte` or `string` for json mode entities? (right now it returns `string` data)
2. Should the client apis return `[]byte` or `string` to the user? (seems that we're setting on `[]byte`, but sometimes we're having to convert from `string` to `[]byte` because of the current state of things related to #1)

I'm not positivities

api/client/client.go
asutula: I'm not sure if you're saying things are fine as is or they need to be changed? Like, for example, you say "if here you just returned an error...". Here I do just return and error, but it's not clear if you are saying that is good or bad. Let's chat through this more.

api/client/client.go
asutula: Nice, good call. Didn't know that about read-only channel syntax... cool!

api/client/client_test.go
asutula: Cool, thanks. I'll use some of that.

api/client/client_test.go
asutula: Ah ok.

api/client/client.go
sanderpick: Oop, sorry, that was confusing. I was considering the unmarshalled items as part of the return... forget that, I'll rephrase: If you let the user do the unmarshalling, they wouldn't have the expectation that the other methods would also perform unmarshalling, and it would feel more appropirate to deal with `[][]byte` in the client api (not referring to the internal JSON mode API). Though, looking again, this unmarshalling here is just to include the `ID` param, making it useful to keep.

api/client/client.go
sanderpick: Right, the +1 was for your number 1. 

You number 2 relates to my other comment about settling on one approach. Not so much that one is better than the other, but that we should be consistent. Now I'm thinking, because it's quite nice to pass `interface{}` types into model create, maybe we should also return pre-marshalled `interface{}`s here too?

api/service.go
asutula: Nice catch, thx

api/service.go
asutula: Yup

api/client/client.go
sanderpick: In any case, a chat sounds good! Morning may be best as I'm heading out now for a bit.

asutula: Yea, let's chat in the morning for sure.

api/client/client.go
asutula: Issue to track this here https://github.com/textileio/go-textile-threads/issues/141

api/client/client.go
asutula: issue to track json mode using []byte https://github.com/textileio/go-textile-threads/issues/142

asutula: Issue to track return type behavior https://github.com/textileio/go-textile-threads/issues/143

api/client/client_test.go
asutula: Made an issue for this https://github.com/textileio/go-textile-threads/issues/144

api/client/client_test.go
asutula: Issue here https://github.com/textileio/go-textile-threads/issues/145

api/client/client_test.go
asutula: Tracking this here https://github.com/textileio/go-textile-threads/issues/146

Issue #84, opened 2019-11-13
eventstore/query: multisort cirteria


jsign:
Sound reasonable to implement queries that can sort not only on one criteria: `.Sort('Field1', 'Field2', ...)`.

PR #130, opened 2019-11-25
store: jsonmode query support

jsign:
New API for making queries in JSON Mode.

It receives a serializable struct and will return a slice of string which are the json values of the instances of the query.

Created #131 with some further comments about some impact of this new JSONMode feature. I guess that's something to evaluate more in the future, but we all have it somewhat present in our minds.

Comments below.

### DIFF
--- a/eventstore/model.go
+++ b/eventstore/model.go
@@ -126,6 +126,15 @@ func (m *Model) Find(result interface{}, q *Query) error {
 	})
 }
 
+// FindJSON executes a Query in in JSONMode and returns the result.
+func (m *Model) FindJSON(q JSONQuery) (ret []string, err error) {
+	m.ReadTxn(func(txn *Txn) error {
+		ret, err = txn.FindJSON(q)
+		return err
+	})
+	return
+}
+
 // Reduce processes an event into the model.
 func (m *Model) Reduce(event core.Event) error {
 	log.Debugf("reducer %s start", m.name)
--- a/eventstore/query_json.go
+++ b/eventstore/query_json.go
@@ -1,0 +1,232 @@
+package eventstore
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
+	dsquery "github.com/ipfs/go-datastore/query"
+)
+
+type JSONQuery struct {
+	Ands []JSONCriterion
+	Ors  []JSONQuery
+	Sort JSONSort
+}
+type JSONCriterion struct {
+	FieldPath string
+	Operation JSONOperation
+	Value     JSONValue
+}
+
+type JSONValue struct {
+	String *string
+	Bool   *bool
+	Float  *float64
+}
+
+type JSONSort struct {
+	FieldPath string
+	Desc      bool
+}
+
+type JSONOperation int
+
+const (
+	Eq = JSONOperation(eq)
+	Ne = JSONOperation(ne)
+	Gt = JSONOperation(gt)
+	Lt = JSONOperation(lt)
+	Ge = JSONOperation(ge)
+	Le = JSONOperation(le)
+)
+
+type marshaledValue struct {
+	value   map[string]interface{}
+	rawJson []byte
+}
+
+func (t *Txn) FindJSON(q JSONQuery) ([]string, error) {
+	dsq := dsquery.Query{
+		Prefix: t.model.dsKey.String(),
+	}
+	dsr, err := t.model.store.datastore.Query(dsq)
+	if err != nil {
+		return nil, fmt.Errorf("error when internal query: %v", err)
+	}
+
+	var values []marshaledValue
+	for {
+		res, ok := dsr.NextSync()
+		if !ok {
+			break
+		}
+		val := make(map[string]interface{})
+		if err := json.Unmarshal(res.Value, &val); err != nil {
+			return nil, fmt.Errorf("error when unmarshaling query result: %v", err)
+		}
+		ok, err = q.matchJSON(val)
+		if err != nil {
+			return nil, fmt.Errorf("error when matching entry with query: %v", err)
+		}
+		if ok {
+			values = append(values, marshaledValue{value: val, rawJson: res.Value})
+		}
+	}
+
+	if q.Sort.FieldPath != "" {
+		var wrongField, cantCompare bool
+		sort.Slice(values, func(i, j int) bool {
+			fieldI, err := traverseFieldPathMap(values[i].value, q.Sort.FieldPath)
+			if err != nil {
+				wrongField = true
+				return false
+			}
+			fieldJ, err := traverseFieldPathMap(values[j].value, q.Sort.FieldPath)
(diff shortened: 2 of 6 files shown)


eventstore/model.go
```diff
@@ -126,6 +126,15 @@ func (m *Model) Find(result interface{}, q *Query) error {
 	})
 }
 
+// FindJSON executes a Query in in JSONMode and returns the result.
```
jsign: New API for Querys in JSONMode.
Receives a new datastructure to represent the desired query, and return an array of jsons strings with the results.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
+package eventstore
```
jsign: Most of the new code for querys on JSON Mode.
It has some similarities with Go Type Mode, but variable types kind of forced to be different in subtle ways.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+type JSONOperation int
+
+const (
+	Eq = JSONOperation(eq)
+	Ne = JSONOperation(ne)
+	Gt = JSONOperation(gt)
+	Lt = JSONOperation(lt)
+	Ge = JSONOperation(ge)
+	Le = JSONOperation(le)
+)
```
jsign: This represents the data-structure used to make querys.
The operators enum values map directly to internal ones.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+	FieldPath string
+	Operation JSONOperation
+	Value     JSONValue
+}
+
+type JSONValue struct {
+	String *string
+	Bool   *bool
+	Float  *float64
+}
```
jsign: Should have a unique non-nil attribute.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+			break
+		}
+		val := make(map[string]interface{})
+		if err := json.Unmarshal(res.Value, &val); err != nil {
+			return nil, fmt.Errorf("error when unmarshaling query result: %v", err)
+		}
+		ok, err = q.matchJSON(val)
+		if err != nil {
+			return nil, fmt.Errorf("error when matching entry with query: %v", err)
+		}
```
jsign: Datastore json value are unmarshaled to `map[string]interface{}`. Then the Query (`q`) is evaluated against this map.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+			return 0, nil
+		}
+		if f < *critVal.Float {
+			return -1, nil
+		}
+		return 1, nil
+	}
+	log.Fatalf("no underlying value for json criterion was provided")
+	return 0, nil
+}
```
jsign: This is somewhat analogous with the [`compare()` in `criterion.go`](https://github.com/textileio/go-textile-threads/blob/master/eventstore/criterion.go#L186). As mentioned in #131, comparing both functions makes clear the point of the limitations of JSONMode queries vs Go typed queries.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+	case Gt:
+		return result > 0, nil
+	case Lt:
+		return result < 0, nil
+	case Le:
+		return result < 0 || result == 0, nil
+	case Ge:
+		return result > 0 || result == 0, nil
+	default:
+		panic("invalid operation")
```
jsign: Also, notice that JSONMode queries don't have arbitrary `func` operations as in Go Type mode.
In Go Type [you can provide a custom func to sort your data](https://github.com/textileio/go-textile-threads/blob/master/eventstore/query_test.go#L74). This could be done also in JSONMode, but since now will be used by gRPC API doesn't make sense. (If #131 is done, we can add it)

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+			return reflect.Value{}, fmt.Errorf("instance field %s doesn't exist in type %s", fieldPath, value)
+		}
+		v, ok := m[fields[i]]
+		if !ok {
+			return reflect.Value{}, fmt.Errorf("instance field %s doesn't exist in type %s", fieldPath, value)
+		}
+		curr = v
+	}
+	return reflect.ValueOf(curr), nil
+}
```
jsign: The analogous of `traverseFieldPath()` which uses reflection working on go types. Traverses a field path by navigating the `map[string]interface{}` up to reaching the desired field (for latter operation evaluation)

eventstore/query_jsonmode_test.go
```diff
@@ -0,0 +1,658 @@
...
+	totreadMid = 124.36
+	totreadEq2 = 150.0
+	totreadMax = 1221.25
+	totreadMin = 12.22
+	ratingMin  = 1.0
+	rating1    = 3.2
+	ratingMid  = 3.6
+	ratingMax  = 5.0
+
+	jsonQueries = []jsonQueryTest{
```
jsign: A bunch of tests to good coverage, since all this feels brittle:
- Ands for each possible datatype `string`, `bool`, `float`
- Ors
- Ordering (again for each possible datatype to be sure, with asc and desc)
- Mixing ones: `and` with `or` with `sort` (asc/desc)

Also, helps to understand a bit how this is used, capabilities, etc.

eventstore/query_json.go
```diff
@@ -0,0 +1,232 @@
...
+	"sort"
+	"strings"
+
+	dsquery "github.com/ipfs/go-datastore/query"
+)
+
+type JSONQuery struct {
+	Ands []JSONCriterion
+	Ors  []JSONQuery
+	Sort JSONSort
```
asutula: Sounds from the title like this PR doesn't address sorting, but since this is here I'll just leave a thought... Would be cool to specify an array of sort objects for secondary etc sorting.

jsign: Sorry, the title is wrong. After creating the PR I added the sorting feature.

Regarding your advice, yes, I created a GH some weeks ago to add the multilevel sorting. #84

asutula: Sweet!

jsign:
> This is pretty advanced stuff, but looks great as far as I can tell. One question... I see in the tests that you build a query statically in one declaration. Can you imagine an API for more easily writing a query? Something like the fluent style of the Gotype-based query API?

Yes, the way that it works for non-json mode `.Where('name').Eq('Ignacio').And('Age').Eq(30).Or(...)` (like you say, the fluent-style). 
For this `FindJSON` I left the input as the struct since I think it would make easier to call it from the gRPC port.

jsign:
Thanks for the review. I guess I'll merge it now since this may unblock your work and can start playing with it. Despite that, would be nice to hear if this makes sense for @sanderpick  too... so I accept further comments/questions even if merged. As aaron said, this is a bit complex so I'd never deny more eyes.

asutula:
I'm not seeing how to access the fluent api of `JSONQuery`. Can you point me in the right direction?

PR #132, opened 2019-11-26
threads: fetch all unknown records

### LINKED ISSUE #128: store: putRecord should fetch history and notify in order
If a record isn't known, `putRecord` fetched the record and transforms it into the current log head.
The problem here is that the received record may be multiple steps ahead in the height of the current head, thus indirectly should add more records. 

We should address two things:
- We should guard the above proccess, since a concurrent put record of a lower height record can be interpreted as a new head too (since the first call is yet to fetch this older record, which will be interpreted as _not known_ since isn't yet in the _blockstore_).
- Since new records can fetch more previous records, we should take care of:
   - Not only notifying the pushed record, but all the records consecuence of the fetching until known head.
   - Notifying the fetched records on _bus_ in the correct order (i.e: the put record block will be notified last).

jsign:
Fixes #128 

Apart from that main fix, I'll add some further comments with some considerations that I'd like @sanderpick  opinion.

### DIFF
--- a/cbor/record.go
+++ b/cbor/record.go
@@ -101,17 +101,17 @@ func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record
 	if err != nil {
 		return nil, err
 	}
-
-	event, err := EventFromNode(block)
+	event, ok := block.(*Event)
+	if !ok {
+	event, err = EventFromNode(block)
 	if err != nil {
 		return nil, err
 	}
-
+}
 	header, err := event.GetHeader(ctx, dag, nil)
 	if err != nil {
 		return nil, err
 	}
-
 	body, err := event.GetBody(ctx, dag, nil)
 	if err != nil {
 		return nil, err
--- a/cbor/record.go
+++ b/cbor/record.go
@@ -103,11 +103,11 @@ func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record
 	}
 	event, ok := block.(*Event)
 	if !ok {
-	event, err = EventFromNode(block)
-	if err != nil {
-		return nil, err
+		event, err = EventFromNode(block)
+		if err != nil {
+			return nil, err
+		}
 	}
-}
 	header, err := event.GetHeader(ctx, dag, nil)
 	if err != nil {
 		return nil, err
--- a/cbor/record.go
+++ b/cbor/record.go
@@ -101,17 +101,17 @@ func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record
 	if err != nil {
 		return nil, err
 	}
-	// (jsign): shouldn't we check if block.(*cbor.Event) to avoid  EventFromNode()?
+
 	event, err := EventFromNode(block)
 	if err != nil {
 		return nil, err
 	}
-	// (jsign): same here?
+
 	header, err := event.GetHeader(ctx, dag, nil)
 	if err != nil {
 		return nil, err
 	}
-	// same here?
+
 	body, err := event.GetBody(ctx, dag, nil)
 	if err != nil {
 		return nil, err
--- a/cbor/record.go
+++ b/cbor/record.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-cid"
-	"github.com/ipfs/go-ipld-cbor"
-	"github.com/ipfs/go-ipld-format"
+	cbornode "github.com/ipfs/go-ipld-cbor"
+	format "github.com/ipfs/go-ipld-format"
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	mh "github.com/multiformats/go-multihash"
 	"github.com/textileio/go-textile-core/crypto"
@@ -101,14 +101,17 @@ func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record
 	if err != nil {
 		return nil, err
 	}
+	// (jsign): shouldn't we check if block.(*cbor.Event) to avoid  EventFromNode()?
 	event, err := EventFromNode(block)
 	if err != nil {
 		return nil, err
 	}
+	// (jsign): same here?
 	header, err := event.GetHeader(ctx, dag, nil)
 	if err != nil {
 		return nil, err
 	}
+	// same here?
 	body, err := event.GetBody(ctx, dag, nil)
 	if err != nil {
 		return nil, err
--- a/client.go
+++ b/client.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	// reqTimeout is the duration to wait for a request to complete.
-	reqTimeout = time.Second * 5
+	reqTimeout = time.Second * 10
 )
 
 // getLogs in a thread.
--- a/eventstore/storethread.go
+++ b/eventstore/storethread.go
@@ -85,7 +85,14 @@ func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
 
 			event, err := threadcbor.EventFromRecord(ctx, a.api, rec.Value())
 			if err != nil {
-				log.Fatalf("error when getting event from record: %v", err) // ToDo: Buffer them and retry...
+				block, err := rec.Value().GetBlock(ctx, a.api)
+				if err != nil { // ToDo: Buffer them and retry...
+					log.Fatalf("error when getting block from record: %v", err)
+				}
+				event, err = threadcbor.EventFromNode(block)
+				if err != nil {
+					log.Fatalf("error when decoding block to event: %v", err)
+				}
 			}
 
 			readKey, err := a.api.Store().ReadKey(a.threadID)
--- a/service.go
+++ b/service.go
@@ -118,15 +118,10 @@ func (s *service) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.Push
 	}
 
 	lg := logFromProto(req.Log)
-	head, err := s.threads.store.Heads(req.ThreadID.ID, lg.ID)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
+
 	// Fix concurrency: adding head without having the block guarantee
-	if head == nil {
-		if err := s.threads.store.AddLog(req.ThreadID.ID, lg); err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
-		}
+	if err := s.threads.store.AddLog(req.ThreadID.ID, lg); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
 	func() {
--- a/service.go
+++ b/service.go
@@ -118,11 +118,19 @@ func (s *service) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.Push
 	}
 
 	lg := logFromProto(req.Log)
-	if err := s.threads.store.AddLog(req.ThreadID.ID, lg); err != nil {
+	head, err := s.threads.store.Heads(req.ThreadID.ID, lg.ID)
+	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if head == nil {
+		if err := s.threads.store.AddLog(req.ThreadID.ID, lg); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	go func() {
+		s.threads.pullLock.Lock()
+		defer s.threads.pullLock.Unlock()
 		// Get log records for this new log
 		recs, err := s.getRecords(
 			s.threads.ctx,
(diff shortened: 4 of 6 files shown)


client.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	// reqTimeout is the duration to wait for a request to complete.
-	reqTimeout = time.Second * 5
+	reqTimeout = time.Second * 10
```
jsign: Raised to avoid some deadlines from clients.

cbor/record.go
```diff
@@ -101,10 +101,13 @@ func RecordToProto(ctx context.Context, dag format.DAGService, rec thread.Record
 	if err != nil {
 		return nil, err
 	}
-	event, err := EventFromNode(block)
+	event, ok := block.(*Event)
+	if !ok {
+	event, err = EventFromNode(block)
```
jsign: Is this reasonable? Maybe for some reason the `block` already is a decoded `Event`, so to avoid unmarshaling again.

Wouldn't be useful that `Record` interface has a `GetEvent` that returns the underlying event and does whatever necessary to achieve that? It would be useful here and in other parts (that I'd mention with double asterisk (**) in other comments).

eventstore/storethread.go
```diff
@@ -85,7 +85,14 @@ func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
...
 			if err != nil {
-				log.Fatalf("error when getting event from record: %v", err) // ToDo: Buffer them and retry...
+				block, err := rec.Value().GetBlock(ctx, a.api)
+				if err != nil { // ToDo: Buffer them and retry...
+					log.Fatalf("error when getting block from record: %v", err)
+				}
+				event, err = threadcbor.EventFromNode(block)
+				if err != nil {
+					log.Fatalf("error when decoding block to event: %v", err)
+				}
```
jsign: Bang. Here we had a problem that was uncovered now that I fetch all the unkown records when we get a far ahead record for a log.

Records that aren't local, seems that never have the `Event` cached. So I should do here all the path to get it. Getting the raw block for the record, then unmarshal it to an Event.

Here is the ** I mentioned before. A `Record` method to get the underlying event maybe be useful here?

service.go
```diff
@@ -118,11 +118,13 @@ func (s *service) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.Push
...
+	// Fix concurrency: adding head without having the block guarantee
 	if err := s.threads.store.AddLog(req.ThreadID.ID, lg); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	go func() {
+	func() {
 		// Get log records for this new log
 		recs, err := s.getRecords(
 			s.threads.ctx,
```
jsign: Disabled `go` here, mostly to avoid concurrency.

I think that this `AddLog` that is here is wrong, but I'd like some confirmation/opinion.
If our log is at height 10, and the received log information has head at height 1, then `AddLog` will override the head with a lower height. 

Am I missing something? 

I think the solution is just deleting that code and let `putRecord` do the set head job (as can be seen below in other comment. I'll reference this point in the other comment with ### mark ).

service.go
```diff
@@ -222,6 +224,9 @@ func (s *service) GetRecords(ctx context.Context, req *pb.GetRecordsRequest) (*p
 
 // PushRecord receives a push record request.
 func (s *service) PushRecord(ctx context.Context, req *pb.PushRecordRequest) (*pb.PushRecordReply, error) {
+	// ToDo: fix concurrency
+	s.threads.pullLock.Lock()
+	defer s.threads.pullLock.Unlock()
```
jsign: Global lock!, temporary.

threads.go
```diff
@@ -187,6 +187,14 @@ func (t *threads) Close() (err error) {
 	if len(errs) > 0 {
 		return fmt.Errorf("failed while closing threads; err(s): %q", errs)
 	}
+
+	t.pullLock.Lock()
+	defer t.pullLock.Unlock()
+	// Wait for all thread pulls to finish
+	for _, semaph := range t.pullLocks {
+		semaph <- struct{}{}
+	}
```
jsign: Quite important thing that was missing.
On `Close()`, we should wait for all thread pulls that might be running before saying we're done.

This basically tries to get all 1 buffered channels (semaphores) of threads and basically wait to _lock_ all of them. If that's the case, no thread pull is running.

threads.go
```diff
@@ -271,22 +279,24 @@ func (t *threads) AddThread(
...
-	// ToDo: fix concurrency
 	t.pullLock.Lock()
 	defer t.pullLock.Unlock()
 	if ptl, ok = t.pullLocks[id]; !ok {
 		ptl = make(chan struct{}, 1)
 		t.pullLocks[id] = ptl
 	}
-	// t.pullLock.Unlock()
+	return ptl
+}
```
jsign: Some helper function to get the semaphore for a thread.
Will be more useful for later changes that keep fixing concurrency per thread.

threads.go
```diff
@@ -541,11 +551,24 @@ func (t *threads) Subscribe(opts ...options.SubOption) tserv.Subscription {
 
 // putRecord adds an existing record. See PutOption for more.
 func (t *threads) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec thread.Record) error {
```
jsign: Ok, main purpose of the PR. Putting a new record when also fetching all missing history up to the current head. And notify in bus in order of all added things.

threads.go
```diff
@@ -541,11 +551,24 @@ func (t *threads) Subscribe(opts ...options.SubOption) tserv.Subscription {
...
+			break
+		}
+		r, err := t.GetRecord(ctx, id, cid)
+		if err != nil {
+			return err
+		}
+		unknownRecords = append(unknownRecords, r)
+		cid = r.PrevID()
 	}
-	if knownRecord {
```
jsign: Iterate from received record, while navigating previous ones, up to the point that we recognize one of them (should be the current head).
All the unkonwn records, are stored in a list `unkownRecords`.

threads.go
```diff
@@ -541,11 +551,24 @@ func (t *threads) Subscribe(opts ...options.SubOption) tserv.Subscription {
...
+	if len(unknownRecords) == 0 {
```
jsign: We had the received record? Nothing extra to do.

Further note: We're assuming that having a Record CID in the blockstore means we know that *the current log* has that record in it (and how knows, maybe not *other log*, since the blockstore is shared). That seems quite strange (two logs with a record with the same CID). Was this your original rationale too?

threads.go
```diff
@@ -555,42 +578,51 @@ func (t *threads) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
-	}
-	body, err := event.GetBody(ctx, t, nil)
-	if err != nil {
-		return err
-	}
-	err = t.AddMany(ctx, []format.Node{rec, event, header, body})
-	if err != nil {
-		return err
-	}
+	for i := len(unknownRecords) - 1; i >= 0; i-- {
```
jsign: Iterate from right to left (so, in inverse log order) each unknown record.
This will process things _in order_, in particular, will notify in `bus` in the correct order.

threads.go
```diff
@@ -555,42 +578,51 @@ func (t *threads) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
+		block, err := r.GetBlock(ctx, t)
+		if err != nil {
+			return err
+		}
+		event, ok := block.(*cbor.Event)
+		if !ok {
+			event, err = cbor.EventFromNode(block)
+			if err != nil {
+				return fmt.Errorf("invalid event: %v", err)
+			}
```
jsign: The cast was always failing. That's why on  `!ok` I unmarshal the raw block to an `Event`.
(Again, other ** case here?).

If you take a look at deleted code, this extra work wasn't there.
I think this worked before because records that are received from gRPC are not partial. But previous records we fetch now with `r.GetBlock(..)` (thus, only with the `DAGResolver` *are partial*).
Make sense?

threads.go
```diff
@@ -555,42 +578,51 @@ func (t *threads) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
+		}
+		err = t.AddMany(ctx, []format.Node{r, event, header, body})
+		if err != nil {
+			return err
+		}
+
+		log.Debugf("put record %s (thread=%s, log=%s)", r.Cid().String(), id, lg.ID)
 
+		// Notify local listeners
+		err = t.bus.SendWithTimeout(&record{
```
jsign: In order delivery of new discovered records.

threads.go
```diff
@@ -555,42 +578,51 @@ func (t *threads) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
+			threadID: id,
+			logID:    lg.ID,
+		}, time.Second)
+		if err != nil {
+			return err
+		}
+	}
 	// Update head
-	if err = t.store.SetHead(id, lg.ID, rec.Cid()); err != nil {
+	if err = t.store.SetHead(id, lg.ID, unknownRecords[0].Cid()); err != nil {
```
jsign: After looping for all the unknown records in order (reverse slice order), we set the current head to the first element (which is always the one we received as a param in the method).

If you recall a previous comment, this is the ### mark.

jsign:
This was an important step to have some correctness in the underlying working of threads.
The next step will be making concurrency safe per-thread _at least_ (for some work it could be concurrent _per log_ but we can leave that for an extra opt in the future).

TBH, I tried to make those changes in this PR, but they're a bit more complicated because I think I should move some code to other places (so to no double lock the same mutex, etc). Prefer to keep this PR as short as possible.

threads.go
sanderpick: Clever!

threads.go
```diff
@@ -271,22 +279,27 @@ func (t *threads) AddThread(
...
 	if ptl, ok = t.pullLocks[id]; !ok {
 		ptl = make(chan struct{}, 1)
 		t.pullLocks[id] = ptl
 	}
-	// t.pullLock.Unlock()
+	return ptl
+}
 
+// PullThread for new records.
+// Logs owned by this host are traversed locally.
```
sanderpick: Not related to this PR, but this comment must be outdated / relic from the old pull functionality.

threads.go
sanderpick: Hmm, good point! At least currently where all events are encrypted with the single-use keys, it's not possible to duplicate a record across a log. If it was possible, I'm not sure how we'd be able to check if the record actually in _the log record chain_ w/o using a diff blockstore for each log. Maybe we'd have to use a new index in the datastore.

threads.go
sanderpick: Sweet! For extra large traversals, say `unknownRecords` becomes _massive_, we could write these to disk as an unprocessed record queue.

threads.go
sanderpick: Ah yeah, makes sense! Somewhat related enhancemen: It would be nice to fetch these blocks from known peers that can return the fully loaded records. As in, not from the plain `DAGResolver` (perhaps only as a last resort).

eventstore/storethread.go
jsign: By non-local events I mean events generated by *other* peers. And yes, records received by grpc service are fully loaded. 

But here's the tricky part: we receive the *head* fully loaded. If while adding the head we discover that we are missing extra records up to our known head, those missing records will be fetched from the `DAGService`, and *those* records aren't fully loaded!

When one of those discovered missing record fetched with the `DAGResolver` (not received from gRPC) is dispatched to the store: bang!.

Makes more sense?

service.go
jsign: Please keep that comment in mind for my next PR that I'm fixing the concurrency problems and deleting this temporary global lock.

I move some things of place since I want to make some clear distinctions between thread-safe methods and not thread-safe. (or must say, goroutine safe). I added comments to methods and that PR will be heavily commented.

I want to refresh what you mention in that PR if possible.

threads.go
jsign: Yeah, basically using a 1 buffered channel is almost the same as a mutex. The thing is that this way you can test for lock and dismiss if locked. (with a default case).

I'm not sure I explained this decision originally. In `PullThread` we want to *try* to pull the thread, but if it's already being pulled we want to dismiss and go on. You can't "try" a Lock() with a mutex, but you can with a channel.

Anyway, this will make more sense in the next PR when these _semaphores_ are fully used (and delete the global lock). Did it now just not to forget and let the threadservice close when still are threads being pulled.

threads.go
jsign: I have a very cool gem to tell you about this blockstore assumption we did. Took me quite a long time to figure out that was a quite insane bug. 
I'll give further details about this in the next PR.

threads.go
jsign: Yes, that's right... will grow linearly with how far behind we're. 
That sound a reasonable option, we can think it further.

threads.go
jsign: Oh, well.. I think I clarified this further in an above big comment.
Yeah, more directed fetching is always nicer to avoid flooding other peers.

There're also a lot of paralellized fetching for new records for logs, which aren't cancelled when one of the addrs replies with information. I think there're many optimizations we can do to all this process... eventually we will have time for this.

jsign:
> Great! I left some non-blocking questions and made some comments that could be turned into issues if the ideas sounds reasonable.

Replied to all comments.

I'll merge tomorrow. Feel free to create issues for many points of discussion. Will be nice to have them written before forgetting since all are quite important/interesting.

eventstore/storethread.go
sanderpick: Yep yep, makes sense

threads.go
sanderpick: 👍🏼

sanderpick:
I’ve made a note to create some issues from the discussions. Will go when I’m back on!

PR #135, opened 2019-11-29
store: proper closing of adapter

### LINKED ISSUE #88: storethread: consider own cancellation impact on routines
If `Threadservice` closes the channel, it will finish one direction of event flow but not the other.
Fix to terminate both, and also think about meaning to `Store`.

jsign:
More caring about closing properly the glue between Store and Thread (`SingleThreadAdapter`).
We want proper stopping of the flow of (bidirectional) events between these two worlds.

Also Fixes #88, tbh still now sure about it is the correct decision. Switched to considering those situations an Error but keep things alive (instead of Fatal)

Explanation in comments.

### DIFF
--- a/eventstore/store.go
+++ b/eventstore/store.go
@@ -56,10 +56,12 @@ type Store struct {
 	dispatcher    *dispatcher
 	eventcodec    core.EventCodec
 	threadservice threadservice.Threadservice
+	adapter       *singleThreadAdapter
 
 	lock       sync.RWMutex
 	modelNames map[string]*Model
 	jsonMode   bool
+	closed     bool
 
 	localEventsBus *localEventsBus
 	stateChanged   *stateChangedNotifee
@@ -170,7 +172,8 @@ func (s *Store) Start() error {
 			return err
 		}
 	}
-	adapter := newSingleThreadAdapter(s.ctx, s, id)
+	adapter := newSingleThreadAdapter(s, id)
+	s.adapter = adapter
 	adapter.Start()
 	return nil
 }
@@ -294,6 +297,16 @@ func (s *Store) readTxn(m *Model, f func(txn *Txn) error) error {
 
 // Close closes the store
 func (s *Store) Close() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if s.closed {
+		return
+	}
+	s.closed = true
+
+	if s.adapter != nil {
+		s.adapter.Close()
+	}
 	s.cancel()
 	s.localEventsBus.bus.Discard()
 	s.stateChanged.bus.Discard()
--- a/eventstore/storethread.go
+++ b/eventstore/storethread.go
@@ -20,27 +20,42 @@ const (
 
 // SingleThreadAdapter connects a Store with a Threadservice
 type singleThreadAdapter struct {
-	ctx      context.Context
-	api      tserv.Threadservice
-	store    *Store
-	threadID thread.ID
-	ownLogID peer.ID
+	api        tserv.Threadservice
+	store      *Store
+	threadID   thread.ID
+	ownLogID   peer.ID
+	closeChan  chan struct{}
+	goRoutines sync.WaitGroup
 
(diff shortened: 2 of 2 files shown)


eventstore/store.go
```diff
@@ -294,6 +297,16 @@ func (s *Store) readTxn(m *Model, f func(txn *Txn) error) error {
 
 // Close closes the store
 func (s *Store) Close() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if s.closed {
+		return
+	}
+	s.closed = true
```
jsign: Guard re-closing since would be fatal.

eventstore/store.go
```diff
@@ -294,6 +297,16 @@ func (s *Store) readTxn(m *Model, f func(txn *Txn) error) error {
...
+
+	if s.adapter != nil {
```
jsign: Use new `Close()` function to wait for the storethread adapter to finish its job, being:
- sending new events from store to threads
- sending new external events from threads to store

This new `Close()` function in this adapter will wait gracefully until we're sure all current things finished correctly. Before this function, lifecycle was handleled by a `context.Context` but that gives not guarantees about finishing gracefully on `ctx.Done()`.

eventstore/storethread.go
```diff
@@ -20,29 +20,44 @@ const (
...
-	api      tserv.Threadservice
-	store    *Store
-	threadID thread.ID
-	ownLogID peer.ID
+	api        tserv.Threadservice
+	store      *Store
+	threadID   thread.ID
+	ownLogID   peer.ID
+	closeChan  chan struct{}
+	goRoutines sync.WaitGroup
```
jsign: Represents the two backgrounds goroutines that we should wait on close.

eventstore/storethread.go
```diff
@@ -20,29 +20,44 @@ const (
...
+// of thread<->store
+func (a *singleThreadAdapter) Close() {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+	if a.closed {
+		return
+	}
+	a.closed = true
+	close(a.closeChan)
+	a.goRoutines.Wait()
```
jsign: 1) Close-Signal to background goroutines
2) Wait them to finish
Done.

eventstore/storethread.go
```diff
@@ -62,27 +77,28 @@ func (a *singleThreadAdapter) Start() {
...
 		select {
-		case <-a.ctx.Done():
-			log.Infof("cancelling thread event dispatch to store for thread %s since ctx was cancelled", a.threadID)
+		case <-a.closeChan:
+			log.Debug("closing thread-to-store flow on thread %s", a.threadID)
 			return
 		case rec, ok := <-sub.Channel():
 			if !ok {
-				log.Info("ending dispatch of events to store since thread channel was closed") // ToDo: reconsider action
+				log.Errorf("notification channel closed, not listening to external changes anymore")
```
jsign: Switched to `Error` since this shouldn't be OK. Tempted to `Fatal`, since current result would mean that no more flow of events will happen from Threads to Store. (For some reason Threads closed our listener...)

eventstore/storethread.go
```diff
@@ -120,26 +135,26 @@ func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
...
 		select {
-		case <-a.ctx.Done():
-			log.Infof("cancelling sending store local events to own thread log for thread %s", a.threadID)
+		case <-a.closeChan:
+			log.Infof("closing store-to-thread flow on thread %s", a.threadID)
 			return
 		case event, ok := <-l.Channel():
 			if !ok {
-				log.Infof("ending sending store local event to own thread since channel was closed for thread %s", a.threadID)
+				log.Errorf("ending sending store local event to own thread since channel was closed for thread %s", a.threadID)
```
jsign: Same as above comment.
For some reason the Store closed our statechanged listener... switched from `Info` to `Error` since shouldn't be OK.

PR #136, opened 2019-11-29
api: pointer slice gotcha

sanderpick:


api/service.go
```diff
@@ -72,10 +72,10 @@ func (s *service) StartFromAddress(ctx context.Context, req *pb.StartFromAddress
 	if addr, err = ma.NewMultiaddr(req.GetAddress()); err != nil {
 		return nil, err
 	}
-	if readKey, err = symmetric.NewKey([]byte(req.GetReadKey())); err != nil {
```
sanderpick: Redundant type assertion.

api/service.go
```diff
@@ -339,7 +339,8 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
 func (s *service) processCreateRequest(req *pb.ModelCreateRequest, createFunc func(...interface{}) error) (*pb.ModelCreateReply, error) {
 	values := make([]interface{}, len(req.Values))
 	for i, v := range req.Values {
-		values[i] = &v
+		s := v
```
sanderpick: Allocate a new memory address, otherwise we end up with a slice of pointers that all point to the same address.

PR #149, opened 2019-12-04
Use reflect api to improve find api, more tests

### LINKED ISSUE #143: Consistent return type for client methods
Sometimes we unmarshal json data into user go types, sometimes we return `[]byte` and the user is forced to do that. We need to make sure the behavior across methods makes sense.

asutula:
This allows the user to specify a dummy type when calling find and then use a simple type assertion when receiving the results.

This was a change we decided to make to help the create/save/find apis feel consistent.

closes #143


api/client/read.go
```diff
@@ -69,25 +70,38 @@ func (t *ReadTransaction) FindByID(entityID string, entity interface{}) error {
...
+		for i, result := range x.ModelFindReply.GetEntities() {
+			target := reflect.New(elementType).Interface()
+			err := json.Unmarshal(result, target)
+			if err != nil {
+				return nil, err
+			}
+			val := results.Index(i)
+			val.Set(reflect.ValueOf(target))
+		}
+		return results.Interface(), nil
```
sanderpick: Possible to make a reusable function?

asutula: Yea of course. Good call.

PR #156, opened 2019-12-05
store: batch txn mutations in single Record

### LINKED ISSUE #93: eventcodec: consider single event per txn
As discussed in #74.
Consider making events have all changes of a txn. Currently multiple events are generated per txn, which means that on a partial sync a peer could apply changes and leave state on a partial-sane state.

jsign:
Fixes #93 

The main intention of this PR is making #93 real. Apart from necessary changes for that issue, I took advantage and did others all under the mantra of simplifying things and make code more coherent.

The PR was bigger than expected, but I'm satisfied with all the changes. Some things in the code were a bit weird, but now all seem to fit more correctly.

I'll add many comments to help review.

~~This PR is still a draft since depends on merging a [`go-textile-core` PR](https://github.com/textileio/go-textile-core/pull/34). When merged, a further `go.mod` change will be necessary.~~

### DIFF
--- a/eventstore/dispatcher.go
+++ b/eventstore/dispatcher.go
@@ -20,7 +20,7 @@ var (
 
 // Reducer applies an event to an existing state.
 type Reducer interface {
-	Reduce(event core.Event) error
+	Reduce(events []core.Event) error
 }
 
 // dispatcher is used to dispatch events to registered reducers.
@@ -29,14 +29,14 @@ type Reducer interface {
 // Every event is dispatched to every registered reducer. When a given reducer is registered, it returns a `token`,
 // which can be used to deregister the reducer later.
 type dispatcher struct {
-	store    datastore.Datastore
+	store    datastore.TxnDatastore
 	reducers []Reducer
 	lock     sync.RWMutex
 	lastID   int
 }
 
 // NewDispatcher creates a new EventDispatcher
-func newDispatcher(store datastore.Datastore) *dispatcher {
+func newDispatcher(store datastore.TxnDatastore) *dispatcher {
 	return &dispatcher{
 		store: store,
 	}
@@ -56,29 +56,41 @@ func (d *dispatcher) Register(reducer Reducer) {
 }
 
 // Dispatch dispatches a payload to all registered reducers.
-func (d *dispatcher) Dispatch(event core.Event) error {
+// The logic is separated in two parts:
+// 1. Save all txn events with transaction guarantees.
+// 2. Notify all reducers about the known events.
+func (d *dispatcher) Dispatch(events []core.Event) error {
 	d.lock.Lock()
 	defer d.lock.Unlock()
-	key, err := getKey(event)
+
+	txn, err := d.store.NewTransaction(false)
 	if err != nil {
 		return err
 	}
-	// Encode and add an Event to event store
-	b := bytes.Buffer{}
-	e := gob.NewEncoder(&b)
-	if err := e.Encode(event); err != nil {
-		return err
+	defer txn.Discard()
+	for _, event := range events {
+		key, err := getKey(event)
+		if err != nil {
+			return err
+		}
+		// Encode and add an Event to event store
+		b := bytes.Buffer{}
+		e := gob.NewEncoder(&b)
+		if err := e.Encode(event); err != nil {
+			return err
+		}
+		if err := d.Store().Put(key, b.Bytes()); err != nil {
+			return err
+		}
 	}
-	if err := d.Store().Put(key, b.Bytes()); err != nil {
-		return err
-	}
+	txn.Commit()
 	// Safe to fire off reducers now that event is persisted
 	g, _ := errgroup.WithContext(context.Background())
 	for _, reducer := range d.reducers {
 		reducer := reducer
 		// Launch each reducer in a separate goroutine
 		g.Go(func() error {
-			return reducer.Reduce(event)
+			return reducer.Reduce(events)
 		})
 	}
 	// Wait for all reducers to complete or error out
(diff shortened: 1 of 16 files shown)


eventstore/model.go
```diff
@@ -136,38 +132,6 @@ func (m *Model) FindJSON(q JSONQuery) (ret []string, err error) {
 	return
 }
 
-// Reduce processes an event into the model.
-func (m *Model) Reduce(event core.Event) error {
```
jsign: Yep. Models aren't reducers anymore.
TBH, this always felt weird. They kind of checked if the current event was for them (with that `if event.Type() != m.name` code), and if it corresponds, call the event codec.

As we can see, being in the Model world to do this adds little value. Most of the code in this old method does `m.store.XXXX` for things. Which was already a smell.

Now that what is reduce are aggregates of things of a txn, this becomes even more apparent. In the future, a txn might change instances of multiple models. A single model can't reduce really txn changes.

Now the `Store` itself is the reducer. (The only reducer now in the code). That makes more sense. A store will receive a slice of Events (which might contain changes for multiple models), and will take care of applying the changes.

eventstore/model.go
```diff
@@ -352,19 +316,15 @@ func (t *Txn) Commit() error {
 	if t.discarded || t.commited {
 		return errAlreadyDiscardedCommitedTxn
 	}
-
-	events, err := t.model.store.eventcodec.Create(t.actions)
+	events, node, err := t.model.store.eventcodec.Create(t.actions)
```
jsign: The `EventCodec` now returns the encoded `Node` of all the actions done in this txn.
This `Node` is sent via the `notifyTxnEvents` which will be received by `StoreThread`, which will `AddRecord` to the peer local Log.

eventstore/store.go
```diff
@@ -119,7 +120,7 @@ func newStore(ts threadservice.Threadservice, config *StoreConfig) (*Store, erro
 			return nil, err
 		}
 	}
-
+	s.dispatcher.Register(s)
```
jsign: As mentioned before, the only registered reducer is the Store, not each model.

eventstore/store.go
```diff
@@ -256,18 +255,44 @@ func (s *Store) GetModel(name string) *Model {
 	return s.modelNames[name]
 }
 
+// Reduce processes txn events into the models.
```
jsign: The `Reduce` function moved from Model to Store.
Looks more clean from my POV>

eventstore/storethread.go
```diff
@@ -145,19 +145,13 @@ func (a *singleThreadAdapter) storeToThread(wg *sync.WaitGroup) {
...
 			}
-			n, err := event.Node()
-			if err != nil {
-				log.Fatalf("error when generating node for own log for thread %s: %v", a.threadID, err)
-			}
 			ctx, cancel := context.WithTimeout(context.Background(), addRecordTimeout)
-			log.Debugf("adding new local store event to own log from entityid: %s", event.EntityID())
-			_, err = a.api.AddRecord(ctx, a.threadID, n)
-			if err != nil {
+			if _, err := a.api.AddRecord(ctx, a.threadID, node); err != nil {
```
jsign: Much less work done here.
As mentioned before, now the `StoreThread` receives the `node` just ready to be added as a record. The `EventCodec` already does the work that was done here (doing some wierd `event.Node()` call that doesn't exist anymore).

jsonpatcher/jsonpatcher.go
```diff
@@ -11,9 +11,9 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
```
jsign: Diff in this file is quite insane diff in this file... pretty hard to understand.

A bit of a summary here:
- Instead of `WrapObject` the json-patches (which were single changes), now what's `WrapObject` is a new type:
```
type recordEvents struct {
	Patches []patchEvent
}
```
which basically is a set of `patchEvents`.

In the way of making this change, I avoided some json-serializations and some other things that makes things clearer (might be better to see the direct file and not the diff).

Most of this code is the heart of the main PR intention.

eventstore/dispatcher.go
```diff
@@ -20,7 +20,7 @@ var (
 
 // Reducer applies an event to an existing state.
 type Reducer interface {
-	Reduce(event core.Event) error
+	Reduce(events []core.Event) error
```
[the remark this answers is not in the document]
sanderpick: Awesome

eventstore/listeners.go
```diff
@@ -38,8 +39,8 @@ func (s *Store) notifyStateChanged(actions []Action) {
 	s.stateChangedNotifee.notify(actions)
 }
 
-func (s *Store) broadcastLocalEvent(e core.Event) error {
-	return s.localEventsBus.bus.SendWithTimeout(e, busTimeout)
+func (s *Store) notifyTxnEvents(node ipldformat.Node) error {
+	return s.localEventsBus.send(node)
```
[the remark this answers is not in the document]
sanderpick: > This broadcaster was used to notify the `StoreThread` (glue between `Store` and `Threadservice`) about new local changes in the `Store`, so to add this changes in the local Log of the peer.

Still the case, right? I like the feel of using the `Threadservice` API like a normal user... adding raw nodes.

eventstore/keytransform.go
```diff
@@ -0,0 +1,199 @@
+package eventstore
```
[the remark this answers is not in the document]
sanderpick: Nice! Generics would probably have helped here.

eventstore/model.go
sanderpick: Cool, I like it... conceptually simpler

jsonpatcher/jsonpatcher.go
sanderpick: Awesome

eventstore/listeners.go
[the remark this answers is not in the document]
jsign: Yeah, feels good.  Every component using the other as it should without hacks or special APIs.

bitcard:
mark

PR #160, opened 2019-12-06
Multiple issue fixes

### LINKED ISSUE #81: eventstore/query: check slice type on Find()
Check that return slice for the query is of the proper type of the model.

### LINKED ISSUE #89: storethread: retries on failed fetchs
Implement some mechanism to do some minimum retries before considering a network fech of data fatal (i.e: the body of an `Event`).

### LINKED ISSUE #97: store: enforce model/type registration to be pointers
Registered models/types should be pointers, since the store will autogenereate IDs. If it wasn't a pointer, the client couldn't see the generated ID without making some query about inserted data (which can't be by ID which is weird).

So, feels like error prone to accept non-pointer type instances.

### LINKED ISSUE #111: general: check for Close() ignored errors
Make them explicit and let the caller decide if ignore them.

jsign:
Not heavy stuff in this PR.

- Fixes #89 
- Fixes #81 
- Fixes #111
- Fixes #97
- Added some extra missing tests thanks to linter reporting unused var 
- Fixes some lint checks.

Also fixes tests that were broken in master due to a previous merged PR. All working now.

Extra code comments to help review.

### DIFF
--- a/eventstore/manager.go
+++ b/eventstore/manager.go
@@ -105,11 +105,18 @@ func (m *Manager) GetStore(id uuid.UUID) *Store {
 }
 
 // Close all the in-mem stores.
-func (m *Manager) Close() (err error) {
+func (m *Manager) Close() error {
+	var err error
 	for _, s := range m.stores {
-		s.Close()
+		if err = s.Close(); err != nil {
+			log.Error("error when closing manager datastore: %v", err)
+		}
 	}
-	return m.config.Datastore.Close()
+	err2 := m.config.Datastore.Close()
+	if err != nil {
+		return err
+	}
+	return err2
 }
 
 // getStoreConfig copies the manager's base config and
--- a/eventstore/query.go
+++ b/eventstore/query.go
@@ -11,9 +11,12 @@ import (
 )
 
 var (
-	// ErrInvalidSortingField is returned when a query sorts
-	// a result by a non-existent field in the model schema.
+	// ErrInvalidSortingField is returned when a query sorts a result by a
+	// non-existent field in the model schema.
 	ErrInvalidSortingField = errors.New("sorting field doesn't correspond to instance type")
+	// ErrInvalidSliceType is returned when a query receives a result by a
+	// slice type which doesn't correspond to the model being queried.
+	ErrInvalidSliceType = errors.New("slice type doesn't correspond to model type")
 )
 
 // Query allows to build queries to fetch data from a model.
@@ -69,8 +72,11 @@ func (q *Query) OrderByDesc(field string) *Query {
 // pointers with the correct model type. If the slice isn't empty, will be emptied.
 func (t *Txn) Find(res interface{}, q *Query) error {
 	valRes := reflect.ValueOf(res)
-	if valRes.Kind() != reflect.Ptr || valRes.Elem().Kind() != reflect.Slice {
-		panic("result should be a slice")
+	if valRes.Kind() != reflect.Ptr || // Should be a pointer
+		valRes.Elem().Kind() != reflect.Slice || // To a slice
+		valRes.Elem().Type().Elem().Kind() != reflect.Ptr || // To a pointer
+		valRes.Elem().Type().Elem() != t.model.valueType { // To the model type
+		return ErrInvalidSliceType
 	}
 	if q == nil {
 		q = &Query{}
--- a/eventstore/query_jsonmode_test.go
+++ b/eventstore/query_jsonmode_test.go
@@ -338,6 +338,37 @@ var (
 			},
 		},
 		jsonQueryTest{
+			name:   "LeByRatingMin",
+			resIdx: []int{},
+			query: JSONQuery{
+				Ands: []JSONCriterion{
+					JSONCriterion{
+						FieldPath: "Meta.Rating",
+						Operation: Le,
+						Value: JSONValue{
+							Float: &ratingMin,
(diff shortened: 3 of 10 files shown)


eventstore/manager.go
```diff
@@ -105,11 +105,18 @@ func (m *Manager) GetStore(id uuid.UUID) *Store {
 }
 
 // Close all the in-mem stores.
-func (m *Manager) Close() (err error) {
+func (m *Manager) Close() error {
```
jsign: Decided the most reasonable thing is try to continue closing things even of something fails in the middle.
Each failing thing is logged, and the error returned is the last error that happened.
I think this may be better than just returning on first error since would leave other things open that might be closable.

eventstore/query.go
```diff
@@ -69,8 +72,11 @@ func (q *Query) OrderByDesc(field string) *Query {
 // pointers with the correct model type. If the slice isn't empty, will be emptied.
 func (t *Txn) Find(res interface{}, q *Query) error {
 	valRes := reflect.ValueOf(res)
-	if valRes.Kind() != reflect.Ptr || valRes.Elem().Kind() != reflect.Slice {
-		panic("result should be a slice")
+	if valRes.Kind() != reflect.Ptr || // Should be a pointer
```
jsign: The `res` parameter can't be anything. It should be a concrete type for the query considering the model being queried. Now it checks that is what's expected... since if that check doesn't hold, the code will panic in line: `		resSlice = reflect.Append(resSlice, unsorted[i])`

TBH, these checks has some cost... nothing is free in particular with refect. If the client is sending the wrong thing, maybe it should panic because that's a serious error here. Shouldn't be expected by any means... so maybe these checks can be deleted and make the client responsible to use it correctly.

eventstore/query_jsonmode_test.go
```diff
@@ -97,8 +97,8 @@ var (
 			name:   "EqByTitle1",
```
jsign: Fixes the tests broken in master.

eventstore/query_test.go
```diff
@@ -142,6 +143,16 @@ func TestInvalidSortField(t *testing.T) {
 	}
 }
 
+func TestInvalidSliceType(t *testing.T) {
```
jsign: Extra test to check the new feature of checking that received slice for `Find` controls that has the right type for the queried Model.

eventstore/store.go
```diff
@@ -213,6 +215,11 @@ func (s *Store) Register(name string, defaultInstance interface{}) (*Model, erro
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
+	diType := reflect.TypeOf(defaultInstance)
```
jsign: Check that the registered default instance of the model has a correct type.
They should always be pointers to structs. e.g. `m.Register("Person", &Person{})`.

eventstore/storethread.go
```diff
@@ -101,7 +102,7 @@ func (a *singleThreadAdapter) threadToStore(wg *sync.WaitGroup) {
 			ctx, cancel := context.WithTimeout(context.Background(), fetchEventTimeout)
 			event, err := threadcbor.EventFromRecord(ctx, a.api, rec.Value())
 			if err != nil {
-				block, err := rec.Value().GetBlock(ctx, a.api)
+				block, err := a.getBlockWithRetry(ctx, rec.Value(), 3, time.Millisecond*500)
```
jsign: Fixed 3 retries with 500ms expbackoff. Can eventually be configured with opts

eventstore/storethread.go
```diff
@@ -158,3 +159,17 @@ func (a *singleThreadAdapter) storeToThread(wg *sync.WaitGroup) {
...
+func (a *singleThreadAdapter) getBlockWithRetry(ctx context.Context, rec thread.Record, cantRetries int, backoffTime time.Duration) (format.Node, error) {
+	var err error
+	for i := 1; i <= cantRetries; i++ {
+		n, err := rec.GetBlock(ctx, a.api)
+		if err == nil {
+			return n, nil
+		}
+		log.Warningf("error when fetching block %s in retry %d", rec.Cid(), i)
+		time.Sleep(backoffTime)
+		backoffTime *= 2
```
jsign: exponential backoff

eventstore/util.go
```diff
@@ -14,7 +14,7 @@ import (
 	host "github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
-	peerstore "github.com/libp2p/go-libp2p-peerstore"
+	"github.com/libp2p/go-libp2p-core/peerstore"
```
jsign: Deprecated import

examples/chat/main.go
```diff
@@ -55,32 +55,32 @@ const (
...
-					"Time": {
-						"type": "integer"
-					}
-				},
-				"additionalProperties": false,
-				"type": "object"
-			}
-		}
-	}`
+	// schema = `{
```
jsign: `schema` was unused. Preferred to comment it instead of deleting it to have a second opinion if this was expected?

jsign:
Maybe @sanderpick  might worth taking a look at some PR comments. Mostly the `schema` unused var that I'm not sure had some missing intention.

Not a hurry at all... just to not be missing something.

PR #162, opened 2019-12-09
Create invite grpc api for getting thread info

carsonfarmer:


carsonfarmer:
I guess once this gets merged, we'll need to bump the version and push out releases for `@textile/threads-client-grpc` right?

carsonfarmer:
Then we can update the version in https://github.com/textileio/js-threads-client/pull/21, bump that one, and also release.

api/client/client.go
```diff
@@ -145,6 +146,26 @@ func (c *Client) ModelDelete(storeID, modelName string, entityIDs ...string) err
 	return err
 }
 
+// CreateInvite retrives the components required to create a Thread/store invite.
+func (c *Client) CreateInvite(storeID string) ([]string, error) {
```
sanderpick: Could we call this `GetStoreLink` instead? `CreateInvite` sounds like something is being created to me... where in fact we're just retrieving a pointer to the store.

carsonfarmer: Yup, great idea!

PR #165, opened 2019-12-11
chore: always use multiaddr

sanderpick:
From working on daemon + CLI, it's ugly to mix multi addresses w/ normal TCP addresses. This just enforces that all things are multi.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	logging "github.com/ipfs/go-log"
+	ma "github.com/multiformats/go-multiaddr"
 	tserv "github.com/textileio/go-textile-core/threadservice"
 	pb "github.com/textileio/go-textile-threads/api/pb"
 	es "github.com/textileio/go-textile-threads/eventstore"
@@ -33,8 +34,8 @@ type Server struct {
 // Config specifies server settings.
 type Config struct {
 	RepoPath  string
-	Addr      string // defaults to 0.0.0.0:9090
-	ProxyAddr string // defaults to 0.0.0.0:9091
+	Addr      ma.Multiaddr // defaults to /ip4/127.0.0.1/tcp/9090
+	ProxyAddr ma.Multiaddr // defaults to /ip4/127.0.0.1/tcp/9091
 	Debug     bool
 }
 
@@ -68,10 +69,8 @@ func NewServer(ctx context.Context, ts tserv.Threadservice, conf Config) (*Serve
 		cancel:  cancel,
 	}
 
-	if conf.Addr == "" {
-		conf.Addr = "0.0.0.0:9090"
-	}
-	listener, err := net.Listen("tcp", conf.Addr)
+	addr := util.TCPAddrFromMultiAddr(conf.Addr, "127.0.0.1:9090")
+	listener, err := net.Listen("tcp", addr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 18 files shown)


api/server.go
```diff
@@ -68,10 +69,8 @@ func NewServer(ctx context.Context, ts tserv.Threadservice, conf Config) (*Serve
 		cancel:  cancel,
 	}
 
-	if conf.Addr == "" {
-		conf.Addr = "0.0.0.0:9090"
-	}
-	listener, err := net.Listen("tcp", conf.Addr)
+	addr := util.TCPAddrFromMultiAddr(conf.Addr, "127.0.0.1:9090")
+	listener, err := net.Listen("tcp", addr)
```
jsign: The switch between `0.0.0.0` and `127.0.0.1` was intentional?
It's done in other places too, this questions count for all places :)

eventstore/model_test.go
```diff
@@ -254,7 +255,7 @@ func TestGetInstance(t *testing.T) {
 	t.Run("WithReadTx", func(t *testing.T) {
 		person := &Person{}
 		err = model.ReadTxn(func(txn *Txn) error {
-			txn.FindByID(newPerson.ID, person)
+			_ = txn.FindByID(newPerson.ID, person)
 			checkErr(t, err)
```
jsign: Wow, your lint fix revealed something that was previously wrong here.
Shouldn't be ignored since the below `checkErr` is doing a wrong check.

eventstore/model_test.go
```diff
@@ -265,7 +266,7 @@ func TestGetInstance(t *testing.T) {
 	t.Run("WithUpdateTx", func(t *testing.T) {
 		person := &Person{}
 		err = model.WriteTxn(func(txn *Txn) error {
-			txn.FindByID(newPerson.ID, person)
+			_ = txn.FindByID(newPerson.ID, person)
 			checkErr(t, err)
```
jsign: Same!

eventstore/model_test.go
```diff
@@ -372,7 +373,7 @@ func TestInvalidActions(t *testing.T) {
 	})
 	t.Run("Save", func(t *testing.T) {
 		p := &PersonFake{Name: "fake"}
-		model.Create(p)
+		_ = model.Create(p)
```
jsign: Uhm, could be checked too.

examples/chat/main.go
```diff
@@ -54,33 +54,6 @@ var (
 const (
 	msgTimeout = time.Second * 10
 	timeLayout = "03:04:05 PM"
-
```
jsign: Yeah, was commented just in case I was missing something since was unused.
Mentioned [here](https://github.com/textileio/go-textile-threads/pull/160#discussion_r355057448)

examples/e2e_counter/main.go
```diff
@@ -11,7 +11,6 @@ import (
 func main() {
```
jsign: I think I'll eventually delete this `e2e_counter`.
I feel is quite dirty and these things become a pain to maintain on changes.
Somehow I'd like to include the `go-foldersync` tests in this repo... would make tests much more robust that this `e2e_counter` or something similar.

Two options:
1. Move the go-foldersync just to this repo and make it an example app.
2. Include in the test step of CI, a cycle new extra step of `git clone`, `replace` the dependency, run tests.

Not sure about it.

api/server.go
sanderpick: For the API, loopback seems like a more reasonable default.

examples/chat/main.go
sanderpick: I missed that one. Yep, not used... I must have forgotten to delete when tinkering with the example.

examples/e2e_counter/main.go
sanderpick: Good call. My vote would be for the first option as it would be easier for other to locally run those tests as well.

eventstore/model_test.go
sanderpick: Will update

eventstore/model_test.go
sanderpick: Will update

eventstore/model_test.go
sanderpick: Will update

examples/e2e_counter/main.go
jsign: Like the idea, will make the todo.

sanderpick:
@jsign inline with previous chats, I removed the default address handling as it was getting nasty. Actual programs can set flag/config defaults.

eventstore/model_test.go
sanderpick: I had to fiddled with the test logic a little to get it to pass correctly

PR #166, opened 2019-12-11
Make the API server addr configurable

abour:
When running a second instance on the same host, the application crashes because the API addr is already binded.


jsign:
Nice. I think it would be better to wait for #165 merge since @sanderpick unified all address expressions to `multiaddr`. Shouldn't be a problem to refix this change according to that.

abour:
Ok, for sure.

abour:
I think that with #165 this PR is now obsolete

sanderpick:
Hey @abour. Apologies for missing this earlier. Agreed, this should be doable now in master.

abour:
No problem !

PR #169, opened 2019-12-14
Adds dockerfile

andrewxhill:


common/version.go
```diff
@@ -0,0 +1,7 @@
+package common
+
+// Set by ldflags
+var GitCommit, GitBranch, GitState, GitSummary, BuildDate string
```
sanderpick: In go-textile, these would be populated by github.com/ahmetb/govvv, which we probably don't need/want here. I'm not sure the best way to do a version file these days... maybe @jsign has some opinions.

sanderpick:
Nice!

common/version.go
jsign: Not really strong ideas here. Didn't know about `govvv` but seems a helper for passing `-ldflags` to the linker.
I think the only improvement that comes to my mind is automating that `Version` variable with something like: `git tag | sort -V | tail -n 1`, to get the latest tagged version.

andrewxhill: okay, pushed up a change that drops common/version and just pulls from latest git tag!

Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+EXPOSE 9090
+EXPOSE 9091
+
+# This just makes sure that:
+# 1. There's an fs-repo, and initializes one if there isn't.
+# 2. The API and Gateway are accessible from outside the container.
+ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start_threads"]
+
+# Execute the daemon subcommand by default
+CMD ["daemon", "--repo=/data/threads", "--apiProxyAddr=/ip4/0.0.0.0/tcp/9091"]
```
sanderpick: The install method above will put this under bin as `daemon` (we should change that so it installs as `threadsd`, if this is working, it must be using the container script differently, as in, not passing a _sub_ command.

Makefile
```diff
@@ -0,0 +1,3 @@
+docker:
+	$(eval VERSION := $(shell git --no-pager describe --abbrev=0 --tags --always))
+	docker build -t go-textile-threads:$(VERSION:v%=%) .
```
sanderpick: 👍

bin/container_daemon
```diff
@@ -0,0 +1,20 @@
...
+	# restart script with new privileges
+	exec su-exec "$user" "$0" "$@"
+fi
+
+# if the first argument is daemon
+if [ "$1" = "daemon" ]; then
+	shift
+fi
+
+exec threads "$@"
```
sanderpick: I don't see how this would work since there's no executable named `threads`, right?

andrewxhill: it moved, https://github.com/textileio/go-textile-threads/pull/169/files#diff-3254677a7917c6c01f55212f86c57fbfR44

Dockerfile
andrewxhill: it does install as daemon, but then i move it to 'threads'. i ripped this from the other docker, where it will just be sure to ignore `daemon` if the user supplies it (or the default cmd does)

https://github.com/textileio/go-textile-threads/pull/169/files#diff-2d3d90a507605bc346c3abfa66979c19R17

bin/container_daemon
sanderpick: ah ok 👍

PR #172, opened 2019-12-16
nits: found during textile integration

sanderpick:
Just a couple small things I encountered when combining this lib with `textile`.

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -3,7 +3,6 @@ package client
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"reflect"
 
 	"github.com/mr-tron/base58"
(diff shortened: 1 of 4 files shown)


api/client/client.go
```diff
@@ -31,9 +31,12 @@ type ListenEvent struct {
 }
 
 // NewClient starts the client
-func NewClient(host string, port int) (*Client, error) {
-	url := fmt.Sprintf("%v:%v", host, port)
-	conn, err := grpc.Dial(url, grpc.WithInsecure())
+func NewClient(maddr ma.Multiaddr) (*Client, error) {
```
sanderpick: In keeping with using multiaddress everywhere...

api/server.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 var (
-	log = logging.Logger("api")
+	log = logging.Logger("threadsapi")
```
sanderpick: More specific :)

eventstore/options.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	defaultDatastorePath = "datastore"
+	defaultDatastorePath = "eventstore"
```
sanderpick: Also, more specific :)

api/client/client.go
carsonfarmer: Yeh really nice, I like this. Should just get this into our heads as the default always.

PR #173, opened 2019-12-17
Granular listen service and client

asutula:


api/client/client.go
```diff
@@ -26,8 +28,8 @@ type Client struct {
 
 // ListenEvent is used to send data or error values for Listen
 type ListenEvent struct {
-	data []byte
-	err  error
+	action es.Action
```
asutula: Feeling a little friction using the `es.Action` in client code because of the `EntityID` alias for `string`.

api/client/client.go
```diff
@@ -234,22 +236,45 @@ func (c *Client) WriteTransaction(storeID, modelName string) (*WriteTransaction,
 	return &WriteTransaction{client: client, storeID: storeID, modelName: modelName}, nil
 }
 
-// Listen provides an update whenever the specified model is updated
-func (c *Client) Listen(storeID, modelName, entityID string) (<-chan ListenEvent, func()) {
+// Listen provides an update whenever the specified store, model type, or model instance is updated
+func (c *Client) Listen(storeID string, listenOptions ...es.ListenOption) (<-chan ListenEvent, func(), error) {
```
asutula: Same friction here... `es.ListenOption`: if you want to specify an entity id, it needs to be an `EntityID`.

asutula: Realized this can return an error for errors that happen pre-request to the server. Makes more sense I think.

api/client/client.go
```diff
@@ -234,22 +236,45 @@ func (c *Client) WriteTransaction(storeID, modelName string) (*WriteTransaction,
...
+	}
+	stream, err := c.client.Listen(ctx, req)
+	if err != nil {
+		cancel()
+		return nil, nil, err
+	}
+	go func() {
+		defer close(channel)
+
+	L:
```
asutula: Label here to break out of the switch nested inside the for loop.

api/client/client.go
```diff
@@ -259,11 +284,27 @@ func (c *Client) Listen(storeID, modelName, entityID string) (<-chan ListenEvent
...
+			switch event.GetAction() {
+			case pb.ListenReply_CREATE:
+				actionType = es.ActionCreate
+			case pb.ListenReply_DELETE:
+				actionType = es.ActionDelete
+			case pb.ListenReply_SAVE:
+				actionType = es.ActionSave
+			default:
+				channel <- ListenEvent{err: fmt.Errorf("unknown listen reply action %v", event.GetAction())}
+				break L
```
asutula: Break all the way out of the for loop.

api/client/client_test.go
```diff
@@ -394,8 +394,15 @@ func TestListen(t *testing.T) {
 	err = client.ModelCreate(storeID, modelName, person)
 	checkErr(t, err)
 
-	channel, discard := client.Listen(storeID, modelName, person.ID)
+	opt := es.ListenOption{
+		Model: modelName,
+		ID:    store.EntityID(person.ID),
```
asutula: Here is that friction in action.

api/client/client_test.go
```diff
@@ -412,12 +419,15 @@ func TestListen(t *testing.T) {
...
-		}
-		if p.Age != 30 {
-			t.Fatalf("expected listen result age = 30 but got: %v", p.Age)
+		// p := &Person{}
+		// if err := json.Unmarshal(val.data, p); err != nil {
+		// 	t.Fatalf("failed to unmarshal listen result: %v", err)
+		// }
+		// if p.Age != 30 {
+		// 	t.Fatalf("expected listen result age = 30 but got: %v", p.Age)
+		// }
```
asutula: Possibly need to add some checks in here on the contents of the actual data. This will depend on how we decide to provide the data. Either we send it along in the event or we do a request for the data here from the client.

api/service.go
```diff
@@ -351,19 +364,26 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
...
+			var replyAction pb.ListenReply_Action
+			switch action.Type {
+			case es.ActionCreate:
+				replyAction = pb.ListenReply_CREATE
+			case es.ActionDelete:
+				replyAction = pb.ListenReply_DELETE
+			case es.ActionSave:
+				replyAction = pb.ListenReply_SAVE
+			}
+			// TODO: do we want to send the entity data in case of create and save?
```
asutula: We were previously sending the entity data with the event. Is this a good idea? If so, we could include it only for create and save events? Is this a bad idea since we don't know how much data it will be?

api/client/client.go
```diff
@@ -24,10 +25,51 @@ type Client struct {
...
+	ListenDelete
+)
+
+// ListenOption represents a filter to apply when listening for data updates
+type ListenOption struct {
+	Type     ListenActionType
+	Model    string
+	EntityID string
+}
+
```
asutula: Modeled this stuff in the client to avoid having to deal with the friction of using core `EntityID` type that is used on the server.

api/client/client.go
```diff
@@ -259,11 +324,28 @@ func (c *Client) Listen(storeID, modelName, entityID string) (<-chan ListenEvent
...
+				actionType = ActionSave
+			default:
+				channel <- ListenEvent{err: fmt.Errorf("unknown listen reply action %v", event.GetAction())}
+				break L
+			}
+			action := Action{
+				Model:    event.GetModelName(),
+				Type:     actionType,
+				EntityID: event.GetEntityID(),
+				Entity:   event.GetEntity(),
```
asutula: Sending the entity JSON bytes along in the action. This is about as good as we can do since we don't know what Go type the bytes represent (the user can register to listen for updates to many/different types of models).

api/client/client_test.go
```diff
@@ -413,12 +420,15 @@ func TestListen(t *testing.T) {
 			t.Fatalf("failed to receive first listen result: %v", val.err)
 		}
 		p := &Person{}
-		if err := json.Unmarshal(val.data, p); err != nil {
+		if err := json.Unmarshal(val.action.Entity, p); err != nil {
```
asutula: If we want to use the entity data, we have to unmarshal it ourselves.

api/client/client.go
asutula: Fixed this by modeling a similar struct in the client package that just uses `string`.

api/client/client.go
sanderpick: Nice, makes sense. Always feels a little odd to use server models in a client, even if you _could_.

api/client/client.go
sanderpick: Sounds handy.

api/client/client_test.go
asutula: Ended up deciding to send the entity data along with the event so this part of the test is re-enabled.

api/service.go
asutula: I decided yes, send the data.

api/client/client.go
sanderpick: Maybe a more descriptive label? Just a thought

api/pb/api.proto
```diff
@@ -148,12 +148,30 @@ message WriteTransactionReply {
...
+        string entityID = 2;
+        enum Action {
+            ALL = 0;
+            CREATE = 1;
+            SAVE = 2;
+            DELETE = 3;
+        }
+        Action action = 3;
+    }
+    repeated Filter filters = 2;
```
sanderpick: 👍

api/pb/api.proto
```diff
@@ -148,12 +148,30 @@ message WriteTransactionReply {
...
-    string entity = 1;
+    string modelName = 1;
+    string entityID = 2;
+    enum Action {
+        CREATE = 0;
+        SAVE = 1;
+        DELETE = 2;
+    }
+    Action action = 3;
+    bytes entity = 4;
```
sanderpick: Nice, makin it easy for the clients... this will be fun to play with.

api/client/client.go
asutula: Cool, done

Issue #175, opened 2019-12-18
Handle new-style PubSub topic handles 


sanderpick:
The new pubsub module deals in `Topic` handles, which are needed to publish as well as subscribe. The service layer will have to keep track of these.

PR #177, opened 2019-12-18
go-textile-threads -> go-threads

sanderpick:
Also,
- Cleans up some warnings from new modules
- Turns off pubsub subscriptions until correctly handled (#175)


eventstore/dispatcher.go
```diff
@@ -83,7 +83,9 @@ func (d *dispatcher) Dispatch(events []core.Event) error {
 			return err
 		}
 	}
-	txn.Commit()
+	if err := txn.Commit(); err != nil {
+		return err
+	}
```
jsign: :+1:

PR #180, opened 2019-12-19
manager: fix reloading stores

sanderpick:
Small one to fix store reloading when an entity does not yet exist.


sanderpick:
There are yep - they shouldn't be returned in the query though.

PR #187, opened 2019-12-31
Adds dart folder for pb builds and automated pub.dev publishing

andrewxhill:
simple dart package publishing requires some folder structure and a pubspec.  so i've

- modified pb makefile to add the *.dart files into a specific subfolder
- i've added a pubspec.yaml that contains the package information for publishing
- i've added a few other metadata files required by pub.dev when publishing a package

i've skipped any automation for now. but steps to update the published package are

```sh
cd api/pb
make clean
make all
cd dart/
# optional if you want to auto bump the version in pubspec
pub global activate pubspec_version
pubspec_version bump patch
# now push the new package
pub publish
```

current version built this way is now up to date with master,

https://pub.dev/packages/threads_client_grpc


api/pb/dart/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
sanderpick: When this is automated, we should just copy the license file from root instead of having it in the repo twice.

.gitignore
```diff
@@ -22,6 +22,10 @@ vendor/
 *.js
 *.d.ts
 
+# Dart PB Files
```
sanderpick: We should include the generated files too... no need to have them in source control.

api/pb/dart/pubspec.yaml
```diff
@@ -0,0 +1,13 @@
+name: threads_client_grpc
```
sanderpick: why not match the js pattern (`threads-client-pb`)?

.gitignore
andrewxhill: okay, wasn't certain on that with this change. will update

api/pb/dart/pubspec.yaml
andrewxhill: oop, i thought that was the pattern

andrewxhill: yeah, i just copied that pattern from the js npm file https://www.npmjs.com/package/@textile/threads-client-grpc

andrewxhill: looks like both exist on npm though...

sanderpick: oh right! forgot i changed it doh: https://github.com/textileio/go-threads/blob/master/api/pb/package.json#L2

PR #192, opened 2020-01-02
follows dart pattern of javascript/ for the js package build

andrewxhill:
just moves the JS proto lib build under a tidy folder.

one thing i couldn't tell, should this file be included in the ignore or is it a requirement of the parent lib so needed, https://github.com/textileio/go-threads/blob/master/api/pb/api.pb.go


api/pb/api.pb.go
```diff
@@ -3,13 +3,14 @@
 
 package api_pb
 
-import proto "github.com/golang/protobuf/proto"
```
sanderpick: Not sure why this is changing... what version of protoc are you on?

andrewxhill: oh interesting, didn't catch that

`libprotoc 3.9.1`

sanderpick:
Do you mean, should we ignore the go files? They're needed by the program after a `go get ...` unless you wanted to require the user to have `protoc` installed, etc. They should be included.

sanderpick:
FYI, I'm working on CI + release stuff now.

andrewxhill:
okay yep, that's what the question was. so that makes sense

api/pb/api.pb.go
andrewxhill: updated to latest `libprotoc 3.11.2` and got the same

andrewxhill: upgraded to latest golang (1.13.5) and got yet another update here

andrewxhill: sorry scratch that, no change by upgrading go. same output as my current pr

andrewxhill: it looks like you have something different than me, carson, aaron as your commits revert it each time, 

https://github.com/textileio/go-threads/commit/5a4faf64aa08f140e83e0cd772f8f992716edea1#diff-8ac45a22aaa1306eba53a5c9ed4c8ac1

https://github.com/textileio/go-threads/commit/e3b917775f58372efe066207c0e1cb6784342bb8#diff-8ac45a22aaa1306eba53a5c9ed4c8ac1

sanderpick: It's likely a version diff with `protoc-gen-go` since those live in GOPATH, not local mods. I'll make sure to `go get -u github.com/golang/protobuf/protoc-gen-go`.

PR #199, opened 2020-01-03
Update logging package

### LINKED ISSUE #197: Incompatibility with newer ipfs/go-log
This project uses `ipfs/go-log` 0.0.1 but adding the `filecoin` project is as a dependency to the mix of our projects is forcing an upgrade to `ipfs/go-log` 1.0.0 (it's not bleeding edge or anything, 2.0 is already out as well). This project is uses code from `whyrusleeping/go-logging` that isn't compatible with `ipfs/go-log` 1.0.0 and `whyrusleeping/go-logging` doesn't seem to be maintained.

The problems are here:

https://github.com/textileio/go-threads/blob/master/util/util.go#L177

and here

https://github.com/textileio/go-threads/blob/master/util/util.go#L178

asutula:
Updates to ipfs/go-log 1.0.0 and removes dependency on whyrusleeping/go-logging.

Fixes #197


util/util.go
```diff
@@ -172,23 +173,26 @@ func SetupDefaultLoggingConfig(repoPath string) {
 		MaxBackups: 3,
 		MaxAge:     30, // days
 	}
+	w := zapcore.AddSync(lj)
```
asutula: running into problems here. trying to configure rotating logs according to https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation but it's not clear to me how to integrate that into ipfs/go-log because I don't see how the ability to set log backends is exposed by ipfs/go-log.

asutula:
Close, but not quite.

asutula:
Related https://github.com/ipfs/go-log/issues/66

util/util.go
```diff
@@ -166,29 +165,24 @@ func DecodeKey(k string) (*sym.Key, error) {
...
+	os.Setenv("GOLOG_FILE", filepath.Join(repoPath, "log", "textile.log"))
+	logging.SetupLogging()
+	logging.SetAllLoggers(logging.LevelError)
 }
 
 // SetLogLevels sets levels for the given systems.
-func SetLogLevels(systems map[string]logger.Level) error {
+func SetLogLevels(systems map[string]logging.LogLevel) error {
 	for sys, level := range systems {
+		l := zapcore.Level(level)
```
asutula: here `level` is a `logging.LogLevel` which is just an alias for `zapcore.Level`, but I can't call any `zapcore.Level` directly on it. I have to explicitly create a `zapcore.Level` from the `logging.LogLevel`. Expected behavior?

util/util.go
```diff
@@ -166,29 +165,24 @@ func DecodeKey(k string) (*sym.Key, error) {
...
-		MaxSize:    10, // megabytes
-		MaxBackups: 3,
-		MaxAge:     30, // days
-	}
-	backendFile := logger.NewLogBackend(lj, "", 0)
-	logger.SetBackend(backendFile)
-	logger.SetFormatter(logger.MustStringFormatter(logging.LogFormats["color"]))
-	logging.SetAllLoggers(logger.ERROR)
+	os.Setenv("GOLOG_LOG_FMT", "color")
+	os.Setenv("GOLOG_FILE", filepath.Join(repoPath, "log", "textile.log"))
```
asutula: This will write a log to the filesystem, but not using lumberjack and with no log rotation.

PR #204, opened 2020-01-07
Adds support for websocket connections and cors

carsonfarmer:
This allows us to use web sockets for connection from browsers and nodejs.


sanderpick:
@carsonfarmer can you merge this before @andrewxhill does a release?

PR #209, opened 2020-01-10
[wip] Don't GetRecord on initial Record

carsonfarmer:
This is a simple check that doesn't `GetRecord` if the given CID is from the initial Record. This makes it possible for remote peers to post 'cached' Blocks within the incoming Record, avoiding having to get the Block from the DAG Service. This is particularly useful for things like chat, and other small messages. It also means, peers without access to IPFS can essentially send along full Blocks. This is also needed as part of our ongoing remote JS-Threads implementation.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -590,9 +590,14 @@ func (t *service) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec 
 		if exist {
 			break
 		}
-		r, err := t.GetRecord(ctx, id, c)
-		if err != nil {
-			return err
+		var r core.Record
+		if c.String() != rec.Cid().String() {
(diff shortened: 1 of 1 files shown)


sanderpick:
Looks good! Just to clarify, this assumes the incoming record is _always_ cached internally, but not missing older records. This must have been working in the past since all peers are connected.

sanderpick:
@carsonfarmer do the tests pass locally?

carsonfarmer:
I thought they were, but just ran it again, and that failure does appear to be related to this PR. Investigating now.

carsonfarmer:
Fixes moved to #216, closing this in favour of that.

PR #211, opened 2020-01-11
Server and client opts

sanderpick:
- Cleans up / updates Dockerfile with latest from `textile`
- Adds gRPC server opts, allowing users to supply things like interceptor functions / other opts they may want
- Adds gRPC dial opts, allowing users to supply RPC credentials / other opts they may want


api/client/client_test.go
```diff
@@ -469,6 +472,15 @@ func makeServer() (*api.Server, func()) {
 		panic(err)
 	}
 	ts.Bootstrap(util.DefaultBoostrapPeers())
+
+	port, err := freeport.GetFreePort()
```
carsonfarmer: oooh, nice

PR #218, opened 2020-01-14
Service API

### LINKED ISSUE #213: Add missing service APIs to the gRPC API
- AddThread
- AddFollower
- maybe others

sanderpick:
Closes #213 

This adds yet another gRPC service and client, this time against the thread service. 

Additional changes:
- Cleans up the top-level API client and p2p service layer tests, enabling them to run in parallel and/or independently.
- Splits the core service interface into `Service`, which provides a host and DAG service, and the actual API, which is fulfilled by core service and the new client. The JavaScript implementation can copy the Go client, providing another API.
- Removes `logstore` from the core `Service` interface.
- Adds additional methods to the `API` interface: `CreateThread` and `GetThread`. Not having these was forcing a bunch of annoying patterns when creating and fetching threads and logs, and necessitated having (dangerous access) to the `logstore`.
- `ThreadInfo` now contains all log information, not just IDs. In almost every case, the log info was fetched anyway. Now that the `logstore` isn't on the API, this allows users to access that info.


Dockerfile
```diff
@@ -51,12 +51,14 @@ COPY --from=0 /lib/x86_64-linux-gnu/libdl.so.2 /lib/libdl.so.2
...
 EXPOSE 5006
-# apiAddr; should be exposed to the public
+# serviceApiProxyAddr; should *not* be exposed to the public unless intercepted by an auth system, e.g., textile
+EXPOSE 5007
+# apiAddr; should *not* be exposed to the public unless intercepted by an auth system, e.g., textile
 EXPOSE 6006
-# apiProxyAddr; should be exposed to the public
-EXPOSE 7006
+# apiProxyAddr; should *not* be exposed to the public unless intercepted by an auth system, e.g., textile
+EXPOSE 6007
```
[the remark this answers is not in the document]
sanderpick: They'll probably need to be updated anyway... I think we're still early enough to make these kind of changes. Having them like this feels more intuitive to me.

service/api/client/client.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+func (c *Client) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.AddOption) (info thread.Info, err error) {
+	args := &core.AddOptions{}
+	for _, opt := range opts {
+		opt(args)
+	}
+	resp, err := c.c.AddThread(ctx, &pb.AddThreadRequest{
+		Addr:      addr.Bytes(),
+		ReadKey:   args.ReadKey.Bytes(),
```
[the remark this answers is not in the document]
sanderpick: AddThread is for adding an existing thread... this can be done without the read key. I think what you want is a modification / addition to create thread. Maybe we can change it so CreateThread takes keys instead of making them internally.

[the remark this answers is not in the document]
carsonfarmer: I see, yes that makes sense. And yes, this is indeed needed. If CreateThread could take keys directly, this would solve that use case of keeping keys 'local' 👍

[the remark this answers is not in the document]
sanderpick: Ok, cool. I think we'll also need an additional method to add an already encrypted event. Maybe that's what AddRecord should be and a new CreateRecord becomes the old AddRecord.

service/service.go
```diff
@@ -213,6 +165,35 @@ func (t *service) Store() lstore.Logstore {
...
+		return
+	}
+	if err = t.store.AddThread(info); err != nil {
+		return
+	}
+	linfo, err := createLog(t.host.ID())
+	if err != nil {
+		return
+	}
+	if err = t.store.AddLog(id, linfo); err != nil {
```
sanderpick: Now that creating a thread is actually part of the API, I added this to ensure that a log is created for the host at creation time. Not having this always felt a little weird, and I got bit by it again debugging a new AddFollower test. This would fail silently if the creator hadn't added a record yet, because there was not logs to send the follower.

service/api/client/client.go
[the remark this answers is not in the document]
carsonfarmer: Awesome, agree re: names. AddRecord for adding existing (encrypted/structured) Record, CreateRecord for taking 'raw' input and bundling it up automagically.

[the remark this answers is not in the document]
sanderpick: So, 
```
AddRecord(ctx context.Context, id thread.ID, event Event) (ThreadRecord, error)
```
That way, the client doesn't need to get the prevID and sign the record. They just give the encrypted event. Sound okay?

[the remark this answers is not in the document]
carsonfarmer: Excellent! Yup makes sense. Actually, this is fantastic, because the signature stuff is probs the most tricky piece of the whole dance in the browser.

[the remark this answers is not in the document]
sanderpick: Ok, cool. Though, thinking about it some more, maybe you do want to handle that piece? That way the service doesn't need the private key? I could see it either way.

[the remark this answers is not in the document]
carsonfarmer: Ah right, duh... no you're right... we don't want the service to have anything beyond the follower key, that's the whole point of the JS stuff really. So we'll still need to do the signature stuff in browser. That's _fine_, the code is already there to do that.

service/api/client/client.go
```diff
@@ -0,0 +1,282 @@
...
+		defer close(channel)
+		for {
+			resp, err := stream.Recv()
+			if err == io.EOF {
+				return
+			}
+			if err != nil {
+				stat := status.Convert(err)
+				if stat.Code() != codes.Canceled {
+					log.Fatalf("error in subscription stream: %v", err)
```
jsign: Maybe not fatal?
I think there're some more there, so may apply to other lines.

sanderpick: Sure, sounds good

sanderpick:
> Wires everywere :P

@jsign, yeah... apologies! Ballooned

cbor/event.go
```diff
@@ -83,6 +84,7 @@ func CreateEvent(
 			Node: codedHeader,
 			obj:  eventHeader,
 		},
+		body: codedBody,
```
sanderpick: Not sure why this was never included in the object returned from `CreateEvent`. Should lead to less trips to the `DAGService`.

core/service/options.go
```diff
@@ -1,33 +1,45 @@
...
 	"github.com/textileio/go-threads/crypto/symmetric"
 )
 
-// AddOptions defines options for adding a new thread.
-type AddOptions struct {
+// KeyOptions defines options for keys when creating / adding a thread.
+type KeyOptions struct {
 	FollowKey *symmetric.Key
 	ReadKey   *symmetric.Key
+	LogKey    crypto.Key
```
sanderpick: This allows a client to dictate the log private key. Can be either private or public.

core/service/options.go
```diff
@@ -1,33 +1,45 @@
...
+	return func(args *KeyOptions) {
 		args.ReadKey = key
 	}
 }
 
+// LogKey defines the public or private key used to write a log records.
+// If this is just a public key, the service itself won't be able to create records.
+// In other words, all records must pre-created and added with AddRecord.
+// If no log key is provided, one will be created internally.
+func LogKey(key crypto.Key) KeyOption {
```
sanderpick: ^ Here's a full note on how the key is used.

core/service/service.go
```diff
@@ -48,8 +48,11 @@ type API interface {
 	// AddFollower to a thread.
 	AddFollower(ctx context.Context, id thread.ID, paddr ma.Multiaddr) (peer.ID, error)
 
-	// AddRecord with body.
-	AddRecord(ctx context.Context, id thread.ID, body format.Node) (ThreadRecord, error)
+	// CreateRecord with body.
+	CreateRecord(ctx context.Context, id thread.ID, body format.Node) (ThreadRecord, error)
+
+	// AddRecord to the given log.
+	AddRecord(ctx context.Context, id thread.ID, lid peer.ID, rec Record) error
```
sanderpick: `AddRecord` takes a fully formed record.

sanderpick: (It must be signed by the log's private key)

service/api/client/client.go
```diff
@@ -60,30 +60,60 @@ func (c *Client) GetHostID(ctx context.Context) (peer.ID, error) {
...
+	}
+	if args.ReadKey != nil {
+		req.Keys.ReadKey = args.ReadKey.Bytes()
+	}
+	if args.LogKey != nil {
+		req.Keys.LogKey, err = args.LogKey.Bytes()
+		if err != nil {
+			return
+		}
+	}
```
sanderpick: ^ All key combinations are possible. If you don't provide a follow key, you'll get errorswhen trying add records. In this sense, it's not "optional", but it's a tradeoff for a nicer API.

service/api/client/client.go
```diff
@@ -158,6 +210,7 @@ func (c *Client) Subscribe(ctx context.Context, opts ...core.SubOption) (<-chan
 	if err != nil {
 		return nil, err
 	}
+	threads := make(map[thread.ID]*symmetric.Key) // Follow-key cache
```
sanderpick: As records come in, we need to fetch the follow key for the thread.... so, keeping a cache of those keys here. See below...

service/api/client/client_test.go
```diff
@@ -172,16 +185,67 @@ func TestClient_AddRecord(t *testing.T) {
...
+			t.Fatal("got bad thread ID from create record")
 		}
 		if rec.LogID().String() == "" {
-			t.Fatal("got bad log ID from add record")
+			t.Fatal("got bad log ID from create record")
+		}
+	})
+}
+
+func TestClient_AddRecord(t *testing.T) {
```
sanderpick: Here's a full AddRecord test. Only the client has the read key and the log private key.

service/api/client/client_test.go
```diff
@@ -172,16 +185,67 @@ func TestClient_AddRecord(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("test add record", func(t *testing.T) {
+		rk, err := symmetric.CreateKey()
+		if err != nil {
+			t.Fatal(err)
+		}
+		event, err := cbor.CreateEvent(context.Background(), nil, body, rk)
```
sanderpick: The tests use the cbor event and record methods, which take a DAGService, which adds the nodes. The client doesn't care about that since the remote service will do this anyway. So, here we pass nil for the DAGService.

service/api/client/client_test.go
```diff
@@ -199,7 +263,7 @@ func TestClient_GetRecord(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	rec, err := client.AddRecord(context.Background(), info.ID, body)
+	rec, err := client.CreateRecord(context.Background(), info.ID, body)
```
sanderpick: CreateRecord is the old AddRecord.

service/api/service.go
```diff
@@ -143,17 +141,46 @@ func (s *service) AddRecord(ctx context.Context, req *pb.AddRecordRequest) (*pb.
...
+
+	threadID, err := thread.Cast(req.ThreadID)
+	if err != nil {
+		return nil, err
+	}
+	info, err := s.s.GetThread(ctx, threadID)
+	if err != nil {
+		return nil, err
+	}
+	logID, err := peer.IDFromBytes(req.LogID)
```
sanderpick: The client tells us what logID to add the record to. The record will have a valid signature if they have "permission" to do this.

service/service.go
```diff
@@ -165,26 +165,33 @@ func (t *service) Store() lstore.Logstore {
...
-	info.ReadKey, err = sym.CreateKey()
-	if err != nil {
-		return
+
+	info = thread.Info{
+		ID:        id,
+		FollowKey: args.FollowKey,
+		ReadKey:   args.ReadKey,
+	}
+	if info.FollowKey == nil {
```
sanderpick: If there's no follow key given with `CreateThread`, the service will create one for you.

service/service.go
```diff
@@ -473,6 +480,11 @@ func (t *service) AddRecord(ctx context.Context, id thread.ID, body format.Node)
 		return
 	}
 
+	// Update head
+	if err = t.store.SetHead(id, lg.ID, rec.Cid()); err != nil {
```
sanderpick: Just moved this to a more consistent location.

service/service.go
```diff
@@ -489,6 +501,32 @@ func (t *service) AddRecord(ctx context.Context, id thread.ID, body format.Node)
...
+	knownRecord, err := t.bstore.Has(rec.Cid())
+	if err != nil {
+		return err
+	}
+	if knownRecord {
+		return nil
+	}
+
+	// Verify node
+	if err = rec.Verify(logpk); err != nil {
```
sanderpick: Verify the incoming record.

service/service.go
```diff
@@ -489,6 +501,32 @@ func (t *service) AddRecord(ctx context.Context, id thread.ID, body format.Node)
...
+		return err
+	}
+
+	return t.PutRecord(ctx, id, lid, rec)
```
sanderpick: And then here we just handle it like a record that came in over the net.

service/service.go
```diff
@@ -897,24 +928,26 @@ func (t *service) updateRecordsFromLog(tid thread.ID, lid peer.ID) {
...
 
 // createLog creates a new log with the given peer as host.
-func createLog(host peer.ID) (info thread.LogInfo, err error) {
-	sk, pk, err := crypto.GenerateEd25519Key(rand.Reader)
-	if err != nil {
-		return
+func createLog(host peer.ID, key crypto.Key) (info thread.LogInfo, err error) {
+	var ok bool
+	if key == nil {
+		info.PrivKey, info.PubKey, err = crypto.GenerateEd25519Key(rand.Reader)
```
sanderpick: Similar to thread follow key, a log private key is created on the service if _no_ log key was given (private / public)

service/util/util.go
```diff
@@ -0,0 +1,24 @@
+package util
```
sanderpick: This will have to be cleaned up at some point. I went back and forth a couple times trying to reuse a `pb.Record` from the p2p gRPC service and never got it to work right. Mostly a hassle trying to combine gogo and golang protos. In the end, having this annoying mapping between records was the lesser of two evils. I'd like to consider moving everything to gogo at some point since it's much faster at marshaling.

cbor/event.go
carsonfarmer: Nice

core/service/options.go
carsonfarmer: This is great @sanderpick. In addition to addressing this 'issue', it opens us up to a whole slew of possibilities re: identity sources, which is awesome. Essentially, anything that satisfies this interface will do, which is pretty awesome, good idea!

service/api/client/client_test.go
carsonfarmer: Beautiful! Aside from the LogKey, this is quite possibly the best part about this new API. Making explicit that you are _not_ responsible for adding the content to IPFS is really nice here. It also makes extending this to a full peer with a DAG Service _later_ sooo much easier (conceptually).

service/service.go
carsonfarmer: Aha, ok this answers my pervious question I think...

service/service.go
```diff
@@ -499,46 +539,35 @@ func (t *service) GetRecord(ctx context.Context, id thread.ID, rid cid.Cid) (cor
...
-type record struct {
+// Record wraps a core.Record within a thread and log context.
+type Record struct {
 	core.Record
 	threadID thread.ID
 	logID    peer.ID
 }
 
+// NewRecord returns a record with the given values.
+func NewRecord(r core.Record, id thread.ID, lid peer.ID) core.ThreadRecord {
```
carsonfarmer: Yes, nice to have this as part of the public API (I mean its needed, but its also nice)

service/service.go
carsonfarmer: Ok, gotcha...

service/service.go
```diff
@@ -466,27 +480,53 @@ func (t *service) AddRecord(ctx context.Context, id thread.ID, body format.Node)
...
 
 	// Notify local listeners
-	r = &record{
-		Record:   rec,
-		threadID: id,
-		logID:    lg.ID,
-	}
-	if err = t.bus.SendWithTimeout(r, time.Second*3); err != nil {
+	r = NewRecord(rec, id, lg.ID)
+	if err = t.bus.SendWithTimeout(r, notifyTimeout); err != nil {
```
jsign: Every time I see the `SendWithTimeout` thing, something feels bad. If some receiver is too slow to read these things it will make this kind of important operations fail. From the subscriber POV, subscribing to something doesn't seem like a super important thing to handle well... and just a single slow receiver might block all operations.

I think ultimately is a problem of the bus API... I think there should be subscribers which can't miss any event in the bus and make the original operation fail (which is a quite aggressive policy, like in the Store since it can't miss an event), and maybe others that should be ignored if can't be pushed the new notification (but not make the original operation fail). Not sure would be a real solution, just first thought.

Not something we should fix now... but this thing seems to be waiting to bite us.

service/service.go
jsign: Finally :)

service/service.go
```diff
@@ -549,28 +578,34 @@ func (t *service) Subscribe(opts ...core.SubOption) core.Subscription {
...
-			if ok {
-				if len(filter) > 0 {
-					if _, ok := filter[r.threadID]; ok {
-						listener.ch <- r
+		listener := t.bus.Listen()
+		for {
+			select {
+			case <-ctx.Done():
+				listener.Discard()
+				close(channel)
```
jsign: Should this two lines be _defered_ at first LOC of the goroutine to cover the other `case`, or isn't needed?

core/service/options.go
sanderpick: Yep, correct. LogK is used to validate records.

service/api/client/client.go
sanderpick: If _no_ `LogKey`, then a private one is created internally. So, if you didn't supply a `ReadKey`, the remote would still not be able to write records. 

There's a matrix of possibilities here. I opted for all keys just being options w/ somewhat predictable consequences when omitted. Though, it's definitely not ideal. Open to alternatives here.

service/api/client/client.go
```diff
@@ -0,0 +1,330 @@
...
+				info, err := c.GetThread(ctx, threadID)
+				if err != nil {
+					log.Fatal(err)
+				}
+				fk = info.FollowKey
+				threads[threadID] = fk
+			}
+			rec, err := threadRecordFromProto(resp, fk)
+			if err != nil {
+				log.Fatal(err)
```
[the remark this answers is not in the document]
sanderpick: Ah, forgot to change, cheers.

service/api/client/client.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	if args.ReadKey != nil {
+		req.Keys.ReadKey = args.ReadKey.Bytes()
+	}
+	if args.LogKey != nil {
+		req.Keys.LogKey, err = args.LogKey.Bytes()
+		if err != nil {
+			return
+		}
+	}
```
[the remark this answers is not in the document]
sanderpick: Good call. Will make an issue.

cbor/event.go
```diff
@@ -66,9 +71,10 @@ func NewEvent(ctx context.Context, dag format.DAGService, body format.Node, rkey
 		return nil, err
 	}
 
-	err = dag.AddMany(ctx, []format.Node{node, codedHeader, codedBody})
-	if err != nil {
-		return nil, err
+	if dag != nil {
```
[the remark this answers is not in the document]
sanderpick: Nice, I like that.

service/api/service.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid read-key: %v", err)
+		}
+		opts = append(opts, core.ReadKey(rk))
+	}
+	var lk crypto.Key
+	if keys.LogKey != nil {
+		lk, err = crypto.UnmarshalPrivateKey(keys.LogKey)
+		if err != nil {
+			lk, err = crypto.UnmarshalPublicKey(keys.LogKey)
```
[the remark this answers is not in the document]
sanderpick: Yep, multi format helping out!

service/api/client/client.go
jsign: Right, just trying to confirm if it was reasonable case or not. I'm OK with making this simple since combinations are a bunch to cover.

service/service.go
sanderpick: Yep, good catch!

service/api/client/client.go
```diff
@@ -0,0 +1,330 @@
...
+				log.Fatal(err)
+			}
+			var fk *symmetric.Key
+			var ok bool
+			if fk, ok = threads[threadID]; !ok {
+				info, err := c.GetThread(ctx, threadID)
+				if err != nil {
+					log.Fatal(err)
+				}
+				fk = info.FollowKey
```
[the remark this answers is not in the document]
sanderpick: Indeed, adding

service/api/client/client.go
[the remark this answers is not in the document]
sanderpick: Eh, just made the quick change

PR #219, opened 2020-01-15
Proposed changes to simplify Event Codec interface

carsonfarmer:
Nothing ground-breaking here. Just moved all the datastore-based logic from EventCodec into the Store's Reduce method. It does however, add a new method to the Event interface. Note that the Store is actually the only place this is used, so I think think 'breaking' change is probably ok? In doing this, it simplifies things a bit for the Store's Reducer.

I'll also leave some 'inline' comments for things I'd like some clarification on.

### DIFF
--- a/jsonpatcher/jsonpatcher.go
+++ b/jsonpatcher/jsonpatcher.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
-	ds "github.com/ipfs/go-datastore"
 	cbornode "github.com/ipfs/go-ipld-cbor"
 	format "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log"
@@ -17,23 +16,13 @@ import (
 	core "github.com/textileio/go-threads/core/store"
 )
 
-type operationType int
-
-const (
-	create operationType = iota
-	save
-	delete
-)
-
 var (
-	log                           = logging.Logger("jsonpatcher")
-	errSavingNonExistentInstance  = errors.New("can't save nonexistent instance")
-	errCantCreateExistingInstance = errors.New("cant't create already existent instance")
-	errUnknownOperation           = errors.New("unknown operation type")
+	log                 = logging.Logger("jsonpatcher")
+	errUnknownOperation = errors.New("unknown operation type")
 )
 
 type operation struct {
-	Type      operationType
+	Type      core.ActionType
 	EntityID  core.EntityID
 	JSONPatch []byte
 }
@@ -91,66 +80,37 @@ func (jp *jsonPatcher) Create(actions []core.Action) ([]core.Event, format.Node,
 	return events, n, nil
 }
 
-func (jp *jsonPatcher) Reduce(events []core.Event, datastore ds.TxnDatastore, baseKey ds.Key) ([]core.ReduceAction, error) {
-	txn, err := datastore.NewTransaction(false)
-	if err != nil {
-		return nil, err
+func (jp *jsonPatcher) Reduce(e core.Event, oldState []byte) (*core.CodecResult, error) {
+	je, ok := e.(patchEvent)
+	if !ok {
+		return nil, fmt.Errorf("event unrecognized for jsonpatcher eventcodec")
 	}
-	defer txn.Discard()
-
-	actions := make([]core.ReduceAction, len(events))
-	for i, e := range events {
-		je, ok := e.(patchEvent)
-		if !ok {
-			return nil, fmt.Errorf("event unrecognized for jsonpatcher eventcodec")
+	switch je.Patch.Type {
+	case core.Create:
+		action := core.ReduceAction{Type: core.Create, Model: e.Model(), EntityID: e.EntityID()}
+		return &core.CodecResult{
+			Action: action,
+			State:  je.Patch.JSONPatch,
+		}, nil
+	case core.Save:
+		patchedValue, err := jsonpatch.MergePatch(oldState, je.Patch.JSONPatch)
+		if err != nil {
+			return nil, fmt.Errorf("error when reducing save event: %v", err)
 		}
-		key := baseKey.ChildString(e.Model()).ChildString(e.EntityID().String())
-		switch je.Patch.Type {
-		case create:
-			exist, err := txn.Has(key)
-			if err != nil {
-				return nil, err
-			}
-			if exist {
-				return nil, errCantCreateExistingInstance
-			}
-			if err := txn.Put(key, je.Patch.JSONPatch); err != nil {
-				return nil, fmt.Errorf("error when reducing create event: %v", err)
-			}
-			actions[i] = core.ReduceAction{Type: core.Create, Model: e.Model(), EntityID: e.EntityID()}
-			log.Debug("\tcreate operation applied")
-		case save:
-			value, err := txn.Get(key)
-			if errors.Is(err, ds.ErrNotFound) {
-				return nil, errSavingNonExistentInstance
-			}
-			if err != nil {
-				return nil, err
-			}
-			patchedValue, err := jsonpatch.MergePatch(value, je.Patch.JSONPatch)
-			if err != nil {
-				return nil, fmt.Errorf("error when reducing save event: %v", err)
-			}
-			if err = txn.Put(key, patchedValue); err != nil {
-				return nil, err
-			}
-			actions[i] = core.ReduceAction{Type: core.Save, Model: e.Model(), EntityID: e.EntityID()}
-			log.Debug("\tsave operation applied")
-		case delete:
-			if err := txn.Delete(key); err != nil {
-				return nil, err
-			}
-			actions[i] = core.ReduceAction{Type: core.Delete, Model: e.Model(), EntityID: e.EntityID()}
-			log.Debug("\tdelete operation applied")
-		default:
-			return nil, errUnknownOperation
-		}
+		action := core.ReduceAction{Type: core.Save, Model: e.Model(), EntityID: e.EntityID()}
+		return &core.CodecResult{
+			Action: action,
+			State:  patchedValue,
+		}, nil
+	case core.Delete:
+		action := core.ReduceAction{Type: core.Delete, Model: e.Model(), EntityID: e.EntityID()}
+		return &core.CodecResult{
+			Action: action,
+			State:  nil,
+		}, nil
+	default:
+		return nil, errUnknownOperation
 	}
-	if err := txn.Commit(); err != nil {
-		return nil, err
-	}
-
-	return actions, nil
 }
 
 type recordEvents struct {
@@ -186,7 +146,7 @@ func createEvent(id core.EntityID, v interface{}, jsonMode bool) (*operation, er
 		}
 	}
 	return &operation{
-		Type:      create,
+		Type:      core.Create,
 		EntityID:  id,
 		JSONPatch: opBytes,
 	}, nil
@@ -215,7 +175,7 @@ func saveEvent(id core.EntityID, prev interface{}, curr interface{}, jsonMode bo
 		return nil, err
 	}
 	return &operation{
-		Type:      save,
+		Type:      core.Save,
 		EntityID:  id,
 		JSONPatch: jsonPatch,
 	}, nil
@@ -223,7 +183,7 @@ func saveEvent(id core.EntityID, prev interface{}, curr interface{}, jsonMode bo
 
 func deleteEvent(id core.EntityID) (*operation, error) {
 	return &operation{
-		Type:      delete,
+		Type:      core.Delete,
 		EntityID:  id,
 		JSONPatch: nil,
 	}, nil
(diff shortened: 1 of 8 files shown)


jsonpatcher/jsonpatcher.go
```diff
@@ -9,31 +9,20 @@ import (
...
-	errSavingNonExistentInstance  = errors.New("can't save nonexistent instance")
-	errCantCreateExistingInstance = errors.New("cant't create already existent instance")
-	errUnknownOperation           = errors.New("unknown operation type")
+	log                 = logging.Logger("jsonpatcher")
+	errUnknownOperation = errors.New("unknown operation type")
 )
 
 type operation struct {
-	Type      operationType
+	Type      core.ActionType
```
carsonfarmer: Switched to using core.ActionType here, as there are multiple nearly identical versions of ActionType floating around, likely due to past scenarios where the event store was its own package.

jsonpatcher/jsonpatcher.go
```diff
@@ -91,66 +80,37 @@ func (jp *jsonPatcher) Create(actions []core.Action) ([]core.Event, format.Node,
...
-			log.Debug("\tsave operation applied")
-		case delete:
-			if err := txn.Delete(key); err != nil {
-				return nil, err
-			}
-			actions[i] = core.ReduceAction{Type: core.Delete, Model: e.Model(), EntityID: e.EntityID()}
-			log.Debug("\tdelete operation applied")
-		default:
-			return nil, errUnknownOperation
+	switch je.Patch.Type {
```
carsonfarmer: So now, we essentially don't mutate _any_ state within the codec. This is now just a pure function that takes the initial state, and returned an updated state. This is slightly more akin to something like Redux in that sense...
I think this makes it easier for external devs to implement their own EventCodecs...

jsonpatcher/jsonpatcher.go
```diff
@@ -252,4 +212,8 @@ func (je patchEvent) Model() string {
 	return je.ModelName
 }
 
+func (je patchEvent) Type() core.ActionType {
```
carsonfarmer: This has been added to the core Event interface.

store/dispatcher_test.go
```diff
@@ -157,5 +157,9 @@ func (n *nullEvent) Node() (format.Node, error) {
 	return nil, nil
 }
 
+func (n *nullEvent) Type() core.ActionType {
```
carsonfarmer: Just returning Delete for now. Perhaps it is worth defining a `core.Noop` ActionType though for future usage, tests, etc?

store/store.go
```diff
@@ -40,6 +40,9 @@ var (
 	dsStorePrefix   = ds.NewKey("/store")
 	dsStoreThreadID = dsStorePrefix.ChildString("threadid")
 	dsStoreSchemas  = dsStorePrefix.ChildString("schema")
+
+	errSavingNonExistentInstance = errors.New("can't save nonexistent instance")
```
carsonfarmer: Pulled over from jsonpatcher.

store/store.go
```diff
@@ -261,24 +264,66 @@ func (s *Store) GetModel(name string) *Model {
...
-	codecActions, err := s.eventcodec.Reduce(events, s.datastore, baseKey)
+	txn, err := s.datastore.NewTransaction(false)
 	if err != nil {
 		return err
 	}
-	actions := make([]Action, len(codecActions))
-	for i, ca := range codecActions {
+	defer txn.Discard()
+
+	actions := make([]Action, len(events))
```
carsonfarmer: So now all the datastore-based logic is in here, rather than in the EventCodec. This is an initial step towards making it easier to do things like indexing, without requiring that a developer implements indexing within the EventCodec (which didn't feel right).

store/store.go
```diff
@@ -261,24 +264,66 @@ func (s *Store) GetModel(name string) *Model {
...
 		case core.Create:
 			actionType = ActionCreate
+			exist, err := txn.Has(key)
+			if err != nil {
+				return err
+			}
+			if exist {
+				return errCantCreateExistingInstance
+			}
+			ca, err = s.eventcodec.Reduce(e, nil)
```
carsonfarmer: As a result, Reduce now runs on each individual Event, rather than the array. They all still happen within a transaction though, so that safety is maintained.

store/store.go
```diff
@@ -261,24 +264,66 @@ func (s *Store) GetModel(name string) *Model {
...
+			}
+			if err != nil {
+				return err
+			}
+			ca, err = s.eventcodec.Reduce(e, value)
+			if err = txn.Put(key, ca.State); err != nil {
+				return err
+			}
+			log.Debug("\tsave operation applied")
 		case core.Delete:
```
carsonfarmer: Both Create and Delete don't actually _need_ the previous state... so we _could_ probably collapse them together is this switch statement...

store/store.go
```diff
@@ -261,24 +264,66 @@ func (s *Store) GetModel(name string) *Model {
...
 		return err
 	}
-	actions := make([]Action, len(codecActions))
-	for i, ca := range codecActions {
+	defer txn.Discard()
+
+	actions := make([]Action, len(events))
+	for i, e := range events {
+		key := baseKey.ChildString(e.Model()).ChildString(e.EntityID().String())
 		var actionType ActionType
```
carsonfarmer: `store.Action` (as used by Listeners) has its own `ActionType` again, is this intentional, or could we also use core.ActionType here as well? It would simplify things a bit it we stuck to a single ActionType, and we actually wouldn't need to have the Type on as many structs/interfaces that way?

store/store.go
```diff
@@ -261,24 +264,66 @@ func (s *Store) GetModel(name string) *Model {
...
+			}
+			if err := txn.Delete(key); err != nil {
+				return err
+			}
+			log.Debug("\tdelete operation applied")
 		default:
 			panic("eventcodec action not recognized")
 		}
-		actions[i] = Action{Model: ca.Model, Type: actionType, ID: ca.EntityID}
+		actions[i] = Action{Model: ca.Action.Model, Type: actionType, ID: ca.Action.EntityID}
```
carsonfarmer: See my comment above about consolidating the `ActionType`s used here.

jsonpatcher/jsonpatcher.go
sanderpick: Makes sense to me. Let's keep the two types here @carsonfarmer.

jsonpatcher/jsonpatcher.go
sanderpick: +1

store/store.go
sanderpick: I _think_ the intention was to put indexing outside of the codec, right @carsonfarmer? I don't think these changes here are needed for that work, unless I am misunderstanding...

jsonpatcher/jsonpatcher.go
carsonfarmer: Ok, that explanation makes sense. Will revert.

jsonpatcher/jsonpatcher.go
carsonfarmer: Ok, so what's a viable alternative here then? Thoughts?

store/store.go
carsonfarmer: Well, the reason it seemed like the indexing would need to happen in the Codec was that to index the data, we need access to the 'raw' data. In theory, only the Model and Codec really know anything about that data. We only want to index valid data, so it has to happen _in_ or _after_ the Codec has done its work. It didn't seem like a good place to have _in_ the codec (as you point out), but ideally the data should be indexed in the same transaction that is being used to add the 'raw' data to the datastore. This currently happens in the Codec.
As a side note, it wasn't clear to me _why_ the Codec should need to know anything about the datastore per se in the first place? Hence pulling out that transaction logic into the Store in the proposal.

carsonfarmer:
So actually, at the end of the day, I think the 'controversial' (not actually) changes can be reverted, and we'll end up with something we can agreed on? I'll push those reverts and let's see how that looks?

store/store.go
sanderpick: The raw data could be retrieved directly from the datastore. I can't remember if it's CBOR or JSON. Maybe it depends on JSON mode. In any case, I don't see why it couldn't be loaded and indexed in a generic way after the codec has done it's job.

sanderpick:
Sounds good! The idea of removing datastore knowledge from the codec SGTM, but have to defer to @jsign there as I haven't had any part in this code.

store/store.go
carsonfarmer: Agreed, it does seem like this is the right approach. I'm not proposing the indexing happens in the Codec, I'm proposing it be done outside as you suggest. What I _am_ proposing here is that, it could be done in the same transaction (which it should) if we move that transaction logic _out_ of the Codec and _into_ the Store. Make sense?

carsonfarmer: (Also yes you are right, it depends on the 'mode'.)

sanderpick: Ok, cool. Sounds good 👍

jsonpatcher/jsonpatcher.go
```diff
@@ -9,14 +9,15 @@ import (
...
-	ds "github.com/ipfs/go-datastore"
 	cbornode "github.com/ipfs/go-ipld-cbor"
 	format "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log"
 	"github.com/multiformats/go-multihash"
 	core "github.com/textileio/go-threads/core/store"
 )
 
+// operationType is the type of the operation.
+// It is intentionally unique from core.ActionType to provide greater flexibility.
```
carsonfarmer: To quell future discussion about consolidating *Types ;)

store/listeners.go
```diff
@@ -43,34 +43,43 @@ func (s *Store) notifyTxnEvents(node format.Node) error {
 	return s.localEventsBus.send(node)
 }
 
+// ActionType is the type of the action.
```
carsonfarmer: To quell future discussion about consolidating *Types ;)

store/store.go
jsign: > As a side note, it wasn't clear to me _why_ the Codec should need to know anything about the datastore per se in the first place? Hence pulling out that transaction logic into the Store in the proposal.

I think most of the discussion relates on how much responsibility and power is given to `EventCodec`s. From my POV, the event codec should transform entity changes in a txn, to whatever the EventCodec decides it's the best way to express it and apply that to the model state (hence, works on the datastore too).

With your change, the EventCodec returns the new state to the Store and ask the Store to execute that change in the datastore. So, it will _reply_ something like: `Create entity with this bytes`. What I think of this is that it adds more constraints to what the EventCodec can do since it must frame every result on returning: bytes + (Create || Delete || Update).

In the same vein, part of the idea of the Reducer receive a slice of Events instead of one-by-one (as in your change) is that it may allow the EventCodec to apply multiple changes of the same txn more efficiently (maybe it won't need a datastore txn all the way, now it's forced). Also, an EventCodec could decide to create a single operation that modifies multiple entities (since it compressed txn changes in a single codec operation)... and this can't be possible now since it must return: bytes + (Create || Delete || Update) for a *single* entity.

As I said earlier, I'm not against reducing the scope/power of what EventCodecs can do. But just trying to mention tradeoffs.

carsonfarmer: All fair points. Based on our offline discussion, we'll be taking a different approach. Though one thing is, actually the only thing the EventCodec really _needs_ to return is the new state (and the action if we want). It doesn't need to specify _what_ to do with the state, the store just takes it, and applies it. If the state is nil, it deletes it, otherwise, it overwrites it. The Store then has no knowledge of _what_ the EventCodec did... so no new constraints are required (note that's not how I implemented it to keep things simple, but it could be that way easily). This _does_ however necessitate that each event is handled in turn, but that's currently how it works anyway so...

jsign: Well, that actually would take out the double-check that was done about not processing two times a `Create` event (which should be fatal). Now it won't complain (since can't distinguish between updates and create, if I'm understanding correctly).
That kind of thing shouldn't happen, since is a sign that something quite bad happened (eventcodec work incorrectly; applying events to corrupted state; threadservice duplicating records).

Actually, this was the way that I realized the threadservice had all the concurrency things I found (since it was panicking for that case).

carsonfarmer: Not sure I follow? If a create event comes thru, the event codec would check if its valid (has existing state), if valid, new state is returned and 'put' by store, if not, throw an error. For a save event, if valid (has existing state), return updated state, if not, same deal as above. Then for delete, return nil as the new state (probs don't need to throw if it doesn't exist). Maybe I'm missing something else?

jsign: So both are accessing the store?
I think there might be some confusion regarding some applies and rollback changes through discussions; so maybe I'm the one confused. I mainly focused on this claim:

> it doesn't need to specify what to do with the state, the store just takes it, and applies it. If the state is nil, it deletes it, otherwise, it overwrites it.

If the EventCodec is doing the check, then shouldn't be a problem.

carsonfarmer: Yes, the EventCodec is doing the check. But I think here's the source of confusion for us both: it doesn't need access to the store if each Event is processed in turn. But if we did want to keep them being processed in batch (so the EventCodec could so some clever stuff?), then it seems like it would need access to the batch at least...
So in the end, it is as you say: we're talking about two approaches to the same thing, and until we have something specific that needs changing, its just a matter of opinions I think 😄.

sanderpick: I would prefer to just leave stuff as it is unless it's needed by some indexer, given that is the goal of this work, not refactoring.

carsonfarmer:
Closing this as 'not needed'. We can reopen the discussion/ideas at a later date if they are needed.

PR #221, opened 2020-01-16
Indexing for model fields

carsonfarmer:
Building indexes currently does not affect any other tests, and adds a unique constraint on the ID field by default (so this is partial support for Unique). Any other fields can be specified using arbitrary 'paths' (e.g., `Book.Title`). Currently only JSON mode is supported. Structs _could_ be added later, but this seems good enough for a POC and early tests. This work stores the indexes 'above' any model key paths in the underlying stores, so as to avoid showing up in queries etc. Finally, it requires an EventCodec to call an IndexFunc within the Reduce method as discussed.
Still left to do:
- [x] Use in Queries
- [x] Full support for Unique
- [x] Tests
- [x] Benchmarking


jsonpatcher/jsonpatcher.go
```diff
@@ -134,12 +142,22 @@ func (jp *jsonPatcher) Reduce(events []core.Event, datastore ds.TxnDatastore, ba
 			if err = txn.Put(key, patchedValue); err != nil {
 				return nil, err
 			}
+			if err := indexFunc(e.Model(), key, value, patchedValue, txn); err != nil {
+				return nil, fmt.Errorf("error when indexing created data: %v", err)
+			}
 			actions[i] = core.ReduceAction{Type: core.Save, Model: e.Model(), EntityID: e.EntityID()}
 			log.Debug("\tsave operation applied")
 		case delete:
+			value, err := txn.Get(key)
```
carsonfarmer: Because we use the values to generate the index, to remove the index, we need the original value again...

carsonfarmer: This is a big piece of the indexing overhead.

jsonpatcher/jsonpatcher.go
```diff
@@ -91,7 +91,12 @@ func (jp *jsonPatcher) Create(actions []core.Action) ([]core.Event, format.Node,
 	return events, n, nil
 }
 
-func (jp *jsonPatcher) Reduce(events []core.Event, datastore ds.TxnDatastore, baseKey ds.Key) ([]core.ReduceAction, error) {
+func (jp *jsonPatcher) Reduce(
+	events []core.Event,
+	datastore ds.TxnDatastore,
+	baseKey ds.Key,
+	indexFunc func(model string, key ds.Key, oldData, newData []byte, txn ds.Txn) error,
```
carsonfarmer: As previously discussed, an `indexFun` is used here to index the values within the Event Codec's Reduce function. If we could assume a particular type of data all the time (say, always a document), then we could probably push this indexing step out to the underlying datastore... but since we can't, and want to support alternative Codecs, here is the best place for now...

carsonfarmer: (assuming we want to modify the indexes in the same transaction as the insertions/updates)

store/README.md
```diff
@@ -40,6 +40,24 @@ provide the public API for creating, deleting, updating, and querying instances
 within this model. They also provide read/write transactions which have 
 _serializable isolation_ within the `Store` scope.
 
+*Indexes*
```
carsonfarmer: Some information about indexes here. We _could_ also include a benchmark run, but since this is so variable, I thought I caveat/informational blurb here instead was more appropriate.

store/bench_test.go
```diff
@@ -0,0 +1,235 @@
+package store
```
carsonfarmer: This could use more benchmarks, including perhaps some range queries. But this seems sufficient for now.

store/encode.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019 Tim Shannon. All rights reserved.
```
carsonfarmer: This comes from badherhold, it could be included directly in `index.go` if that is preferable.

store/index.go
```diff
@@ -0,0 +1,324 @@
+// Copyright 2019 Tim Shannon. All rights reserved.
```
carsonfarmer: Left this in here... It actually is barely any code from badgerhold now, but it was inspired by that work, so seemed appropriate to leave here.

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+	iVal, err := DefaultEncode(indexValue)
+	if err != nil {
+		return err
+	}
+
+	return tx.Put(indexKey, iVal)
+}
+
+// keyList is a slice of unique, sorted keys([]byte) such as what an index points to
+type keyList [][]byte
```
carsonfarmer: This is simply an ordered list of keys, used to keep track of entries that match a given index value. The fact that it is ordered is only slightly optimal as used right now, but provides some nice optimizations should we wish to take advantage of this later.

carsonfarmer: The byte encoding used here makes it pretty fast to encode and decode for use when querying. But if there is something faster than gob, we could alternatively use that (see also `store/encode.go`).

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+	return (i < len(*v) && bytes.Equal((*v)[i], b))
+}
+
+// FilterQuery provides a Query Filter function for querying JSON data
+type FilterQuery struct {
+	query *JSONQuery
+}
+
+// Filter is the primary Filter function
+func (f FilterQuery) Filter(e query.Entry) bool {
```
carsonfarmer: Custom Query Filter to do entry filtering within the `ds.Query`. This isn't much of an optimization per se, but it makes the iterator interface easier to use/build.

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+	iter     query.Results
+}
+
+func newIterator(txn ds.Txn, baseKey ds.Key, q *JSONQuery) *iterator {
+	i := &iterator{
+		txn:   txn,
+		query: q,
+	}
+	// Key field or index not specified, pass thru to base 'iterator'
+	if q.Index == "" {
```
carsonfarmer: If we didn't specify an index, we just essentially wrap the default query.Query and return the results that way. That way, not using an index has no real impact on our query from how things were before.

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	// indexed field, get keys from index
+	indexKey := indexPrefix.Child(baseKey).ChildString(q.Index)
+	dsq := query.Query{
+		Prefix: indexKey.String(),
+	}
+	i.iter, i.err = txn.Query(dsq)
+	first := true
+	i.nextKeys = func() ([]ds.Key, error) {
```
carsonfarmer: If we _do_ want to use an index, here's the main piece. On first call to the query, we check if any results are returned, if non, that means we don't have an index for that key.

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+			result, ok := i.iter.NextSync()
+			if !ok {
+				if first {
+					return nil, ErrNoIndexFound
+				}
+				return nKeys, result.Error
+			}
+			first = false
+			// result.Key contains the indexed value, extract here first
+			key := ds.RawKey(result.Key)
```
carsonfarmer: If we do have an index, we extract the value, perform the comparison, and then start returning the keys underlying that index value.

carsonfarmer: We can do this in batches (iteratorKeyMinCacheSize ) if we want, though in practice, this probably isn't needed some of the time given we'll return the results as an array of strings in the end.

store/index.go
```diff
@@ -0,0 +1,324 @@
...
+		return nKeys, nil
+	}
+
+	return i
+}
+
+// NextSync returns the next key value that matches the iterators criteria
+// If there is an error, ok is false and result.Error() will return the error
+func (i *iterator) NextSync() (result query.Result, ok bool) {
+	if i.query.Index == "" {
```
carsonfarmer: Shortcut to just return 'naive' results if we're not using an index.

store/model.go
```diff
@@ -59,10 +62,39 @@ func newModelFromSchema(name string, schema string, s *Store) *Model {
...
+
+// Indexes is a map of model properties to Indexes
+func (m *Model) Indexes() map[string]Index {
+	return m.indexes
+}
+
+// AddIndex creates a new index based on the given path string.
+// Set unique to true if you want a unique constraint on the given path.
+// See https://github.com/tidwall/gjson for documentation on the supported path structure.
+// @note: This does NOT currently build the index. If items have been added prior to adding
```
carsonfarmer: Leaving this out as a 'future optimization'. I could see arguments for building the index if it didn't already exist, but that seems like building something we might not need yet.

store/query_json.go
```diff
@@ -188,31 +193,26 @@ func (c *JSONCriterion) createcriterion(op JSONOperation, value interface{}) *JS
...
 			return nil, fmt.Errorf("error when unmarshaling query result: %v", err)
 		}
-		ok, err = q.matchJSON(val)
-		if err != nil {
-			return nil, fmt.Errorf("error when matching entry with query: %v", err)
-		}
-		if ok {
-			values = append(values, marshaledValue{value: val, rawJSON: res.Value})
-		}
+		values = append(values, marshaledValue{value: val, rawJSON: res.Value})
```
carsonfarmer: _Could_ just return the marshaledValue from the iterator...

store/query_jsonmode_test.go
```diff
@@ -285,6 +285,22 @@ var (
 			query:   JSONWhere("Meta.Rating").Gt(&ratingMid).JSONOrderByDesc("Meta.TotalReads"),
 			ordered: true,
 		},
+		// Indexing
```
carsonfarmer: Just basic tests to make sure it works, do not test the benefits of indexing here.

store/store.go
```diff
@@ -67,6 +67,21 @@ type Store struct {
 	stateChangedNotifee *stateChangedNotifee
 }
 
+func defaultIndexFunc(s *Store) func(model string, key ds.Key, oldData, newData []byte, txn ds.Txn) error {
```
carsonfarmer: The only one for now, but in theory, other indexing functions could be used/implemented.

store/store.go
```diff
@@ -135,9 +150,25 @@ func (s *Store) reregisterSchemas() error {
...
+			Prefix:   indexPrefix.Child(m.BaseKey()).String(),
+			KeysOnly: true,
+		})
+		if err != nil {
+			return err
+		}
+		defer indexes.Close()
+		for index := range indexes.Next() {
+			path := ds.RawKey(index.Key).Name()
+			if err := m.AddIndex(path, false); err != nil { // @todo: Enable unique constraint
```
carsonfarmer: For discussion: We could encode the unique constraint into the index value for easy rehydration, or we could expect callers to specify the index each time, in which case, we wouldn't even really need this hydration step.

store/query_json.go
```diff
@@ -188,31 +193,26 @@ func (c *JSONCriterion) createcriterion(op JSONOperation, value interface{}) *JS
...
+	defer iter.Close()
 
 	var values []marshaledValue
 	for {
-		res, ok := dsr.NextSync()
+		res, ok := iter.NextSync()
 		if !ok {
 			break
 		}
+		// @fixme: This is the second time we unmarshal here, fix that!
```
jsign: So we have two unmarshals per Query item result, since once is done on `newIterator`?

api/client/client.go
```diff
@@ -96,11 +96,19 @@ func (c *Client) NewStore(ctx context.Context) (string, error) {
 }
 
 // RegisterSchema registers a new model shecma
-func (c *Client) RegisterSchema(ctx context.Context, storeID, name, schema string) error {
+func (c *Client) RegisterSchema(ctx context.Context, storeID, name, schema string, indexes []*store.IndexConfig) error {
```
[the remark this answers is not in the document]
sanderpick: Related to this and the use of protos below, it would be nice to consider a higher level object that contains the JSON Schema and indexes. Ideally, this object would allow you so set indexes alongside the type (mongo style), among other (future) things like pre-commit state validators, which we discussed with OpenBazaar.

api/client/client.go
```diff
@@ -96,11 +96,19 @@ func (c *Client) NewStore(ctx context.Context) (string, error) {
...
+	idx := make([]*pb.RegisterSchemaRequest_IndexConfig, len(indexes))
```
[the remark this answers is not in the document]
sanderpick: Agreed, client user shouldn't have to construct pb objects. (Not 100% sure that's the case here... will read on)

api/client/client_test.go
```diff
@@ -231,6 +231,44 @@ func TestModelFind(t *testing.T) {
...
 
+func TestModelFindWithIndex(t *testing.T) {
+	t.Parallel()
+	client, done := setup(t)
+	defer done()
+	t.Run("test model find", func(t *testing.T) {
+		storeID, err := client.NewStore(context.Background())
+		checkErr(t, err)
+		err = client.RegisterSchema(context.Background(), storeID, modelName, schema, []*store.IndexConfig{
+			&store.IndexConfig{
```
sanderpick: 🎉

store/model.go
```diff
@@ -10,6 +10,7 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	ds "github.com/ipfs/go-datastore"
 	core "github.com/textileio/go-threads/core/store"
+	"github.com/tidwall/gjson"
```
sanderpick: 👍

store/model.go
sanderpick: Cool, makes sense to ignore for now... again, mongodb has a nice pattern here: Let the dev run a function like, `ensureIndexes` or similar that applies index additions / changes. Def something for later.

api/client/client.go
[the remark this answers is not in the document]
carsonfarmer: Not sure I follow here? This is the api/client here, depending on the store... which it already depends on?

api/client/client.go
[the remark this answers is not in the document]
carsonfarmer: I like both of these ideas! Should we try to include them here in this PR, or create issues?

store/model.go
```diff
@@ -59,10 +62,39 @@ func newModelFromSchema(name string, schema string, s *Store) *Model {
...
+// Indexes is a map of model properties to Indexes
+func (m *Model) Indexes() map[string]Index {
+	return m.indexes
+}
+
+// AddIndex creates a new index based on the given path string.
+// Set unique to true if you want a unique constraint on the given path.
+// See https://github.com/tidwall/gjson for documentation on the supported path structure.
+// @note: This does NOT currently build the index. If items have been added prior to adding
+// a new index, they will NOT be indexed a posteriori.
```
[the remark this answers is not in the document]
carsonfarmer: Oh yup, good call!

store/query_json.go
carsonfarmer: At the moment, yes. But I think I'm going to push a quick change that actually just returns the unmarshalled object directly from the iterator instead to avoid this silly overhead.

api/client/client.go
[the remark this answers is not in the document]
sanderpick: For now, my vote would be to make them variadic. Once other stuff is needed in the schema configuration, we could move to something that doesn't require the user to supply the JSON schema themselves... at which point pulling the indexes into that object would make sense.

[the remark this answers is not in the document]
carsonfarmer: Perfect. I'll add that then... easy change.

store/query_json.go
carsonfarmer: Fixed

api/client/client.go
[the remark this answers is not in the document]
jsign: Oh yeah, dismiss. I think both confused by seeing the same method name.

store/query_json.go
jsign: Nice!

carsonfarmer:
Ok, I think all nits etc have been addressed. Ready for merge.

api/client/client.go
```diff
@@ -96,7 +96,7 @@ func (c *Client) NewStore(ctx context.Context) (string, error) {
 }
 
 // RegisterSchema registers a new model shecma
-func (c *Client) RegisterSchema(ctx context.Context, storeID, name, schema string, indexes []*store.IndexConfig) error {
+func (c *Client) RegisterSchema(ctx context.Context, storeID, name, schema string, indexes ...*store.IndexConfig) error {
```
sanderpick: 👍

PR #231, opened 2020-01-30
adds android client wrapper around protoc

andrewxhill:
Signed-off-by: andrew <andrew@textile.io>


.github/workflows/test.yml
```diff
@@ -32,3 +32,30 @@ jobs:
...
         run: go test ./...
+
+  android-client:
+    name: Android Client Test
+    runs-on: ubuntu-latest
+    services:
+      threads:
+        image: textile/go-threads
+        ports:
+        - 6006:6006
```
andrewxhill: this is going to cause CI issues when there are breaking API/pb changes in a PR (so this release image will be out of sync)

.github/workflows/test.yml
```diff
@@ -32,3 +32,30 @@ jobs:
...
+        with:
+          version: '3.11.2'
+      - name: Install Android SDK
+        uses: malinskiy/action-android/install-sdk@release/0.0.5
+      - name: Check out code
+        uses: actions/checkout@v1
+      - name: Test
+        run: |
+             cd api/pb/android
+             ./gradlew threads:test
```
andrewxhill: these run the unit tests in https://github.com/textileio/go-threads/pull/231/files#diff-4804adf0dc5fc30a630683638d98519e

api/pb/android/threads/build.gradle
```diff
@@ -0,0 +1,236 @@
...
+        websiteUrl = siteUrl
+        vcsUrl = gitUrl
+        licenses = allLicenses
+        dryRun = false
+        publish = true
+        override = false
+        publicDownloadNumbers = true
+        version {
+            desc = libraryDescription
+        }
```
andrewxhill: no bintray upload step for ci yet, but should be easy to add when ready for any kind of release

api/pb/android/threads/src/main/java/io/textile/threads/Client.java
```diff
@@ -0,0 +1,155 @@
...
+    public ClientException(String message)
+    {
+        super(message);
+    }
+}
+
+/**
+ * Provides top level access to the Textile API
+ */
+public class Client implements LifecycleObserver {
```
andrewxhill: the bulk of the existing api implementations. focused on the easy to do sync methods until all of it is working well

api/pb/android/threads/src/test/java/io/textile/threads/ClientUnitTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.net.URL;
+
+import io.textile.threads_grpc.GetStoreLinkReply;
+import io.textile.threads_grpc.ModelCreateReply;
+
+import static org.junit.Assert.*;
+
+@RunWith(JUnit4.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class ClientUnitTest {
```
andrewxhill: Here are the existing unit tests run in CI now

api/pb/android/threads/src/test/java/io/textile/threads/ClientUnitTest.java
```diff
@@ -0,0 +1,110 @@
...
+        // @todo once working, store modelid to update in later tests
+    }
+
+    @Test
+    public void t06_ModelSave() throws Exception {
+        // @todo use existing modelid
+        JSONObject person = createPerson("", 22);
+        String[] data = { person.toString() };
+        ModelCreateReply reply = client.ModelCreateSync(storeId, "Person", data);
+        assertEquals(1, reply.getEntitiesCount());
```
andrewxhill: this test passes in other clients currently, e.g. dart client

.github/workflows/test.yml
asutula: not following what you're saying here.

asutula: ah it's a docker image?

asutula: why don't we just spin up the go-threads daemon from the current git revision that the CI is operating on?

andrewxhill: could do, i was doing that in js-threads-client for a bit it just felt hacky

api/pb/android/threads/build.gradle
asutula: 👍

.github/workflows/test.yml
asutula: Doesn't sound like a hack to me, sounds like the right way to do how. It's the only way you can guarantee that the server and client are running the same versions.

asutula: or, I should say "compatible" versions.

andrewxhill:
Right yeah, the types bit i just went fast on intentionally, figured I could optimized later.

the json schema example is a good idea, related to types too so you could get add models nicely.

.github/workflows/test.yml
andrewxhill: yeah, i was wondering if, since the dockerfile is here in the branch maybe i could do the same i'm doing from the local definition... but not sure there is a way for that

sanderpick:
I'm sure I missed the planning here, so just thinking out loud. What's the role of the android tests here? You're not really testing anything additional, since the API is already tested and protoc's generation of the Java Client against the proto files are tested (in protoc).

sanderpick:
Oh, I see, there's a higher level java client in there. You don't want to put that in its own repo? Like JS and Dart?

andrewxhill:
My thinking was for 3 parts,

1. when we do any kind of type manipulation it would be ready to go.
2. provides an integration test to verify the library is functioning. 
3. it verifies that the request constructions are done correctly

asutula:
@sanderpick we were chatting about it yesterday, and putting the client in the same repo like this actually streamlines the development process quite a bit. No more needed to publish a package (one for each language) of generated proto code to use as a dependency of a client project in a separate repo. It also makes it very clear if/how each of the clients was updated to respond to some api change... one PR and done.

sanderpick:
Cool, SGTM 👍

.github/workflows/test.yml
andrewxhill: this https://github.community/t5/GitHub-Actions/Docker-action-in-same-repository-with-Dockerfile-in-root/td-p/38757 didn't seem to do the trick for me

PR #235, opened 2020-01-31
shell: add simple developer shell

andrewxhill:
flow

keeping this in wip for a few, would like to tighten up a couple of areas. it provides the minimal set of things for now, ability to 

* `use` a store by id
* once using a store `getModel` by name
* or once using a store, `listen` to all updates

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -59,8 +59,8 @@ type ListenOption struct {
 
 // ListenEvent is used to send data or error values for Listen
 type ListenEvent struct {
-	action Action
-	err    error
+	Action Action
+	Err    error
 }
 
 // Client provides the client api
@@ -313,7 +313,7 @@ func (c *Client) Listen(ctx context.Context, storeID string, listenOptions ...Li
 			if err != nil {
 				stat := status.Convert(err)
 				if stat == nil || (stat.Code() != codes.Canceled) {
-					channel <- ListenEvent{err: err}
+					channel <- ListenEvent{Err: err}
 				}
 				break
 			}
@@ -326,7 +326,7 @@ func (c *Client) Listen(ctx context.Context, storeID string, listenOptions ...Li
 			case pb.ListenReply_SAVE:
 				actionType = ActionSave
 			default:
-				channel <- ListenEvent{err: fmt.Errorf("unknown listen reply action %v", event.GetAction())}
+				channel <- ListenEvent{Err: fmt.Errorf("unknown listen reply action %v", event.GetAction())}
 				break loop
 			}
 			action := Action{
@@ -335,7 +335,7 @@ func (c *Client) Listen(ctx context.Context, storeID string, listenOptions ...Li
 				EntityID: event.GetEntityID(),
 				Entity:   event.GetEntity(),
 			}
-			channel <- ListenEvent{action: action}
+			channel <- ListenEvent{Action: action}
 		}
 	}()
 	return channel, nil
--- a/threads/main.go
+++ b/threads/main.go
@@ -16,34 +16,35 @@ import (
 	"time"
 )
 
-var apiClient *client.Client
-var apiTimeout = time.Second * 5
-var streamTimeout = time.Second * 30
-var currentStore string
+var (
+	apiClient      *client.Client
+	apiTimeout     = time.Second * 5
+	streamTimeout  = time.Second * 30
+	noStoreMessage = "Store required. `use <store id>`"
+	currentStore   string
+	suggestions    = []prompt.Suggest{
+		{
+			Text:        "use",
+			Description: "Switch the store.",
+		},
+		{
+			Text:        "exit",
+			Description: "Exit.",
+		},
+	}
+	storeSuggestions = []prompt.Suggest{
+		{
+			Text:        "listen",
+			Description: "Stream all store updates.",
+		},
+		{
+			Text:        "getModel",
+			Description: "Get all models by model name.",
+		},
+	}
+)
 
 type any map[string]interface{}
-
-var suggestions = []prompt.Suggest{
-	{
-		Text:        "use",
-		Description: "Switch the store.",
-	},
-	{
-		Text:        "exit",
-		Description: "Exit.",
-	},
-}
-
-var storeSuggestions = []prompt.Suggest{
-	{
-		Text:        "listen",
-		Description: "Stream all store updates.",
-	},
-	{
-		Text:        "getModel",
-		Description: "Get all models by model name.",
-	},
-}
 
 func completer(in prompt.Document) []prompt.Suggest {
 	w := in.GetWordBeforeCursor()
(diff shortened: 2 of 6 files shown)


andrewxhill:
@eightysteele this was based on your experience testing and follow-up convo so cc'ing you here. hope this is close to what you imagined being useful to ge to know threads

api/client/client.go
```diff
@@ -59,8 +59,8 @@ type ListenOption struct {
 
 // ListenEvent is used to send data or error values for Listen
 type ListenEvent struct {
-	action Action
-	err    error
+	Action Action
+	Err    error
```
andrewxhill: needed to make these accessible so i could just import the client instead of recreate it

threads/main.go
```diff
@@ -0,0 +1,196 @@
...
+
+func completer(in prompt.Document) []prompt.Suggest {
+	w := in.GetWordBeforeCursor()
+	if w == "" {
+		return []prompt.Suggest{}
+	}
+	if currentStore != "" {
+		return prompt.FilterHasPrefix(append(suggestions, storeSuggestions...), w, true)
+	}
+	return prompt.FilterHasPrefix(suggestions, w, true)
```
andrewxhill: This just gives the use basic prompts when just started or extended prompts once they are using a store

threads/main.go
```diff
@@ -0,0 +1,196 @@
...
+		log.Fatalf("addr error: %v", err)
+	}
+
+	target, err := util.TCPAddrFromMultiAddr(maddr)
+	if err != nil {
+		log.Fatalf("addr error: %v", err)
+	}
+
+	var opts []grpc.DialOption
+	opts = append(opts, grpc.WithInsecure())
```
andrewxhill: should be easy to extend this out later

threads/main.go
```diff
@@ -0,0 +1,196 @@
...
+		prompt.OptionSelectedSuggestionTextColor(prompt.White),
+		prompt.OptionSelectedSuggestionBGColor(prompt.DarkGray),
+
+		prompt.OptionSuggestionTextColor(prompt.LightGray),
+		prompt.OptionSuggestionBGColor(prompt.Black),
+
+		prompt.OptionDescriptionTextColor(prompt.Turquoise),
+		prompt.OptionDescriptionBGColor(prompt.Black),
+		prompt.OptionSelectedDescriptionTextColor(prompt.Turquoise),
+		prompt.OptionSelectedDescriptionBGColor(prompt.Black),
```
andrewxhill: the defaults were a bit hard to look at after a while, so just toned them down here. not tied to these though

eightysteele:
@andrewxhill yesssss, nice man, this is excellent—exactly what i was imagining. just fired it up. super solid user experience (`completer` and `prettyPrint` are nice details). definitely makes threads feel more tangible.

looking at the [service API proto](https://github.com/textileio/go-threads/blob/master/api/pb/api.proto#L182-L197), could be pretty sick to launch the REPL here and do some basic CRUD stuff (e.g., `NewStore`, `RegisterSchema`, `ModelCreate`, `ModelSave`, `ModelDelete`, etc.) while developing an app, etc.

lmk if i can help flush any of that out—so pumped for this. 💪

threads/main.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+		case <-time.After(streamTimeout):
+			fmt.Println("timeout")
+			return
+		}
+	}
+}
+
+func main() {
+	addr := flag.String("addr", "/ip4/0.0.0.0/tcp/6006", "Threads APIs")
```
eightysteele: qq—presumably the `addr` here will eventually be configurable (since you'll be able to connect either locally to the daemon or remotely to cloud.textile.io). is that right?

threads/main.go
```diff
@@ -0,0 +1,196 @@
...
+		Text:        "listen",
+		Description: "Stream all store updates.",
+	},
+	{
+		Text:        "getModel",
+		Description: "Get all models by model name.",
+	},
+}
+
+func completer(in prompt.Document) []prompt.Suggest {
```
eightysteele: 💯

threads/main.go
sanderpick: This will set a default for the flag. So, you could do `threads -addr=/ip4/0.0.0.0/tcp/3000`.

sanderpick: How about "Threads API address" for a description?

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		fmt.Println(err.Error())
+		return
+	}
+	for {
+		select {
+		case val, ok := <-channel:
+			if !ok {
+				return
+			} else {
```
sanderpick: Else not needed here since you're already returning.

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+	if len(entities) == 0 {
+		fmt.Println("None found")
+		return
+	}
+	for _, el := range entities {
+		prettyPrint(el)
+	}
+}
+
+func listen(id string) {
```
sanderpick: How does one stop listening?

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+				}
+				obj := &any{}
+				if err := json.Unmarshal(val.Action.Entity, obj); err != nil {
+					fmt.Println("failed to unmarshal listen result")
+					continue
+				}
+				prettyPrint(obj)
+			}
+		case <-time.After(streamTimeout):
+			fmt.Println("timeout")
```
sanderpick: So, you can only listen for 30 seconds?

andrewxhill: i've got an update coming for this. user just hits enter to cancel

threads/main.go
andrewxhill: got an updating coming for this one, same spot. will make it user controlled.

sanderpick: 💪

sanderpick:
Welcome, @eightysteele 👋

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		storeExecutor(blocks)
+	}
+}
+
+func getModel(id string, model string, maxAwaitTime time.Duration) {
+	ctx, cancel := context.WithTimeout(context.Background(), apiTimeout)
+	defer cancel()
+
+	rawResults, err := apiClient.ModelFind(ctx, id, model, nil, []*any{})
```
sanderpick: `nil` here hurts readability a little, IMO... you could just pass an empty query: `JSONQuery{}`. In fact, for this reason, this method should probably not take a pointer.

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+		prompt.OptionSelectedDescriptionTextColor(prompt.Turquoise),
+		prompt.OptionSelectedDescriptionBGColor(prompt.Black),
+	)
+	p.Run()
+}
+
+func prettyPrint(obj interface{}) {
+	mapB, err := json.MarshalIndent(obj, "", " ")
+	if err != nil {
+		fmt.Println(obj)
```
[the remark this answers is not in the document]
andrewxhill: nice, k

threads/main.go
```diff
@@ -0,0 +1,210 @@
...
+	"os"
+	"strings"
+	"time"
+)
+
+var (
+	apiClient      *client.Client
+	apiTimeout     = time.Second * 5
+	streamTimeout  = time.Second * 30
+	noStoreMessage = "Store required. `use <store id>`"
```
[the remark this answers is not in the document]
andrewxhill: totally, was hoping we could do something like that. the obvious way would be to construct a dynamic suggestions object from the list. we could provide descriptions for each, perhaps store name is the primary key and description contains the store id, that would allow for autocomplete too. or you could leave the description off which would make it a bit tidier is all. 

perhaps then it would be

`use <store name>`

`use -id=<store id>`

ome way to make it select by names as default and ids as a specific method. or other way around.

threads/main.go
```diff
@@ -0,0 +1,210 @@
+package main
+
+import (
```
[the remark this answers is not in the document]
andrewxhill: it does it actually... not sure what makes it _not_ run on some saves

andrewxhill:
okay, thanks for the comments guys. taking this out of draft. pushed a bunch of cleanup. i think it makes a nice mvp from which we can expand quite a lot

PR #238, opened 2020-02-01
ci: adds test build script for linux releases

andrewxhill:
gtg. before merge I'll,

- [x] retarget the workflow at only releases
- [x] enable the final step of each of the jobs to try and publish the outputs to the releases


.github/workflows/release.yml
```diff
@@ -0,0 +1,159 @@
...
+        run: |
+              export PATH=${PATH}:`go env GOPATH`/bin
+              SUMMARY=${GITHUB_SHA}
+              if [[ $GITHUB_REF == refs/tags/* ]]; then
+                SUMMARY=${GITHUB_REF/refs\/tags\//}
+              fi  
+              DATE=$(date --iso-8601=seconds)
+              FLAGS="-X github.com/textileio/go-threads/common.GitSummary=${SUMMARY} -X github.com/textileio/go-textile/common.BuildDate=${DATE} -X github.com/textileio/go-textile/common.GitCommit=${GITHUB_SHA} -X github.com/textileio/go-textile/common.GitBranch=master -X github.com/textileio/go-textile/common.GitState=clean"
+              gox -ldflags="-w $FLAGS" -osarch="linux/amd64 linux/386 linux/arm" -output="threadsd-{{.OS}}-{{.Arch}}" ./threadsd
+              gox -ldflags="-w $FLAGS" -osarch="linux/amd64 linux/386 linux/arm" -output="threads-{{.OS}}-{{.Arch}}" ./threads
```
sanderpick: You can remove these `ldflags` since they won't be picked up internally.

.github/workflows/release.yml
```diff
@@ -0,0 +1,159 @@
...
+              echo $(ls ${OUT}/)
+      # - name: Upload multiple assets to release
+      #   uses: radiusnetworks/release-asset-action@v1
+      #   with:
+      #     files: |
+      #       dist/cli/go-threads_${GITHUB_REF##*/}_linux-amd64.tar.gz
+      #       dist/cli/go-threads_${GITHUB_REF##*/}_linux-386.tar.gz
+      #       dist/cli/go-threads_${GITHUB_REF##*/}_linux-arm.tar.gz
+      #     github-token: ${{ secrets.GITHUB_TOKEN }}
+  release-win-mac:
```
sanderpick: Since we don't need `CGO` enabled, you don't need to supply a C toolchain. In other words, you should be able to just build the mac and windows versions on the linux machine above with `gox`.

andrewxhill: okay, fixed it up. looks like these machines are CPU=1 right now

dist/install
```diff
@@ -0,0 +1,48 @@
...
+#
+# Installation script for threads. It tries to move $bin in one of the
+# directories stored in $binpaths.
+
+INSTALL_DIR="$(dirname "$0")"
+
+threadsbin="$INSTALL_DIR/threads"
+threadsdbin="$INSTALL_DIR/threadsd"
+binpaths="/usr/local/bin /usr/bin"
+
```
andrewxhill: I guess this isn't really going to cut it for windows as an install script. Should I just remove it from that archive?

andrewxhill: okay, omitting 'install' from the windows archive

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Get dependencies
+        if: steps.cache-dependencies.outputs.cache-hit != 'true'
+        run: |
+          export PATH=${PATH}:`go env GOPATH`/bin
+          go get -v -t -d ./...
+      - name: Install gox
+        run: |
+             export PATH=${PATH}:`go env GOPATH`/bin
```
sanderpick: It might not like the extra indents...

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+             go get github.com/mitchellh/gox  
+      - name: Compile
+        run: |
+              export PATH=${PATH}:`go env GOPATH`/bin
```
sanderpick: Same here...

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+                  mv go-threads_${VERSION}_${OSARCH}.zip ${OUT}/
+              else
+                  tar -czvf go-threads_${VERSION}_${OSARCH}.tar.gz LICENSE README.md install threadsd threads
+                  mv go-threads_${VERSION}_${OSARCH}.tar.gz ${OUT}/
+              fi
+          done
+          cd .. && rm -rf tmp
+      - name: Mock Publish
+        run: |
+              OUT=~/dist/cli
```
sanderpick: And here...

PR #240, opened 2020-02-04
Service: Handle error when starting gRPC server

burdiyan:
Currently error returned by Start() method of the gRPC server is not handler when instantiating a thread service.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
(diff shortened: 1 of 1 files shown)


carsonfarmer:
Oops, looks like some test failures there. Will have to double check the cause there before merging.

burdiyan:
@carsonfarmer Yep, sorry for that. Should be fixed now.

carsonfarmer:
For future reference, we require all commits to be signed (in addition to the DCO). You can add this pretty easily by default (which is a good idea in general): https://help.github.com/en/github/authenticating-to-github/signing-commits

PR #242, opened 2020-02-04
Docker: remove flags, enable env vars

sanderpick:
There's some nits included in here as well that I've added locally, just hadn't needed a PR until now.

### DIFF
--- a/service/client.go
+++ b/service/client.go
@@ -97,7 +97,7 @@ func (s *server) pushLog(ctx context.Context, id thread.ID, lg thread.LogInfo, p
 	client := pb.NewServiceClient(conn)
 	_, err = client.PushLog(cctx, lreq)
 	if err != nil {
-		log.Warnf("push log to %s failed: %s", pid.String(), err)
+		return fmt.Errorf("push log to %s failed: %s", pid.String(), err)
 	}
 	return err
 }
--- a/threadsd/main.go
+++ b/threadsd/main.go
@@ -2,11 +2,12 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
+	"os"
 
 	logging "github.com/ipfs/go-log"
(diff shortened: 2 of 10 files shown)


threadsd/main.go
```diff
@@ -2,11 +2,12 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
+	"os"
 
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
+	"github.com/namsral/flag"
```
sanderpick: This is a drop in replacement for `flag` that looks to env vars after command line args.

service/client.go
```diff
@@ -97,7 +97,7 @@ func (s *server) pushLog(ctx context.Context, id thread.ID, lg thread.LogInfo, p
 	client := pb.NewServiceClient(conn)
 	_, err = client.PushLog(cctx, lreq)
 	if err != nil {
-		log.Warnf("push log to %s failed: %s", pid.String(), err)
+		return fmt.Errorf("push log to %s failed: %s", pid.String(), err)
```
sanderpick: This might fail the existing `AddFollower` tests... we'll see. I was seeing cases where the log was successfully pushed to the new follower, but the addresses were updated anyway, meaning that get records would probably fail forever. May we well error up front.

carsonfarmer: Looks like all tests are still passing. Safe to move forward with this then?

sanderpick: Nice 👍

threadsd/main.go
```diff
@@ -53,6 +58,14 @@ func main() {
 		}
 	}
 
+	log.Debugf("repo: %v", *repo)
```
carsonfarmer: YESSSS!

service/client.go
sanderpick: (updated comment above which was confusing)

PR #243, opened 2020-02-07
Deps: Update cid package

sanderpick:
Needed for https://github.com/textileio/textile/pull/74


sanderpick:
Should be good now with https://github.com/textileio/go-threads/pull/243/commits/5b3d35fbdfe74ab87a19989b7bd082f2ad497f5a. Some sentinel error in the `cid` package was replaced with something similar in a `varint` package that we needed to catch... 🤷‍♂

sanderpick:
I'm not sure what's up with DCO not finishing but I'll merge anyway.

PR #256, opened 2020-02-27
Android textile

andrewxhill:
Adds a temporary method to the android-threads lib that will allow the use of project tokens for remote threads. Ideally this will move to it's own auth lib after we've tested it with the first user.


api/pb/android/app/src/main/java/io/textile/textileexample/MainActivity.java
```diff
@@ -5,34 +5,57 @@
...
-            client.Connect();
+            /**
+             * To use hosted threads from Textile, use the following instead of DefaultConfig
+             */
+            /*
+                Config config = new TextileConfig(
+                    PROJECT_TOKEN,
+                    DEVICE_UUID
+                );
+                // Optionally, restore a session for the same user by supplying the session ID
```
eightysteele: maybe include a note here about where to find the `session ID` (e.g., after a `textile login` you'll find it in `~/.textile/auth.yml`)

andrewxhill: no, that's for developer accounts. for app accounts, they just need the project token `textile project add token` and then session id is generated as a result of running the above without a provided session id a first time

api/pb/android/app/src/main/java/io/textile/textileexample/MainActivity.java
```diff
@@ -5,34 +5,57 @@
...
+                config.setSession(SESSION_ID);
+                ...
+                // If it's your first time starting the client, you can get and locally store the session id
+                String SESSION_ID = client.getSession();
+                // The above only works when using the TextileConfig
+            */
+
+            Config config = new DefaultConfig();
+            client = new Client(config);
+            client.init((success)->{
```
eightysteele: quick sanity check—curious if we need to call `client.start()` after a successful `init()` here.

andrewxhill: i think you mean starting a store, so you'd do like,

```java
String storeId = client.NewStoreSync();
client.StartSync(storeId);
```

api/pb/android/threads/src/main/java/io/textile/threads/Client.java
```diff
@@ -83,13 +102,29 @@ public void StartFromAddressSync (String storeID, String address, ByteString fol
...
+
 
     public GetStoreLinkReply GetStoreLinkSync (String storeID) {
         GetStoreLinkRequest.Builder request = GetStoreLinkRequest.newBuilder();
         request.setStoreID(storeID);
         return blockingStub.getStoreLink(request.build());
     }
 
+
+    public void GetStoreLink (String storeID, StreamObserver<GetStoreLinkReply> responseObserver) {
```
eightysteele: maybe add a comment that this method won't work until you do a `client.start()`.

api/pb/android/threads/src/main/java/io/textile/threads/Session.java
```diff
@@ -0,0 +1,6 @@
+package io.textile.threads;
+
+class Session {
```
eightysteele: what's the `id` here? presumably it's not the session id since that is probably stored in `session_id`?

api/pb/android/app/src/main/java/io/textile/textileexample/MainActivity.java
eightysteele: noice!

api/pb/android/app/src/main/java/io/textile/textileexample/MainActivity.java
eightysteele: yup yup 👍

api/pb/android/threads/src/main/java/io/textile/threads/Client.java
```diff
@@ -1,63 +1,72 @@
...
+     * @return the current session id or null
+     */
+    public String getSession() {
+        return this.config.getSession();
+    }
+    /**
+     * Method must be called before using the Client and while the device has an internet connection.
+     * @param ready
+     */
+    public void init(Consumer<Boolean> ready) {
```
asutula: Doing some reading about `Consumer`s, it's not clear to me that is the correct API here. Seems like it should be returning a `Promise` since there is some async work done in here.

api/pb/android/threads/src/main/java/io/textile/threads/TextileConfig.java
```diff
@@ -0,0 +1,114 @@
...
+        RequestBody body = RequestBody.create(JSON, jsonString);
+        Request request = new Request.Builder()
+                .url(cloud)
+                .post(body)
+                .build();
+        client.newCall(request).enqueue(new Callback() {
+            @Override
+            public void onFailure(Request request, IOException throwable) {
+                throwable.printStackTrace();
+                System.out.println("Info: err " + throwable.getMessage());
```
asutula: Yea, this is a bit of the problem with trying to use `Consumer`... There's no way to return the error from this async http back to the caller.

api/pb/android/threads/src/main/java/io/textile/threads/Session.java
asutula: Good question, but I also don't see that this class is used anywhere. Could remove if that's true.

asutula: Oh, I see it's used in `TextileConfig.refreshSession`.

andrewxhill: It's only an internal class

api/pb/android/threads/src/main/java/io/textile/threads/Client.java
andrewxhill: what do you mean by promise exactly?

asutula: Sorry, meant `Future` https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html

api/pb/android/threads/src/main/java/io/textile/threads/TextileConfig.java
andrewxhill: okay, updated to use a Future

andrewxhill:
should be gtg

api/pb/android/threads/src/main/java/io/textile/threads/Client.java
andrewxhill: holding off until fixed in threads to not need it (coming up)

PR #260, opened 2020-03-10
Auto-start

### LINKED ISSUE #230: Auto-start stores at creation time
Curious, wouldn’t it be nicer if a thread store just started itself after creation? I know we’ve stretched the capabilities of a store beyond it’s original use-case (multi-store, multi-tenant, etc.), but given it’s current trajectory, ideally a remote user wouldn’t have to “start” a store after registering models. They would just register them whenever needed without thinking much about it.

So, we’d have NewStore and NewStoreFromAddr instead of NewStore, Start, and StartFromAddr.

This would also make https://github.com/textileio/go-threads/issues/115 go away.

### LINKED ISSUE #264: Migrate manager keys to thread IDs
There does not seem to be a usecase where a DB has more than one thread. Therefore, to reduce the burden on consumers, the manager's DBs can be accessed with the thread ID. This saves having to maintain another piece of state.

### LINKED ISSUE #267: Consolidate gRPC servers
There should only be one listener for the net and db services.

sanderpick:
Closes #267 
Closes #230 
Closes #264 

- Auto-starting stores via `NewDB` and `NewDBFromAddr`
- Manager IDs are now `thread.ID`s
- Consolidates gRPC services under a single server, which can be also be used in the same way under a single server in `textile`


service/service.go
```diff
@@ -242,12 +226,22 @@ func (t *service) AddThread(
 	}); err != nil {
 		return
 	}
+	if args.ReadKey != nil {
```
sanderpick: As mentioned in slack, we can add a log here (like `CreateThread`) if there's a read key. Not a perfect heuristic, but works.

service/service.go
```diff
@@ -260,19 +254,12 @@ func (t *service) AddThread(
...
 	}
 
-	// @todo: ensure not overwrite with newer info from owner?
 	for _, l := range lgs {
 		if err = t.createExternalLogIfNotExist(id, l.ID, l.PubKey, l.PrivKey, l.Addrs); err != nil {
 			return
 		}
 	}
 
-	go func() {
```
sanderpick: Not auto-pulling. This was making life hard for the DB.

threadsd/main.go
```diff
@@ -76,26 +72,77 @@ func main() {
...
+	}
+
+	server := grpc.NewServer()
+	listener, err := net.Listen("tcp", target)
+	if err != nil {
+		log.Fatal(err)
+	}
+	go func() {
+		pb.RegisterAPIServer(server, service)
+		netpb.RegisterAPIServer(server, netService)
```
sanderpick: Same server.

sanderpick:
TODO: The Android client needs updating.

asutula:
👀

api/pb/api.proto
```diff
@@ -6,14 +6,30 @@ option java_package = "io.textile.threads_grpc";
...
+
+message GetDBInfoReply {
+    repeated string addresses = 1;
+    bytes followKey = 2;
+    bytes readKey = 3;
 }
 
 message NewCollectionRequest {
-    string DBID = 1;
+    string dbID = 1;
```
asutula: cool thanks for changing all these.

cc @andrewxhill this should make the generated js/ts code better

api/client/client.go
```diff
@@ -177,40 +176,20 @@ func (c *Client) Save(ctx context.Context, dbID, collectionName string, items ..
...
-	res := make([]string, len(addrs))
-	for i := range addrs {
-		addr := addrs[i]
-		fKey := base58.Encode(resp.GetFollowKey())
-		rKey := base58.Encode(resp.GetReadKey())
-		res[i] = addr + "?" + fKey + "&" + rKey
-	}
-	return res, nil
-}
-
```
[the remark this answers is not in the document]
asutula: Ok, `getDbInfo` it seems. Nice!

api/service.go
```diff
@@ -6,182 +6,208 @@ import (
...
-func (s *service) NewDB(context.Context, *pb.NewDBRequest) (*pb.NewDBReply, error) {
-	log.Debugf("received new db request")
+// Config specifies server settings.
+type Config struct {
+	RepoPath string
+	Debug    bool
+}
+
+// NewService starts and returns a new service with the given threadservice.
+// The threadservice is *not* managed by the server.
```
asutula: After this is merged, we should make the "service" (threadservice) > "network" refactor. Lots of ambiguous working in this comment and the method signature just below.

api/service.go
```diff
@@ -354,12 +380,12 @@ func (s *service) Listen(req *pb.ListenRequest, server pb.API_ListenServer) erro
...
 			listenActionType = db.ListenSave
 		default:
-			return status.Errorf(codes.InvalidArgument, "invalid filter action %v", filter.GetAction())
+			return status.Errorf(codes.InvalidArgument, "invalid filter action %v", filter.Action)
 		}
 		options[i] = db.ListenOption{
 			Type:       listenActionType,
-			Collection: filter.GetCollectionName(),
-			ID:         core.InstanceID(filter.InstanceID),
+			Collection: filter.CollectionName,
```
asutula: What's the difference between accessing properties in the generated code using the "get" method vs the property directly? Never been sure about that.

api/service.go
```diff
@@ -499,20 +525,21 @@ func (s *service) processFindRequest(req *pb.FindRequest, findFunc func(q *db.JS
 	return &pb.FindReply{Instances: byteInstances}, nil
 }
 
-func (s *service) getDB(idStr string) (*db.DB, error) {
-	id, err := uuid.Parse(idStr)
+func (s *Service) getDB(idStr string) (d *db.DB, id thread.ID, err error) {
```
asutula: Nice, been thinking I need to start using named return values so the code is more self documenting.

asutula: Although, if self documenting code is what we're going for, I guess names better than `d` and `id` would be in order.

db/db.go
```diff
@@ -68,19 +64,58 @@ type DB struct {
...
+			return nil, err
+		}
+	}
+	return newDB(ts, id, config)
+}
+
+// NewDBFromAddr creates a new DB from a thread hosted by another peer at address,
+// which will *own* ds and dispatcher for internal use.
+// Saying it differently, ds and dispatcher shouldn't be used externally.
+func NewDBFromAddr(ctx context.Context, ts s.Service, addr ma.Multiaddr, followKey, readKey *sym.Key, opts ...Option) (*DB, error) {
```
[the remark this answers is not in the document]
asutula: Related to the question I asked in the client layer (about no thread id being passed in here), I guess it's because the thread id was already created on another peer?

db/db.go
```diff
@@ -68,19 +64,58 @@ type DB struct {
...
+	if len(ti.Logs) == 0 {
+		ti, err = ts.CreateThread(ctx, id, s.FollowKey(sym.New()), s.ReadKey(sym.New()))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return newDB(ts, id, config)
+}
+
+// NewDBFromAddr creates a new DB from a thread hosted by another peer at address,
```
asutula: This seems really nice and clean

examples/local_eventstore/books/main.go
```diff
@@ -21,10 +24,10 @@ type bookStats struct {
 }
 
 func main() {
-	s, clean := createMemDB()
+	d, clean := createMemDB()
```
asutula: I tried to find all these and rename them, but obv missed some. Thanks for cleaning up!

service/api/service.go
```diff
@@ -6,35 +6,61 @@ import (
...
+	ts core.Service
+}
+
+// Config specifies server settings.
+type Config struct {
+	Debug bool
+}
+
+// NewService starts and returns a new service.
+func NewService(ts core.Service, conf Config) (*Service, error) {
```
asutula: In the case of the threads api server and this service server, it seems like you removed the "sever" and just added a `NewService` method... I'm guessing that is just because the "server" wasn't really providing any functionality other than creating the service?

service/service.go
```diff
@@ -72,15 +72,7 @@ type Config struct {
...
-func NewService(
-	ctx context.Context,
-	h host.Host,
-	bstore bs.Blockstore,
-	ds format.DAGService,
-	ls lstore.Logstore,
-	conf Config,
-	opts ...grpc.ServerOption,
-) (core.Service, error) {
+func NewService(ctx context.Context, h host.Host, bstore bs.Blockstore, ds format.DAGService, ls lstore.Logstore, conf Config, opts ...grpc.ServerOption) (core.Service, error) {
```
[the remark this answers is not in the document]
asutula: Yea wish this was a thing where the go tooling just formatted it a certain way always. 🤷‍♂

threadsd/main.go
asutula: beautiful

db/db.go
jsign: I see that in `newDb` the _DB_ is started.
Shouldn't this method also receive the Models to register before the start, to avoid the problem of receiving patches without having the models to validate? (If that's the case (because i'm maybe missing something), would be nice to add that to the comments?)

db/dbthread.go
```diff
@@ -67,10 +65,12 @@ func (a *singleThreadAdapter) Start() {
 	a.started = true
 	li, err := a.api.GetThread(context.Background(), a.threadID)
 	if err != nil {
-		log.Fatalf("error when getting/creating own log for thread %s: %v", a.threadID, err)
+		log.Fatalf("error geting thread %s: %v", a.threadID, err)
```
jsign: _s/geting/getting_

db/manager.go
```diff
@@ -42,93 +47,119 @@ func NewManager(ts service.Service, opts ...Option) (*Manager, error) {
...
 		m.dbs[id] = s
 	}
-
 	return m, nil
 }
 
 // NewDB creates a new db and prefix its datastore with base key.
-func (m *Manager) NewDB() (id uuid.UUID, db *DB, err error) {
-	id, err = uuid.NewRandom()
+func (m *Manager) NewDB(ctx context.Context, id thread.ID) (*DB, error) {
```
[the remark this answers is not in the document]
jsign: Coo. Is calling this method multiple times with the same ID safe? (quick question thinking if internal `AddThread` etc are safe enough)

db/db.go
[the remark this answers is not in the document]
sanderpick: Yep. The thread ID is part of the address, which was created by a different peer.

service/service.go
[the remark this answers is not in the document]
sanderpick: Totally.

api/service.go
sanderpick: Totally. I can smash that in after this quick.

examples/local_eventstore/books/main.go
sanderpick: Yep yep!

service/api/service.go
sanderpick: Yep, exactly. After removing the gRPC server, it felt like an unnecessary layer.

api/service.go
sanderpick: I think they try to return the zero-value of the property _other than nil_. So, in the case of strings, the getter returns `""` instead of nil. But, in the case of types who's zero value _is nil_, they just return `nil`, e.g.,

```
func (m *NewDBFromAddrRequest) GetReadKey() []byte {
	if m != nil {
		return m.ReadKey
	}
	return nil
}
```

I'm not sure this is actually useful though, since we often have cases where nil is a valid value. So, we end up checking anyway. Honestly though, I was just going for consistency.

api/service.go
sanderpick: I've been liking the named returns when one of them is not a pointer receiver, i.e., `string` or `thread.ID`. This way, you don't have a bunch of `return "", err` lines in the method, since the value is already initialized. Otherwise, I like the normal style, which can be more explicit, and therefore more readable.

sanderpick: That said, I agree about the short names... seems to be a Go custom that doesn't have a good explanation.

sanderpick:
@asutula regarding the threads "shell", it will get replaced with the JS version that @eightysteele wrote: https://github.com/eightysteele/thread-shell

crypto/symmetric/symmetric.go
```diff
@@ -12,17 +12,26 @@ type Key struct {
...
+		panic(err)
+	}
+	return k
+}
+
+// FromBytes returns a key by wrapping k.
+func FromBytes(k []byte) (*Key, error) {
 	if len(k) != 44 {
 		return nil, fmt.Errorf("invalid key")
 	}
```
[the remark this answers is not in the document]
sanderpick: As in, not needed?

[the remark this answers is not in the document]
jsign: Always return a nil key. Am I I missing something?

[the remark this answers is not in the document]
jsign: Oh. Collapsed code. Forget!

db/collection.go
```diff
@@ -358,11 +358,15 @@ func (t *Txn) Commit() error {
...
+		if err := t.collection.db.dispatcher.Dispatch(events); err != nil {
+			return err
+		}
 	}
-	if err := t.collection.db.notifyTxnEvents(node); err != nil {
-		return err
+	if node != nil {
+		if err := t.collection.db.notifyTxnEvents(node); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
sanderpick: Good points. I should have made a note about this change. I added these conditionals because there were some panics on the tests... the `mockEventCodec` returns a `nil` node:

```
func (dec *mockEventCodec) Create([]core.Action) ([]core.Event, format.Node, error) {
	dec.called = true
	return nil, nil, nil
}
```

This was causing a panic downstream on the `localEventBus`: 

```
func (leb *localEventsBus) Listen() *LocalEventListener {
	l := &LocalEventListener{
		listener: leb.bus.Listen(),
		c:        make(chan format.Node),
	}

	go func() {
		for v := range l.listener.Channel() {
			events := v.(format.Node)
			l.c <- events
		}
		close(l.c)
	}()

	return l
}
```

Specifically, `events := v.(format.Node)` is not safe when `v` is nil. 

In any case, I can make your change... having both events and node empty is still not valid.

db/db.go
```diff
@@ -68,19 +64,58 @@ type DB struct {
...
+	ti, err := ts.GetThread(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if len(ti.Logs) == 0 {
+		ti, err = ts.CreateThread(ctx, id, s.FollowKey(sym.New()), s.ReadKey(sym.New()))
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
sanderpick: Well, `GetThread` does not error in the case where we have no info about a thread. That method pulls info from all the log books. We could certainly make it error if all the keys _and_ logs are empty.

db/db.go
sanderpick: Sure, we could allow collections to be created here. I didn't do that originally because as long as we allow them to be added _at any point_, including after the db is started, the safety measure of adding them when the db is started is sorta mute, right?

api/service.go
asutula: Cool, yea I'd like to be consistent about using one way or another. Using the property sounds good to me.

db/manager.go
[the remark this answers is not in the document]
sanderpick: It is safe at the manager level... I'll take a look at the internal methods. I don't think they are safe enough.

db/db.go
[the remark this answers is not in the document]
jsign: Yeah, my point was that the: _get X, don't have an error, then create X_ feels weird without knowing the implementation details. Mostly, from a client-side perspective. 
Feels like the `CreateThread` method call should be called `CreateThreadLog` or similar. 

Not sure we should do something about it now, but sharing this since I haven't used the APIs for some time and this doubt bubbled up.

db/db.go
jsign: Say I want to join an existing _Db_ in the safest way possible.
What would be the correct flow?

Asking that because if the first line of code is `NewDBFromAddr`, then that isn't safe since it will start syncing existing patches possibly without giving me time to register the necessary models (even if the next line in the code is registering the models). Am I missing some other existing _Db_ bootstrap code flow?

sanderpick: Nope, you're correct... but the solution is a bit more involved. We need a way to load a the models from the address. Otherwise, we're back in the same position as before: I have to manually start the DB.

jsign: Well, the long-term solution I completely agree should be some ideas we brainstormed some days ago about models being first-class citizens.

But, if you can receive a slice of Models as a param in this method which the method registers before calling the `Start`, should suffice to avoid this problem and still auto-starts the _DB_.

Just asking if it's a timing decision since the full solution will be short-term, or maybe this idea was missed.

jsign: To be clear, just mentioning what could be a safer temporal solution to that problem. That doesn't change the fact that a less safe one might be best to avoid changing the API again later.

sanderpick: I see what you mean! Okay, I'll add the register models to this method. Do you agree the same pattern is not needed in `NewDB`?

jsign: Yep, `NewDB` doesn't need that since nothing should happen between that `NewDB` and the next line that some models are registered!

sanderpick: Okay, sounds good. Agreed. Change coming after I finish updating the Android client.

PR #261, opened 2020-03-10
Model > Collection, Entity > Instance

### LINKED ISSUE #259: RegisterSchema > NewCollection
Per retreat decisions

asutula:
This completes the database layer naming refactor so that is works like this:

`DB`s contain `Collection`s which manage `Instance`s which have `InstaceID`s. No more "model" or "entity" wording left anywhere.

I also dropped the "model" prefix from the CRUD method names in the `Collection` API and all APIs derived from it (so it's simply `Create`, `Save`, `Delete`, etc).

closes #259


asutula:
cc @carsonfarmer just so you know about the latest naming conventions, see the PR description.

PR #271, opened 2020-03-17
Error out when no local thread info

carsonfarmer:
Minor nit discovered when working with the gRPC APIs. Calling getThread when the given Thread does not exist on the remote peer leads to a nil pointer panic. This checks for and returns an error when the FollowKey is not available.

### DIFF
--- a/net/api/service.go
+++ b/net/api/service.go
@@ -112,6 +112,9 @@ func (s *Service) GetThread(ctx context.Context, req *pb.GetThreadRequest) (*pb.
 	if err != nil {
 		return nil, err
 	}
+	if info.ReadKey == nil && info.FollowKey == nil {
+		return nil, fmt.Errorf("empty thread info")
+	}
 	return threadInfoToProto(info)
 }
 


net/api/service.go
```diff
@@ -112,6 +112,9 @@ func (s *Service) GetThread(ctx context.Context, req *pb.GetThreadRequest) (*pb.
 	if err != nil {
 		return nil, err
 	}
+	if info.ReadKey == nil && info.FollowKey == nil {
```
carsonfarmer: Is this the best place to perform this check? Or should we do this in `threadInfoToProto`? In theory, one shouldn't call this at all if the thread doesn't exist locally, but there is a non-zero likelihood that this will get called for a thread that doesn't exist.

sanderpick: `GetThread` should now return `ErrThreadNotFound` if it's not available. Maybe you saw this before the last big PR went in?

carsonfarmer: Ah yes indeed!

PR #272, opened 2020-03-18
Follow key -> Service key and a single public API Thread key

### LINKED ISSUE #104: Combine follow+read key string representation
Users should not have to pass around two keys when dealing with a store/threads. We can do something like `base58(follow-key):base58(read-key)`.

### LINKED ISSUE #151: Follower -> Service
Pretty minimal code changes in this repo:
https://github.com/textileio/go-textile-threads/search?q=Follower&unscoped_q=Follower

sanderpick:
Closes #151 
Closes #104 

As discussed, this changes _follow_ key to _service_ key and `AddFollower` to `AddReplicator`.

Additionally, this introduces a new concept: _Thread Key_, which is a smart wrapper around a service and read key, allowing for easy marshaling and string conversion. There are two variants, _key_ and _service-only key_. 

Part of the motivation for this change is to simplify the public handling of thread keys, i.e., in URLs, invites, etc. For this reason, I did not bubble these changes down in to the log store layer, which does not concern itself with this higher level `thread.Key`.

Very much like `thread.ID`, `thread.Key` can be represented as a base32-encoded string:

#### Thread Key
```
brv7t5l2h55uklz5qwpntcat26csaasfchzof3emmdy6povabcd3a2to2qdkqdkto2prfhizerqqudqsdvwherbiy4nazqxjejgdr4oy
```

#### Thread Service-Key
```
bp2vvqody5zm6yqycsnazb4kpqvycbdosos352zvpsorxce5koh7q
```

The full thread key string encodes both the service and read key bytes, while the service key string only encodes the service key bytes. 

Creating a thread without giving a `thread.Key` option will result in the underlying net layer creating and storing a full thread key (with service and read components). Previously, not supplying key options (the old follow and read options), resulted in a follow-only thread. Looking back, I can't remember why I did this... doesn't feel very intuitive given that the keys are _optional_. So, tl;dr, if you want to create a thread where the net layer doesn't own the read key, use a `thread.Key` that only has a service component when calling `CreateThread`.

Lastly, this removes the silly nonce handling in `symmetric.Key`. Nonces are now created during encryption and prepended to ciphertext.


core/net/net.go
```diff
@@ -40,13 +40,16 @@ type API interface {
 	GetThread(ctx context.Context, id thread.ID) (thread.Info, error)
 
 	// PullThread for new records.
+	// Logs owned by this host are traversed locally.
+	// Remotely addressed logs are pulled from the network.
+	// Is thread-safe.
```
sanderpick: Nit: Moving all doc strings to the interface.

core/thread/key.go
```diff
@@ -0,0 +1,122 @@
+package thread
```
sanderpick: Here's the new `thread.Key`.

crypto/asymmetric/asymmetric.go
```diff
@@ -89,9 +89,9 @@ func decrypt(ciphertext []byte, sk ic.PrivKey) ([]byte, error) {
 	return nil, fmt.Errorf("could not determine key type")
 }
 
-func publicToCurve25519(k *ic.Ed25519PublicKey) (*[32]byte, error) {
-	var cp [32]byte
-	var pk [32]byte
+func publicToCurve25519(k *ic.Ed25519PublicKey) (*[EphemeralPublicKeyBytes]byte, error) {
+	var cp [EphemeralPublicKeyBytes]byte
```
sanderpick: Just making use of the constants that were already here.

crypto/symmetric/symmetric.go
```diff
@@ -30,52 +40,76 @@ func New() *Key {
...
 	if err != nil {
 		return nil, err
 	}
 	aesgcm, err := cipher.NewGCM(block)
 	if err != nil {
 		return nil, err
 	}
-	ciph := aesgcm.Seal(nil, k.raw[32:], plaintext, nil)
-	return ciph, nil
+	nonce := make([]byte, NonceBytes)
```
sanderpick: Like we discussed a long time ago, there's no reason to keep the nonce _in the key_. Now, the nonce is created for each ciphertext, and then prepended to it.

db/db.go
```diff
@@ -75,7 +74,7 @@ func NewDB(ctx context.Context, network net.Net, id thread.ID, opts ...Option) (
 
 	if _, err := network.GetThread(ctx, id); err != nil {
 		if errors.Is(err, lstore.ErrThreadNotFound) {
-			if _, err = network.CreateThread(ctx, id, net.FollowKey(sym.New()), net.ReadKey(sym.New())); err != nil {
+			if _, err = network.CreateThread(ctx, id); err != nil {
```
sanderpick: As mentioned in the description, you no longer _need_ to supply your own keys.

sanderpick: (You would if you want to keep the read key local)

sanderpick: (Or log private key)

net/api/client/client_test.go
```diff
@@ -211,15 +199,12 @@ func TestClient_AddRecord(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
+	tk := thread.NewRandomServiceKey()
 	sk, pk, err := crypto.GenerateEd25519Key(crand.Reader)
 	if err != nil {
 		t.Fatal(err)
 	}
-	info, err := client.CreateThread(context.Background(), id, core.FollowKey(fk), core.LogKey(pk))
+	info, err := client.CreateThread(context.Background(), id, core.ThreadKey(tk), core.LogKey(pk))
```
sanderpick: Here's an example of supplying a locally-created service-only key, and a locally-created private key.

net/api/pb/javascript/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@textile/thread-net-grpc",
+  "name": "@textile/threads-net-grpc",
```
sanderpick: Changing this @carsonfarmer.

api/client/client_test.go
```diff
@@ -54,11 +53,8 @@ func TestNewDBFromAddr(t *testing.T) {
 	t.Run("test new db from address", func(t *testing.T) {
 		addr, err := ma.NewMultiaddr(info.Addresses[0])
 		checkErr(t, err)
-		fk, err := sym.FromBytes(info.FollowKey)
-		checkErr(t, err)
-		rk, err := sym.FromBytes(info.FollowKey)
```
jsign: *Seems* that this also deleted a bug :P (`rk := ... info.FollowKey)`?)

core/thread/key.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"fmt"
+
+	mbase "github.com/multiformats/go-multibase"
+	sym "github.com/textileio/go-threads/crypto/symmetric"
+)
+
+var (
+	// Indicates an invalid byte slice was given to KeyFromBytes.
+	ErrInvalidKey = fmt.Errorf("invalid key")
```
jsign: Nit: comment should be prefixed with `ErrInvalidKey`.

core/thread/key.go
```diff
@@ -0,0 +1,122 @@
...
+	if len(b) == sym.KeyBytes*2 {
+		rk, err = sym.FromBytes(b[sym.KeyBytes:])
+		if err != nil {
+			return k, err
+		}
+	}
+	return Key{sk: sk, rk: rk}, nil
+}
+
+// KeyFromString returns a key by decoding a base58-encoded string.
```
jsign: Nit: Should this say base32? (considering the PR description and `String()`)

core/thread/key.go
```diff
@@ -0,0 +1,122 @@
...
+	return KeyFromBytes(b)
+}
+
+// Service returns the service key.
+func (k Key) Service() *sym.Key {
+	return k.sk
+}
+
+// Read returns the read key.
+func (k Key) Read() *sym.Key {
```
jsign: Totally agree that `Key` suffix is redundant, but looks weird. :)
I still think `Read` is the correct name though.

core/thread/key.go
```diff
@@ -0,0 +1,122 @@
...
+	return k.sk != nil
+}
+
+// CanRead returns whether or not read key is available.
+func (k Key) CanRead() bool {
+	return k.rk != nil
+}
+
+// Marshal returns raw key bytes while conforming to Marshaler.
+func (k Key) Marshal() ([]byte, error) {
```
jsign: The `Marshaler` interface mentioned in the comment [has one method](https://golang.org/pkg/encoding/json/#Marshaler) `MarshalJSON`, not `Marshal`. Also, `Marshaler` outputs json bytes which isn't our case.

Maybe calling this `MarshalBinary` to let `Key` implement [BinaryMarshaler](https://golang.org/pkg/encoding/#BinaryMarshaler), which is a raw binary marshaler?

crypto/symmetric/symmetric.go
```diff
@@ -30,52 +40,76 @@ func New() *Key {
...
+func FromString(k string) (*Key, error) {
+	_, b, err := mbase.Decode(k)
+	if err != nil {
+		return nil, err
+	}
+	return FromBytes(b)
+}
+
+// Marshal returns raw key bytes.
+func (k *Key) Marshal() ([]byte, error) {
```
jsign: Nit: may apply the `MarshalBinary`.

api/client/client_test.go
sanderpick: Oh, yep!

core/thread/key.go
sanderpick: Hmm, right. Sounds like _read_ a `io.Reader`.

core/thread/key.go
sanderpick: Nice, will do!

core/thread/key.go
jsign: Yeah, but just biased from the Go ecosystem. The coincidence of names!

net/api/pb/javascript/package.json
carsonfarmer: Nice. Cheers.

sanderpick:
Thanks for the reviews 🎉

Issue #139, opened 2019-12-02
Log owners should sign sent logs


sanderpick:
See https://github.com/textileio/go-textile-threads/pull/132#discussion_r351397479

Owners should be able to force overwriting log addresses. In order to enable that, the receiver will have to know if the sender is the owner. There's more to think about here, but this extra signature will be helpful.

PR #275, opened 2020-03-19
Re-enable PubSub

### LINKED ISSUE #116: Remove Time from thread event header
Time is currently handled in the event body, which is nice because it gives the higher level component more control. We can remove it from the header.

### LINKED ISSUE #175: Handle new-style PubSub topic handles

### LINKED ISSUE #266: Remove multi-head handling
Not needed!

sanderpick:
Closes #266 
Closes #116 
Closes #175 

Migrates to the the new PubSub API and re-enables publishing and subscription for records.

Additionally,
- Removes multi-head handling in `LogInfo`. I left the underlying store with the ability to handle multiple heads. Seems like some nice flex that doesn't get in the way if needed later.
- Removes the unused `Time` property of `core.Event`.


net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/textileio/go-threads/core/thread"
+	pb "github.com/textileio/go-threads/net/pb"
+)
+
+// TopicHandler receives all pushed thread records.
+type TopicHandler func(*pb.PushRecordRequest)
+
+// TopicManager manages thread pubsub topics.
+type TopicManager struct {
```
sanderpick: Manager hold all active thread topics, their peer event handlers, and subscriptions.

net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+	s *pubsub.Subscription
+}
+
+// NewTopicManager returns a new thread topic manager.
+func NewTopicManager(ctx context.Context, host peer.ID, ps *pubsub.PubSub, handler TopicHandler) *TopicManager {
+	return &TopicManager{
+		ctx:    ctx,
+		host:   host,
+		ps:     ps,
+		handle: handler,
```
sanderpick: This is a callback used by the server, which in the end only cares to receive the single push record requests, which are passed onto the normal `PushRecord` method.

net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if err := topic.t.Close(); err != nil {
+		return err
+	}
+	delete(m.m, id)
+	return nil
+}
+
+func (m *TopicManager) topicValidator(context.Context, peer.ID, *pubsub.Message) bool {
+	// @todo: determine if this is needed (related to host signatures)
```
sanderpick: Related to #139

sanderpick: (Will do next)

net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+	return true
+}
+
+// Publish a record request to a thread.
+func (m *TopicManager) Publish(ctx context.Context, id thread.ID, req *pb.PushRecordRequest) error {
```
sanderpick: And all that was just say we can have a normal Publish method again.

net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+			break
+		}
+		var msg string
+		switch pe.Type {
+		case pubsub.PeerJoin:
+			msg = "JOINED"
+		case pubsub.PeerLeave:
+			msg = "LEFT"
+		}
+		log.Infof("pubsub peer event: %s %s %s", pe.Peer.String(), msg, id.String())
```
sanderpick: This logging is currently the only use of `watch`, but it's handy. Maybe it can be used for something else in the future.

net/server.go
```diff
@@ -36,26 +37,30 @@ func newServer(n *net) (*server, error) {
...
-
 	return s, nil
 }
 
+func (s *server) topicsHandler(req *pb.PushRecordRequest) {
+	if _, err := s.PushRecord(s.net.ctx, req); err != nil {
+		// This error will be "log not found" if the record sent over pubsub
+		// beat the log, which has to be sent directly via the normal API.
+		// In this case, the record will arrive directly after the log via
+		// the normal API.
```
sanderpick: An important consideration...

core/net/event.go
```diff
@@ -31,9 +30,6 @@ type Event interface {
 type EventHeader interface {
 	format.Node
 
-	// Time returns the wall-clock time at which this event was created.
-	Time() (*time.Time, error)
```
carsonfarmer: 👍

core/thread/id.go
```diff
@@ -324,5 +324,5 @@ type LogInfo struct {
 	PubKey  crypto.PubKey
 	PrivKey crypto.PrivKey
 	Addrs   []ma.Multiaddr
-	Heads   []cid.Cid
+	Head    cid.Cid
```
carsonfarmer: 👍

net/net.go
```diff
@@ -188,6 +188,11 @@ func (t *net) CreateThread(_ context.Context, id thread.ID, opts ...core.KeyOpti
 	if err = t.store.AddLog(id, linfo); err != nil {
 		return
 	}
+
+	if err = t.server.addTopic(id); err != nil {
```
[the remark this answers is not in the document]
sanderpick: Oop, outdated by a method rename.. same concept.

logstore/logstore.go
```diff
@@ -198,8 +198,10 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
 	if err = ls.AddAddrs(id, lg.ID, lg.Addrs, pstore.PermanentAddrTTL); err != nil {
 		return err
 	}
-	if err = ls.AddHeads(id, lg.ID, lg.Heads); err != nil {
-		return err
+	if lg.Head.Defined() {
```
carsonfarmer: Much nicer

net/net.go
```diff
@@ -731,11 +741,7 @@ func (t *net) createRecord(ctx context.Context, id thread.ID, lg thread.LogInfo,
 		return nil, err
 	}
 
-	prev := cid.Undef
-	if len(lg.Heads) != 0 {
-		prev = lg.Heads[0]
-	}
-	return cbor.CreateRecord(ctx, t, event, prev, lg.PrivKey, sk)
+	return cbor.CreateRecord(ctx, t, event, lg.Head, lg.PrivKey, sk)
```
carsonfarmer: Very nice

net/api/service.go
```diff
@@ -315,16 +315,16 @@ func threadInfoToProto(info thread.Info) (*pb.ThreadInfoReply, error) {
 		for j, addr := range lg.Addrs {
 			addrs[j] = addr.Bytes()
 		}
-		heads := make([][]byte, len(lg.Heads))
-		for k, head := range lg.Heads {
-			heads[k] = head.Bytes()
+		var head []byte
+		if lg.Head.Defined() {
+			head = lg.Head.Bytes()
 		}
```
jsign: Nit: I think this if can be avoided, since `cid.Undef.Bytes() == empty slice`

net/client.go
```diff
@@ -355,8 +355,8 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
 	}
 
 	// Finally, publish to the thread's topic
-	if err = s.publish(id, req); err != nil {
-		log.Error(err)
+	if err = s.topics.Publish(ctx, id, req); err != nil {
+		return err
```
jsign: Flashing thought: Makes sense to relax if pubsub pushing fails? Maybe log and not err?

net/net.go
```diff
@@ -499,7 +509,7 @@ func (t *net) CreateRecord(ctx context.Context, id thread.ID, body format.Node)
 		return nil, err
 	}
 
-	log.Debugf("added record %s (thread=%s, log=%s)", rec.Cid().String(), id, lg.ID)
+	log.Debugf("added record %s (thread=%s, log=%s)", rec.Cid().String(), id.String(), lg.ID.String())
```
jsign: I'm curious about this change since I've seen it in other places. 
The `%s` format already calls `.String()`: [example](https://play.golang.org/p/CFm00mmwrEM)

Do you have this as a lint check or similar?

net/pubsub.go
```diff
@@ -0,0 +1,165 @@
...
+		if err = proto.Unmarshal(msg.Data, req); err != nil {
+			log.Errorf("error unmarshaling published record", err)
+			continue
+		}
+
+		log.Debugf("received multicast record from %s", from.String())
+
+		m.handle(req)
+	}
+}
```
jsign: Some comments/questions:

1) The `New..` receives a `ctx` that it's used to cancel new items processing. But the process has no way of knowing if some things is being processed before, say, finishing the server. That might make a hard-stop of the logic at any point of the event handler. Maybe that's safe because all logic have store-transactionality.

2) On every registration, a new goroutine is fired, but I'm not sure they've terminated accordingly on a `Remove` call. Goroutines only depend on the manager's context. So it seems they'd leak.

3) What do you think of making `handler` also receive a `ctx`? Seeing `net/server.go` I see that indirectly uses `n.ctx` but I'm not sure that might be the best. You can use the manager `ctx` in the handler, so to also cancel on-going pubsub handling if the manager receives a cancellation.

Some fuzzy writing can get into more details/suggestions or comments if something isn't clear enough!

jsign: Regarding 2., I'd ask: that `break` if `.Next` is returned is assumed that the topic was closed? If that's the only case that could error, I guess that's the goroutine finalization logic!

net/client.go
sanderpick: Yep, good call.

net/net.go
sanderpick: Oh yeah, the the usage is mixed so I was just going for consistency. I'll take a quick look and change what I can see in `net`.

net/pubsub.go
sanderpick: 1) Right, I think it's safe since records coming in over pubsub end up just hitting the server's `PushRecord` method, as if they had come in _normally_, which has its own safety measures.

2) The subscription `.Next` does receive an error "subscription cancelled by calling sub.Cancel()", which triggers the `break`, which is all triggered by `Remove`ing the topic. However, the `EventHandler` does not seems to function that way. I'll see about using a different context.

3) 👍

net/server.go
```diff
@@ -36,33 +37,38 @@ func newServer(n *net) (*server, error) {
...
-	//}
-
-	// @todo: s.pubsub.RegisterTopicValidator()
-
 	return s, nil
 }
 
+// pubsubHandler receives records over pubsub.
+func (s *server) pubsubHandler(ctx context.Context, req *pb.PushRecordRequest) {
+	if _, err := s.PushRecord(ctx, req); err != nil {
```
sanderpick: Context now coming from the pubsub topic.

net/pubsub.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	topic.s.Cancel()
+	topic.h.Cancel()
+	if err := s.ps.UnregisterTopicValidator(id.String()); err != nil {
+		return err
+	}
+	if err := topic.t.Close(); err != nil {
+		return err
+	}
+	topic.cancel()
```
sanderpick: Topic now has its own cancel func.

net/pubsub.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	h, err := pt.EventHandler()
+	if err != nil {
+		return err
+	}
+	if err = s.ps.RegisterTopicValidator(id.String(), s.topicValidator); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(s.ctx)
```
sanderpick: And context derived from the parent, which in this case is `net`'s context.

PR #277, opened 2020-03-20
All json all the time

asutula:
This PR removes the two db modes of normal go mode and json mode. The only mode now is what was called json mode, there is no "mode" nomenclature any more.

Also snuck in an update to better support creating `Collection`s as part of db creation.

The `Collection` and `jsonpatcher` apis still take `interface{}` arguments event though all the data is `string` and `*string`. I plan on cleaning that up in the next PR where all methods will explicitly accept and return `[]byte` for json data.


core/db/db.go
```diff
@@ -24,8 +24,10 @@ func (e InstanceID) String() string {
 }
 
 func IsValidInstanceID(instanceID string) bool {
-	_, err := uuid.Parse(instanceID)
-	return err == nil
+	// ToDo: Decide how we want to apply instance id requirements
+	return len(instanceID) > 0
+	// _, err := uuid.Parse(instanceID)
+	// return err == nil
```
asutula: Disabled this for now, made the issue https://github.com/textileio/go-threads/issues/280

db/collection.go
```diff
@@ -42,20 +41,7 @@ type Collection struct {
...
-		schemaLoader: schemaLoader,
-		valueType:    reflect.TypeOf(defaultInstance),
-		db:           d,
-		indexes:      make(map[string]Index),
-	}
-	return c
-}
-
-func newCollectionFromSchema(name string, schema string, d *DB) *Collection {
+func newCollection(name string, schema string, d *DB) *Collection {
```
asutula: This is where we'll want some schema verification of `ID`

https://github.com/textileio/go-threads/issues/281

db/collection_test.go
```diff
@@ -34,35 +35,48 @@ func TestMain(m *testing.M) {
...
+	// 	db, clean := createTestDB(t)
+	// 	defer clean()
+	// 	cc := CollectionConfig{
+	// 		Name:   "FailingType",
+	// 		Schema: util.SchemaFromInstance(&FailingType{}),
+	// 	}
+	// 	if _, err := db.NewCollection(cc); err != ErrInvalidCollectionType {
+	// 		t.Fatal("the collection should be invalid")
+	// 	}
+	// })
```
asutula: This is disabled until we start verifying that the schema specifies ID, tracking here https://github.com/textileio/go-threads/issues/281

db/db.go
```diff
@@ -114,7 +112,7 @@ func NewDBFromAddr(ctx context.Context, network net.Net, addr ma.Multiaddr, key
 
 // newDB is used directly by a db manager to create new dbs
 // with the same config.
-func newDB(n net.Net, id thread.ID, config *Config, collections ...CollectionConfig) (*DB, error) {
+func newDB(n net.Net, id thread.ID, config *Config) (*DB, error) {
```
asutula: `CollectionConfig`s are now passed in as part of `Config`

db/manager.go
```diff
@@ -187,16 +187,17 @@ func (m *Manager) Close() error {
...
 		RepoPath: base.RepoPath,
 		Datastore: wrapTxnDatastore(base.Datastore, kt.PrefixTransform{
 			Prefix: dsDBManagerBaseKey.ChildString(id.String()),
 		}),
-		EventCodec: base.EventCodec,
-		JsonMode:   base.JsonMode,
-		Debug:      base.Debug,
+		EventCodec:  base.EventCodec,
+		Debug:       base.Debug,
+		Collections: append(base.Collections, collections...),
```
asutula: We combine the default collections that the manager was created with (using `Option`s) with any default collections specified in the call to `manager.NewDB` or `manager.NewDBFromAddress`.

db/options.go
```diff
@@ -20,16 +20,16 @@ type Option func(*Config) error
...
-	EventCodec core.EventCodec
-	JsonMode   bool
-	Debug      bool
-	LowMem     bool
+	RepoPath    string
+	Datastore   ds.TxnDatastore
+	EventCodec  core.EventCodec
+	Debug       bool
+	LowMem      bool
+	Collections []CollectionConfig
```
asutula: Default collections are now part of the db config.

db/options.go
```diff
@@ -81,3 +74,12 @@ func WithEventCodec(ec core.EventCodec) Option {
...
 }
+
+// WithCollections is used to specify collections that
+// will be created
+func WithCollections(collections ...CollectionConfig) Option {
+	return func(sc *Config) error {
+		sc.Collections = collections
+		return nil
+	}
+}
```
asutula: Option func for setting collections

db/query_more_test.go
```diff
@@ -0,0 +1,166 @@
+package db
```
asutula: These were the old non-json mode query tests. I figured they might cover some use cases not covered by the json mode tests, so I left them here and ported them to json mode.

asutula: Probably some overlap, but cool

examples/e2e_counter/reader.go
```diff
@@ -19,23 +20,27 @@ func runReaderPeer(repo string) {
 	checkErr(err)
 	defer n.Close()
 
-	d, err := db.NewDBFromAddr(context.Background(), n, writerAddr, key, db.WithRepoPath(repo))
+	cc := db.CollectionConfig{
+		Name:   "counter",
+		Schema: util.SchemaFromInstance(&myCounter{}),
+	}
+
+	d, err := db.NewDBFromAddr(context.Background(), n, writerAddr, key, db.WithRepoPath(repo), db.WithCollections(cc))
```
asutula: Heres an example of creating a db from invite and passing in a collection config so the auto start works nicely

util/util.go
```diff
@@ -130,3 +132,27 @@ func MustParseAddr(str string) ma.Multiaddr {
...
+	}
+	val := string(JSON)
+	return &val
+}
+
+func InstanceFromJSONString(s string, i interface{}) {
+	if err := json.Unmarshal([]byte(s), i); err != nil {
+		panic(err)
+	}
+}
```
asutula: Some utilities for converting between go types and json strings

util/util.go
```diff
@@ -130,3 +132,27 @@ func MustParseAddr(str string) ma.Multiaddr {
...
 }
+
+func SchemaFromInstance(i interface{}) string {
+	schema := jsonschema.Reflect(i)
+	JSON, err := json.MarshalIndent(schema, "", "  ")
+	if err != nil {
+		panic(err)
+	}
+	return string(JSON)
+}
```
asutula: Utility to create a json schema from a go type

db/manager.go
jsign: Nice

examples/e2e_counter/reader.go
jsign: Feels good 💯 💯 💯

examples/e2e_counter/reader.go
```diff
@@ -9,6 +9,7 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
```
jsign: A quick refresher: This `e2e_counter` was some of the first e2e things that I created when we had a first working version of _DB_. We have chatted some times to remove this code if may be confusing or adding too much burden to refactors or maintainability.  No strong opinion, but just refreshing this since I _originally_ made this _example_... no feelings with it. If it adds value in some direction, np with leaving this too.

The _books_ example was made roughly at the same time... but feels that example it's much nicer to show somebody a minimal way of using things.

Just leaving al this as a note here.

asutula:
Thanks for the review @jsign. Yea, all that reflect stuff goes away with the removal of `interface{}` and use of `[]byte`. I have that going on a branch now and it feels good, cleans up the code a lot.

examples/e2e_counter/reader.go
asutula: Ah ok. If anyone has some strong feeling about removing it, I'm fine with that, but also don't have a strong opinion.

db/options.go
sanderpick: Perf!

examples/e2e_counter/reader.go
sanderpick: Yep!

PR #279, opened 2020-03-20
Finish request verifications

### LINKED ISSUE #139: Log owners should sign sent logs

### LINKED ISSUE #278: PushLogs should require verification
`PushRecords` and `PushLogs` should require payload signing and verification. Not needed for `GetRecords` and `GetLogs`, as these are more of an ACL concern.

### LINKED ISSUE #287: Thread credentials
Both the `Net` and `DB` APIs are currently wide open. The server can intercept requests and do global authentication (like in textile), but this isn't enough to act on ACL rules embedded (in the near future) in the thread ID. This should be done more natively. 

This can be seen more clearly by looking at the different `Net` methods. For each one, we can note what kind of authentication is needed. The mechanisms at our disposal are:

- At the most global level, Firewalls (mentioning for completeness)
- gRPC Auth, checked against some external resource (like Textile Developers / Users)
- An identity proof + thread ACL, or in others words, a proof that you are who you say you are. This is achieved with a public key (identity) and a proof (signature) that you have the corresponding private key. 

- `GetHostID`: (Firewall + Auth)
- `CreateThread`: (Firewall + Auth)
- `AddThread`: (Firewall + Auth)
- `GetThread`: (Firewall + Auth + ACL)
- `PullThread`: (Firewall + Auth + ACL)
- `DeleteThread`: (Firewall + Auth + ACL)
- `AddReplicator`: (Firewall + Auth + ACL)
- `CreateRecord`: (Firewall + Auth + ACL)
- `AddRecord`: (Firewall + Auth + ACL)
- `GetRecord`: (Firewall + Auth + ACL)
- `Subscribe`: (Firewall + Auth + ACL)

So, ACLs are needed for authentication _within a thread_. Authentication outside of a thread must be handled by some other means, e.g, _who is allowed to create a thread?_. The answer to this will be controlled by the server owner, using firewalls and/or gRPC Auth.

On the other hand, ACL checks can be performed natively in the `Net` layer. _Thread credentials_ can be sent allow with requests to provide a verifiable record of who is calling. This record can (later) be used to determine whether or not the caller has adequate permission.

sanderpick:
Closes #139 
Closes #278
Closes #287 

There are a couple related by different changes here:

- Adds an additional author signature to log records.
  - `CreateRecord` uses the local host as author.
  - `AddRecord` takes a fully composed record, so it's up to the caller to use the correct author (to be later checked against an ACL).
- Completes verifications in the p2p networking API (get/push logs, get/push records). Previously, only `PushRecord` performed any kind of verification. Now, each request carries an inner body message which is signed by the sender.
- Adds thread credentials (see #287 for a detailed overview)


api/client/client.go
```diff
@@ -86,15 +87,23 @@ func (c *Client) Close() error {
 }
 
 // NewDB creates a new DB with ID
-func (c *Client) NewDB(ctx context.Context, id thread.ID) error {
-	_, err := c.c.NewDB(ctx, &pb.NewDBRequest{
-		DbID: id.String(),
+func (c *Client) NewDB(ctx context.Context, creds thread.Credentials) error {
```
sanderpick: Both these new DB methods now take credentials instead of a plain ID. Most APIs that took ID now take credentials. See below for what credentials are.

api/pb/android/threads/src/test/java/io/textile/threads/ClientUnitTest.java
```diff
@@ -24,7 +26,7 @@
 public class ClientUnitTest {
 
     static Client client;
-    static String dbId = "bafk7ayo2xuuafgx6ubbcn2lro3s7oixgujdda6shv4";
+    static String dbId = "AVXwYdq9KAKa/qBCJulxduX3IuaiRjB6R68=";
```
sanderpick: Since we don't have a Java Thread ID lib, I'm just hard-hardcoding a base64 representation of a raw thread ID bytes.

cbor/record.go
```diff
@@ -20,31 +20,45 @@ func init() {
 
 // record defines the node structure of a record.
 type record struct {
-	Block cid.Cid
-	Sig   []byte
-	Prev  cid.Cid `refmt:",omitempty"`
+	Block     cid.Cid
+	Sig       []byte
+	AuthorSig []byte
```
sanderpick: Here's the new signature. When a node receives a record, it should be able to independently verify who wrote it, which may not be the sender, e.g., on an initial thread join, where all logs and records are pushed.

core/net/net.go
```diff
@@ -30,35 +30,38 @@ type API interface {
 	// GetHostID returns the host's peer id.
 	GetHostID(ctx context.Context) (peer.ID, error)
 
-	// CreateThread with id.
-	CreateThread(ctx context.Context, id thread.ID, opts ...KeyOption) (thread.Info, error)
+	// CreateThread with credentials.
+	CreateThread(ctx context.Context, creds thread.Credentials, opts ...KeyOption) (thread.Info, error)
```
sanderpick: Again, `ID` is replaced by `Credentials` on all APIs.

core/thread/credentials.go
```diff
@@ -0,0 +1,98 @@
+package thread
```
sanderpick: This is where the new `Credentials` are defined.

core/thread/credentials.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Credentials are used to prove a caller identity, which is needed
+// for thread ACL checks.
+type Credentials interface {
+	// ThreadID must return the signed thread ID.
+	ThreadID() ID
+
+	// Sign must return the public key and signature which can
+	// be used to verify the caller.
+	Sign() (crypto.PubKey, Signature, error)
```
sanderpick: This method provides a way for a caller to conceal their private key for signing.

core/thread/credentials.go
```diff
@@ -0,0 +1,98 @@
...
+		opt(args)
+	}
+	return DefaultCreds{threadID: id, privKey: args.PrivKey}
+}
+
+func (c DefaultCreds) ThreadID() ID {
+	return c.threadID
+}
+
+func (c DefaultCreds) Sign() (crypto.PubKey, Signature, error) {
```
sanderpick: The default way to use credentials comes with a signing method that used the provided private key. Note, the caller does not have access to the key.

core/thread/credentials.go
```diff
@@ -0,0 +1,98 @@
...
+// SignedCreds are useful when the private key is not available.
+// In this case, the public key and signature are used directly.
+type SignedCreds struct {
+	threadID  ID
+	pubKey    crypto.PubKey
+	signature Signature
+}
+
+// NewSignedCreds returns new pre-signed credentials from the source.
+func NewSignedCredsFromBytes(id, pubKey, signature []byte) (creds SignedCreds, err error) {
```
sanderpick: This is used by APIs where the caller is across a network boundary... the signature had already been made, so `Sign` will just return it.

db/manager.go
```diff
@@ -201,3 +214,42 @@ func getDBConfig(id thread.ID, base *Config, collections ...CollectionConfig) *C
 		Collections: append(base.Collections, collections...),
 	}
 }
+
+// putDBCreds persists db creds for a thread.
+func (m *Manager) putDBCreds(id thread.ID, creds thread.Credentials) error {
```
sanderpick: In order to restart DBs, the manager needs to save credentials. This mean storing a public key and the thread signature if they are present.

db/manager.go
```diff
@@ -89,15 +97,19 @@ func NewManager(network net.Net, opts ...Option) (*Manager, error) {
 }
 
 // NewDB creates a new db and prefixes its datastore with base key.
-func (m *Manager) NewDB(ctx context.Context, id thread.ID, collections ...CollectionConfig) (*DB, error) {
+func (m *Manager) NewDB(ctx context.Context, creds thread.Credentials, collections ...CollectionConfig) (*DB, error) {
```
sanderpick: Again, using credentials.

net/api/client/client_test.go
```diff
@@ -202,11 +202,16 @@ func TestClient_AddRecord(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
-	info, err := client.CreateThread(context.Background(), id, core.ThreadKey(tk), core.LogKey(pk))
+	authorSk, _, err := crypto.GenerateEd25519Key(crand.Reader)
+	if err != nil {
+		t.Fatal(err)
+	}
+	creds := thread.NewDefaultCreds(id, thread.WithPrivKey(authorSk))
+	_, err = client.CreateThread(context.Background(), creds, core.WithThreadKey(tk), core.WithLogKey(logPk))
```
sanderpick: Here's an example of creating credentials that actually have an author. The default thread creds constructor can take an optional private key.

net/api/service.go
```diff
@@ -272,6 +275,10 @@ func marshalPeerID(id peer.ID) []byte {
 	return b
 }
 
+func getCredentials(c *pb.Credentials) (thread.Credentials, error) {
+	return thread.NewSignedCredsFromBytes(c.ThreadID, c.PubKey, c.Signature)
```
sanderpick: Using the signed creds...

net/client.go
```diff
@@ -36,13 +38,21 @@ func (s *server) getLogs(ctx context.Context, id thread.ID, pid peer.ID) ([]thre
...
 
-	req := &pb.GetLogsRequest{
-		Header: &pb.GetLogsRequest_Header{
-			From: &pb.ProtoPeerID{ID: s.net.host.ID()},
-		},
+	body := &pb.GetLogsRequest_Body{
 		ThreadID:   &pb.ProtoThreadID{ID: id},
 		ServiceKey: &pb.ProtoKey{Key: sk},
 	}
+	sig, key, err := s.signRequestBody(body)
```
sanderpick: All these requests now sign the body with the net host's key.

net/net.go
```diff
@@ -157,11 +155,12 @@ func (t *net) Store() lstore.Logstore {
 	return t.store
 }
 
-func (t *net) GetHostID(context.Context) (peer.ID, error) {
+func (t *net) GetHostID(_ context.Context) (peer.ID, error) {
 	return t.host.ID(), nil
 }
 
-func (t *net) CreateThread(_ context.Context, id thread.ID, opts ...core.KeyOption) (info thread.Info, err error) {
+func (t *net) CreateThread(_ context.Context, creds thread.Credentials, opts ...core.KeyOption) (info thread.Info, err error) {
```
sanderpick: net/net.go is the layer where the ACL checks will eventually happen (not in this PR). Create/add thread can ensure that the caller is not locking themselves out of the thread.

net/pubsub.go
```diff
@@ -151,23 +151,39 @@ func (s *PubSub) subscribe(ctx context.Context, id thread.ID, topic *topic) {
...
-			continue
-		}
-
 		log.Debugf("received multicast record from %s", from)
 
 		s.handler(ctx, req)
 	}
 }
+
+func (s *PubSub) handleMsg(m *pubsub.Message) (from peer.ID, rec *pb.PushRecordRequest, err error) {
```
sanderpick: Just cleaning up.

core/thread/credentials.go
```diff
@@ -0,0 +1,98 @@
...
+
+// DefaultCreds are useful for either threads that don't require
+// credentials or when the private key is available for local signing.
+type DefaultCreds struct {
+	threadID ID
+	privKey  crypto.PrivKey
+}
+
+// NewDefaultCreds returns new default credentials.
+func NewDefaultCreds(id ID, opts ...CredOption) Credentials {
```
sanderpick: Important note: Credentials with a signing key / signature are only required for cases where the underlying thread is private, which is (will be) defined in the ACL. If the thread is wide open, anyone can do anything.

core/thread/credentials.go
jsign: Looking below, `Signature` is the `thread.ID` signed with the corresponding pk of `cyrpto.PubKey`.
This means that the signature is always the same for that thread.ID. This means that the same `Signature` can be reused as-is in different records.

I'm a bit confused at this point. Why is this necessary? Say, instead of doing the usual thing of the author signing whatever is sending and the caller to verify just that thing that was signed?
I can't imagine how using an "static" signature can be useful for other things.

I'm continuing on the review, so that might click later. Just wanted to share the thought by just looking at this interface!

cbor/record.go
jsign: So, `Sig` is the signature using the Log private key.
And `AuthorSig` is the signature by the owner of the Log?

Not clear yet why both things are necessary at the `record` layer, since whoever owns the private keys of the Log is, de facto, the author. (Or not, and I'm confused).

core/thread/credentials.go
jsign: OK, so seems that the _network boundary_ is the reason that `Sign()` always is deterministic in the output (note receiving params).

This seems to indicate that `Signature` is more a _Key_ than a _Signature_... seems to give another component _something_ to pretend being him *independent of the action/data*.

cbor/record.go
sanderpick: Maybe it wasn't clear in the description, the author sig will be used to check against the thread ACL. My thinking is that the ACL will reference author pub keys, not log pub keys, since an author can have many logs. The ACL implementation (and verifications) are not in the PR.

jsign: Oh, I see.
A next question is: it make sense that a Log is owned by more than one author?

Because if a log is only owned by one author, it seems that the ACL check should be against the Log (which indirectly has the owner, but maybe not stored here).

If a log can be written by more than one author, it seems that isn't a single-writer log?

jsign: Well, that _not stored here_ thing might be the tricky part. Where if not here it can live. Not completely sure. Have you thought any other option than this one?

core/thread/credentials.go
sanderpick: Yep, also forgot to mention this weirdness among all the other comments. I did start down the path of a more normal signature flow (the entire message envelope / body, as in the net layer), but then ran into an issue with restarting DBs. Since these signatures aren't needed yet (the ACL is not implemented), I just went with this simpler solution to get the API in. I think we'll want to have some _skeleton key_ or other way for a db to not need verification _under the hood_, in single thread adapter, but more to consider there.

cbor/record.go
sanderpick: Take the case of an ETH or 3Box key, or whatever other key you want to think of as _your identity_, these would be the author key. So, you'd have one log per author per thread, which is single writer.

Do you mean where else could the author sig live? Originally, I had thought it was only needed in the over-the-wire messages (pb messages), but then realized that sometimes you get records from the non-author, e.g., when you're getting pushed a full log from a replicator.

jsign: Yeah, sure... I can receive logs from the author, or some other peer that is sharing the info.

But for ACL verification, I care about that distinction? Maybe that's the point I'm missing.
If the ACL says _Author1 can't write to the thread_ , I can apply that rule not caring about if I received a new record from Author1 owned-log from himself directly or from a sharing peer.

No matter how sends me a new record from a Log owned by Author1, I'll deny it.

jsign: To be clear, I'm forcing some questions just to double-check if make sense to you. I think for me will click when the ACL impl lands, and see the pieces together. So feel free to dismiss a bit my noise.

sanderpick: Right, that's why it the signature should be in the record, and not in the over-the-wire message... sorry, that comment above may have been confusing, I was trying to address your comment about where to store the signature. Though, I may have misunderstood the question. 

Another reason author key != log key, is because log are created as needed when a peer joins. But the static ACL is created when the thread is created... at which time, you may not know all the log keys.

sanderpick: The noise is good! I ran myself in a couple circle with all this stuff... it's good to chat about it and get push back, so the whole thing comes out right!

jsign: Cool, we can continue this chat on another place. Will ✔️

sanderpick:
Yeah totally, the API ease-of-use has held me up here a couple times. The the main reason to let people choose there thread ID is so that they can decide what type to use and how many bytes of randomness. Sort of like choose bit size for a symmetric key. The credentials contain a signature that is based on the thread ID, so they go hand-in-hand. Naming this stuff is hard... any suggestions welcome.

cbor/record.go
```diff
@@ -20,31 +20,45 @@ func init() {
...
+	payload := config.Block.Cid().Bytes()
+	if config.Prev.Defined() {
+		payload = append(payload, config.Prev.Bytes()...)
 	}
-	sig, err := sk.Sign(payload)
+	sig, err := config.Key.Sign(payload)
+	if err != nil {
+		return nil, err
+	}
+	asig, err := config.AuthorKey.Sign(payload)
```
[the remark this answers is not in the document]
sanderpick: One thing to remember here, this can happen on the client when you use `AddRecord`. As I mentioned in the PR description, the host key is used for `CreateRecord`, because in this case, your relying on the remote to create the entire thing. 

And it follows... since the client is doing the authoring, they could use whatever key they want... it just better match the one in the thread ACL. You can imagine a case where the remote has the log private key and not the author's private key.

cbor/record.go
sanderpick: If you implement the Credentials interface, you could handling the signing however you want... use a local key, call to another service, etc. Both ways are possible, unless I'm missing something.

core/thread/credentials.go
sanderpick: @carsonfarmer did my comment above answer your question about the static signature?

cbor/record.go
carsonfarmer: Ooooh, ok fuck yes I hadn't thought about that. We (Textile) could provide a nice wrapper class that'll do this. That opens up sooo many possibilities!

sanderpick:
Thanks for the reviews, guys. I wanted everyone to look at this so we can all start thinking about it. I'm gonna merge now, but may follow up over the weekend with some ideas on how to solve the issue around the manager needing to save creds in order to restart the DBs.

PR #282, opened 2020-03-24
[]byte api

asutula:
Now that the db only deals in json data, this PR updates things to represent all json data as `[]byte`.

There was a mixture of `string` and `[]byte` data in the grpc interface, now it's all `[]byte`.

Also updates `Create` to not modify passed in instances and instead returns a list of instance ids.

Because of those last two points, there will be some downstream client effects cc @andrewxhill and @carsonfarmer


api/client/client.go
```diff
@@ -139,42 +139,35 @@ func (c *Client) NewCollection(ctx context.Context, dbID thread.ID, config db.Co
 }
 
 // Create creates new instances of objects
-func (c *Client) Create(ctx context.Context, dbID thread.ID, collectionName string, items ...interface{}) error {
-	values, err := marshalItems(items)
+func (c *Client) Create(ctx context.Context, dbID thread.ID, collectionName string, items ...interface{}) ([]string, error) {
```
asutula: Client `Create` now returns list of instance ids, doesn't modify instances passed in.

api/client/client_test.go
```diff
@@ -132,9 +132,10 @@ func TestSave(t *testing.T) {
 
 		person := createPerson()
 
-		err = client.Create(context.Background(), dbID, collectionName, person)
+		ids, err := client.Create(context.Background(), dbID, collectionName, person)
 		checkErr(t, err)
 
+		person.ID = ids[0]
```
asutula: Here's how you now have to update your local model's ID after creating the instance in the db.

api/pb/api.proto
```diff
@@ -98,7 +98,7 @@ message FindByIDRequest {
 }
 
 message FindByIDReply {
-    string instance = 1;
+    bytes instance = 1;
```
asutula: All bytes in here now for json data

core/db/db.go
```diff
@@ -59,9 +59,9 @@ type Action struct {
 	// CollectionName of the instance in action
 	CollectionName string
 	// Previous is the instance before the action
-	Previous interface{}
+	Previous []byte
 	// Current is the instance after the action was done
-	Current interface{}
+	Current []byte
```
asutula: `Action` now explicitly sends `[]byte`s

db/collection.go
```diff
@@ -95,19 +95,23 @@ func (c *Collection) WriteTxn(f func(txn *Txn) error) error {
 	return c.db.writeTxn(c, f)
 }
 
-// FindByID finds an instance by its ID and saves it in v.
+// FindByID finds an instance by its ID.
 // If doesn't exists returns ErrNotFound.
-func (c *Collection) FindByID(id core.InstanceID, v interface{}) error {
-	return c.ReadTxn(func(txn *Txn) error {
-		return txn.FindByID(id, v)
+func (c *Collection) FindByID(id core.InstanceID) (instance []byte, err error) {
```
asutula: Direct go api changes here... find by id returns bytes now and below crate returns the list of ids

asutula:
Opening this up for review so I can address any concerns asap, working on updating the android client now...

asutula:
Android code updated and tests passing

core/db/db.go
jsign: `interface{}` :wave:

PR #283, opened 2020-03-25
Validate Collection Schema

### LINKED ISSUE #281: Validate collection schema on creation
Now that we have pure json mode, there is no validation of the schema being used to create a `Collection`. We'll want to verify that the schema specifies the `ID` field.

Initial thoughts are to write an ID-specifying schema to validate the usr-provided schema.

asutula:
This one does a couple things... 

Makes `CollectionConfig.Schema` a proper `Schema` struct. This is nice because it provided a programatic API to the schema we can use for validation. Also makes the API nicer to work with not having to pass in a schema string.

Performs validation of the provided schema when a new collection is created. Currently, that means checking that the schema defines an `ID` property of type `string`.

closes #281


db/collection.go
```diff
@@ -41,16 +43,39 @@ type Collection struct {
...
+		typeName := parts[len(parts)-1]
+		refDefinition := schema.Definitions[typeName]
+		if refDefinition == nil {
+			return nil, ErrInvalidCollectionSchema
+		}
+		properties = refDefinition.Properties
+	}
+	if !hasIDProperty(properties) {
+		return nil, ErrInvalidCollectionSchema
+	}
```
asutula: Above is the schema validation. As far as I can tell, there are two ways an `ID` might be defined, and we have to handle them both. First is the "expanded" schema case where a type is defined in the top level of the schema. Second is when the top level schema specifies a `Ref` to type nested in the schema `Definitions`.

api/pb/api.proto
```diff
@@ -12,7 +12,7 @@ message NewDBRequest {
 
 message CollectionConfig {
     string name = 1;
-    string schema = 2;
+    bytes schema = 2;
```
asutula: Schema was always json data, so make it `bytes` like the rest of the json data we deal with

db/collection.go
asutula: I still like the idea of validating the schema against another schema, but this was a more clear approach for now.

db/collection_test.go
```diff
@@ -53,30 +63,42 @@ func TestNewCollection(t *testing.T) {
...
+		db, clean := createTestDB(t)
+		defer clean()
+		cc := CollectionConfig{
+			Name:   "FailingType",
+			Schema: util.SchemaFromInstance(&FailingType{}, true),
+		}
+		if _, err := db.NewCollection(cc); err != ErrInvalidCollectionSchema {
+			t.Fatal("the collection should be invalid")
+		}
+	})
```
asutula: Tests reenabled to make sure schema validation is actually working

db/db.go
```diff
@@ -190,7 +197,7 @@ func (d *DB) reCreateCollections() error {
 // CollectionConfig describes a new Collection.
 type CollectionConfig struct {
 	Name    string
-	Schema  string
+	Schema  *jsonschema.Schema
```
asutula: Not just a string anymore

util/util.go
```diff
@@ -135,13 +135,18 @@ func MustParseAddr(str string) ma.Multiaddr {
 	return addr
 }
 
-func SchemaFromInstance(i interface{}) string {
-	schema := jsonschema.Reflect(i)
-	JSON, err := json.MarshalIndent(schema, "", "  ")
-	if err != nil {
+func SchemaFromInstance(i interface{}, expandedStruct bool) *jsonschema.Schema {
```
asutula: Most of the noise in this PR is because of adding the `expandedStruct` option here. When set to `true`, the generated schema will use top level type definitions instead of using a ref to a nested definition. We need to be able to test both cases.

asutula:
Ok, ready for review here.

asutula:
Android failure is that rate limit related to some GH action thing... Should be fine when it re-runs soon.

Issue #286, opened 2020-03-26
Backfill index on creation


asutula:
Currently we start building the index from the time it was created, but we should go though all existing data and index it as well.

asutula:
Feeling like this might be quite complex. Indexing can take an unbounded amount of time, so that has me thinking it will have to be backed by a durable, persistent, job queue kinda setup.

asutula:
You'd also want it to be non-blocking

sanderpick:
I think most DBs block while doing an initial index (or back filling), e.g., (https://docs.mongodb.com/manual/core/index-creation/), but some of the more modern one (Rethink?) do not. The simpler solution might be a good first step.

sanderpick:
Our situation might actually be a bit more dangerous though, since we're handing full admin control over to a dev with a client lib. At least they can only mess up their own DB (once ACLs are a thing).

PR #288, opened 2020-03-27
Always persist IndexConfig

### LINKED ISSUE #284: Add ID index on Collection creation
Looks like it got lost along the way somehow

### LINKED ISSUE #285: Index definition isn't persisted
In the case of adding an index to an existing Collection, the index definition isn't persisted, so on DB restart, it won't exist anymore. Index definitions are persisted when provided at Collection creation time.

asutula:
Adds tests for adding indexes as well

fixes #285 
fixes #284

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -93,8 +93,46 @@ func (c *Collection) Indexes() map[string]Index {
 // Adding an index will override any overlapping index values if they already exist.
 // @note: This does NOT currently build the index. If items have been added prior to adding
 // a new index, they will NOT be indexed a posteriori.
-func (c *Collection) AddIndex(path string, unique bool) error {
-	c.indexes[path] = Index{
+func (c *Collection) AddIndex(config IndexConfig) error {
+	indexKey := dsDBIndexes.ChildString(c.name)
+	exists, err := c.db.datastore.Has(indexKey)
+	if err != nil {
+		return err
+	}
+
+	indexes := map[string]IndexConfig{}
+
+	if exists {
+		indexesBytes, err := c.db.datastore.Get(indexKey)
+		if err != nil {
+			return err
+		}
+		if err = json.Unmarshal(indexesBytes, &indexes); err != nil {
+			return err
+		}
+	}
+
+	// if the index being added is for path ID
+	if config.Path == idFieldName {
+		// and there already is and index on ID
+		if _, exists := indexes[idFieldName]; exists {
+			// just return gracefully
+			return nil
+		}
+	}
+
+	indexes[config.Path] = config
+
+	indexBytes, err := json.Marshal(indexes)
+	if err != nil {
+		return err
+	}
+
+	if err := c.db.datastore.Put(indexKey, indexBytes); err != nil {
+		return err
+	}
+
+	c.indexes[config.Path] = Index{
 		IndexFunc: func(field string, value []byte) (ds.Key, error) {
 			result := gjson.GetBytes(value, field)
 			if !result.Exists() {
@@ -102,7 +140,7 @@ func (c *Collection) AddIndex(path string, unique bool) error {
 			}
 			return ds.NewKey(result.String()), nil
 		},
-		Unique: unique,
+		Unique: config.Unique,
 	}
 	return nil
 }
--- a/db/db.go
+++ b/db/db.go
@@ -177,16 +177,21 @@ func (d *DB) reCreateCollections() error {
 			return err
 		}
 
-		var indexes []IndexConfig
+		var indexes map[string]IndexConfig
 		index, err := d.datastore.Get(dsDBIndexes.ChildString(name))
 		if err == nil && index != nil {
 			_ = json.Unmarshal(index, &indexes)
 		}
 
+		indexValues := make([]IndexConfig, len(indexes))
+		for _, value := range indexes {
+			indexValues = append(indexValues, value)
+		}
+
 		if _, err := d.NewCollection(CollectionConfig{
 			Name:    name,
 			Schema:  schema,
-			Indexes: indexes,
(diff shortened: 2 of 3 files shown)


db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
 // Adding an index will override any overlapping index values if they already exist.
 // @note: This does NOT currently build the index. If items have been added prior to adding
 // a new index, they will NOT be indexed a posteriori.
-func (c *Collection) AddIndex(path string, unique bool) error {
-	c.indexes[path] = Index{
+func (c *Collection) AddIndex(config IndexConfig) error {
```
asutula: All persistence of `IndexConfig` moved in here so that it is persisted no matter if this is called when creating a new Collection from the DB or adding an index on an already created Collection.

db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
...
+	indexKey := dsDBIndexes.ChildString(c.name)
+	exists, err := c.db.datastore.Has(indexKey)
+	if err != nil {
+		return err
+	}
+
+	indexes := map[string]IndexConfig{}
```
asutula: Moved the storage format from `[]IndexConfig` to `map[string]IndexConfig` to make it easier to overwrite existing index configs and easier to detect an already existing index on `ID`

db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
...
+
+	if exists {
+		indexesBytes, err := c.db.datastore.Get(indexKey)
+		if err != nil {
+			return err
+		}
+		if err = json.Unmarshal(indexesBytes, &indexes); err != nil {
```
asutula: Read persisted configs into memory

db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
...
+			return err
+		}
+	}
+
+	// if the index being added is for path ID
+	if config.Path == idFieldName {
+		// and there already is and index on ID
+		if _, exists := indexes[idFieldName]; exists {
+			// just return gracefully
+			return nil
```
asutula: If we're trying to add an index on `ID` but one already exists, just return

db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
...
+		}
+	}
+
+	indexes[config.Path] = config
```
asutula: Update our in memory model with the new config

db/collection.go
```diff
@@ -93,16 +93,54 @@ func (c *Collection) Indexes() map[string]Index {
...
+
+	indexBytes, err := json.Marshal(indexes)
+	if err != nil {
+		return err
+	}
+
+	if err := c.db.datastore.Put(indexKey, indexBytes); err != nil {
```
asutula: Persist it

db/db.go
```diff
@@ -177,16 +177,21 @@ func (d *DB) reCreateCollections() error {
 			return err
 		}
 
-		var indexes []IndexConfig
+		var indexes map[string]IndexConfig
```
asutula: Use the updated storage format when re creating persisted indexes on startup

db/db.go
```diff
@@ -229,24 +234,13 @@ func (d *DB) NewCollection(config CollectionConfig) (*Collection, error) {
...
-	}
-
-	indexBytes, err := json.Marshal(config.Indexes)
-	if err != nil {
-		return nil, err
-	}
-	indexKey := dsDBIndexes.ChildString(config.Name)
-	exists, err = d.datastore.Has(indexKey)
-	if err != nil {
+	if err := c.AddIndex(IndexConfig{Path: idFieldName, Unique: true}); err != nil {
```
asutula: Be sure we add the `ID` index on `Collection` creation.

asutula: This used to exist apparently, but was lost somewhere along the way

db/db.go
```diff
@@ -229,24 +234,13 @@ func (d *DB) NewCollection(config CollectionConfig) (*Collection, error) {
...
 		return nil, err
 	}
-	if !exists {
-		if err := d.datastore.Put(indexKey, indexBytes); err != nil {
+
+	for _, cfg := range config.Indexes {
+		// @todo: Should check to make sure this is a valid field path for this schema
```
sanderpick: Did you write this todo @carsonfarmer ? Should we make a GH issue?

asutula:
One question I have:

Are there any consequences with the user overwriting an existing index and changing `Unique` from `false` to `true`, for example?

asutula:
That all kinda makes my head hurt. For sure there could be issues once we support back-indexing already existing data. But since that's not yet the case, does switching the value of `Unique` have any consequences from the time it is switched going forward?

db/db.go
carsonfarmer: Yes I did, and yes we should: https://github.com/textileio/go-threads/issues/291

carsonfarmer:
There are no knock-on effects to changing an existing index like that @asutula, it will just no longer check if a key is unique. The only place that the 'uniqueness' is persisted is in the index definition. The indexed data itself doesn't care about this directly, just the index is responsible check checking this.

carsonfarmer:
The only _consequences_ of course, are that if Unique was false previously, and then set to true, those past values that may or may not be unique will no longer be checked, only new data will be checked for uniqueness. So this won't throw, but it might be unexpected behavior. Hence why something like #286 would be good here as well.

db/collection.go
carsonfarmer: Smart

Issue #137, opened 2019-12-02
threads: add record height in log


jsign:
Add the `height` as metadata in Records. This allows to better understand if new records are already added (height below current head) and may be useful for other things like pulling with offsse/limit.

See [here](https://github.com/textileio/go-textile-threads/pull/133#discussion_r351821794)

jsign:
Consider also to include Record and Event size of the log.

requilence:
+1 for height in the Record.
Maybe consider adding the generic metadata interface for the Record? It may be helpful to add some meta information to it to be readable by the Replicator.

bitcard:
mark

PR #293, opened 2020-03-30
Thread Auth, Take 2

### LINKED ISSUE #188: JSONMode: ModelFind can take single instance
Having to supply a slice as `inerface{}` bit me once. As pointed out by @jsign, that's not very intuitive. We should be able to use reflect and create a slice from the resolved type.

sanderpick:
Currently, restarting a DB Manager requires re-using thread credentials, which is not ideal for obvious reasons. I didn't originally want to deal with this until work on the actual ACL has started, but I realized on Friday that the solution might involve API changes. Sooooo, I dug in this weekend.

Here's the issue: Starting a Manager requires subscribing to a thread and creating records with new events. If these calls require authentication, how can the manager perform them in the absence of the user? Solution:
- A bunch of component refactoring to move the thread adapter into `net` so that it can have private/non-exported access to `net` methods, like `Subscribe`, `GetThread`, etc., without needing authentication.
As a side-effect, the adapter now uses a core `DB` interface so as to avoid a bunch of invalid import cycles.
- All net and DB API methods take an auth option, including DB transactions.
- DB transactions push auth down into the adapter, allowing it to make authenticated `AddRecord` calls.

Other stuff done:
- `thread.Credentials` -> `thread.Auth`
- `thread.Auth` doesn't contain a `thread.ID`, since this is not the primary data for signing, so, the APIs have moved back to using `thread.ID`. `thread.Auth` is _always_ given as an option.
- Fixes #188 (Find can take a single dummy instance, not a slice) since I was in there anyway.

Stuff to note:
- Add collection takes auth, but since this doesn't involve writing anything to the underlying thread, there's no way to actually enforce the rules. Ideally, creating a collection writes the schema to the thread, potentially as a kind of snapshot.
- Note to self, make issue for a delete DB method on DB API (currently missing)


sanderpick:
I'll update the Android client after review so as to avoid double changes there.

jsign:
> * A bunch of component refactoring to move the thread adapter into `net` so that it can have private/non-exported access to `net` methods, like `Subscribe`, `GetThread`, etc., without needing authentication.

I see that that's convenient for solving the problem, but this makes the `net` package know about _DB_, which seems to be a dependency in the wrong direction. 

The _ThreadAdapter_ is something only meaningful for _DB_, so it seems to me that it shouldn't live in `net`. This means _DB_ use `net` _internals_ to work, not the public API.

I think to_accept this as a definitive change, we should say that the _DB_ concepts are part of _threads_. We can't think of _DB_ as an app using _threads_ anymore.

I'm not saying that we should do something now to move on; just exposing this point.

> * `thread.Auth` doesn't contain a `thread.ID`, since this is not the primary data for signing, so, the APIs have moved back to using `thread.ID`. `thread.Auth` is _always_ given as an option.

Like that change, I found that sometimes methods receive _Credentials_ only for knowing the _ThreadID_.

db/db.go
```diff
@@ -152,9 +153,9 @@ func newDB(n net.Net, creds thread.Credentials, config *Config) (*DB, error) {
 		}
 	}
 
-	adapter := newSingleThreadAdapter(d, n, creds)
-	d.adapter = adapter
-	adapter.Start()
+	connector := n.ConnectDB(d, id)
+	d.connector = connector
+	connector.Start()
```
jsign: Thought: Does `Start()` method stil makes sense now that things are auto-started?

db/db.go
```diff
@@ -294,10 +295,10 @@ func (d *DB) Close() error {
 	}
 	d.closed = true
 
-	if d.adapter != nil {
-		d.adapter.Close()
+	if d.connector != nil {
+		_ = d.connector.Close()
```
jsign: Nit: This was the case before this PR, but we could log the possible error? Maybe the right thing is returning it after trying to close things later, I'm not sure. But an error closing the `connector` sounds like an interesting thing to know about.

net/dbconnector.go
```diff
@@ -0,0 +1,140 @@
+package net
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	coredb "github.com/textileio/go-threads/core/db"
```
jsign: Will make a point about something related to `net` knowing about `db` concepts later.

core/thread/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	a.msg = msg
+	if a.sk != nil {
+		sig, err := a.sk.Sign(msg)
+		if err != nil {
+			return nil, nil, err
+		}
+		return sig, a.sk.GetPublic(), nil
+	}
+	return a.sig, a.pk, nil
+}
```
jsign: If `Auth` was created with `NewAuthFromSignature`, then `msg` is ignored.
That seems quite weird, and a bit dangerous... a `Sign(msg)` method that ignores what is being signed, but returns a signature anyway.

jsign:
A lateral comment:

> * A bunch of component refactoring to move the thread adapter into `net` so that it can have private/non-exported access to `net` methods, like `Subscribe`, `GetThread`, etc., without needing authentication.

I see that that's convenient for solving the problem, but this makes the net package know about DB, which seems to be a dependency in the wrong direction.

The ThreadAdapter is something only meaningful for DB, so it seems to me that it shouldn't live in net. This means DB use net internals to work, not the public API.

I think to_accept this as a definitive change, we should say that the DB concepts are part of threads. We can't think of DB as an app using threads anymore.

If this situation happened with DB, sounds it will prob happen with other apps that use threads. If DB need to use private/internals of net, is that an indication that the public API is missing something? (I've not dived into this topic, just thought by analogy so might be wrong).

I'm not saying that we should do something now to move on; just exposing this point to get some opinions!

_PD: For some weird reason I had to repost this since the original disappeared?, luckily I had the tab with it cached_

core/thread/auth.go
sanderpick: This should make sense when you see it in action... the API does this in order to set the message to be validated. I guess it could be an additional API.

sanderpick:
> I see that that's convenient for solving the problem, but this makes the net package know about DB, which seems to be a dependency in the wrong direction.
> 
> The ThreadAdapter is something only meaningful for DB, so it seems to me that it shouldn't live in net. This means DB use net internals to work, not the public API.
> 
> I think to_accept this as a definitive change, we should say that the DB concepts are part of threads. We can't think of DB as an app using threads anymore.
> 
> If this situation happened with DB, sounds it will prob happen with other apps that use threads. If DB need to use private/internals of net, is that an indication that the public API is missing something? (I've not dived into this topic, just thought by analogy so might be wrong).

Well, I think it's just a matter of perspective. With this change, you can still use net by itself, and you could connect another app by implementing the connector. The idea is that the connector provides the missing public API components. 

To further seperate things, I could move the `core/db/Net` interface into it's own package, maybe called `connector`. That might scratch your itch?

jsign:
Oh, so I think there's only a naming _problem_... if instead of `db` names in `net/dbconnector.go` maybe a change to `app` sounds good? 

Looking at it in that way, seems like a generic-api for any app not just for db. Like: constructor`AppConnector`, `appToThread`, `threadToApp`, log messages, etc. 

Not sure it's the right name, but I think makes much clearer your intention. Looks good.

sanderpick:
Sounds great! Change coming.

core/thread/auth.go
jsign: Uhm, yeah. I'd say a different API to do that sounds better, since in that case, it isn't signing anything.

I still have an itch on this logic of having to save a signature, which doesn't correspond always with the corresponding message. Still, something doesn't tick for me.

sanderpick: I don't love it either... the problem is that I'm trying use the same object for,
- A user that has a private key they want kept private, but they don't know how to sign the message since the message creation is done internally... hence the Sign method being called from within the client.
- The API that receives the signed message and pubkey, and then must also construct a `thread.Auth` message for use with the _internal_ thread net API... hence the need to set the message.

I think ultimately, the message being signed might have to change from being a protobuf message to being some custom combination of the method arguments for each API call. In this way, the server could just pass through the "auth" without have to set the message.

sanderpick: Hmm, yeah I think that's better... more tedious, but it would avoid this problem. And another problem that is more subtle... a pure-Go app that doesn't use the gRPC APIs could totally cheat the current setup.

jsign: I agree that using the same struct for both things is what's causing the problem.
For one use case, it seems that the API should allow receiving some `SignedMessage` (which may be a plain struct without methods). And send that to the internal API to deal with that.

Forcing to use the same struct, makes this _fake signings_.

jsign: That might be an indication that the API should always receive `SignedMessages` and let signing be a problem outside the API. If comes through gRPC, was done by the client. If not, in some other place.

_sharing thoughts, not sure covers all the cases_

sanderpick: Yep, if the non-gRPC clients _know how to sign_, then we can do that... that mean the message to be signed is some combo of the method args, not the nicely signable proto messages.

api/service.go
```diff
@@ -601,6 +694,19 @@ func (s *Service) getCollection(creds thread.Credentials, collectionName string)
 	return collection, nil
 }
 
-func getCredentials(c *pb.Credentials) (thread.Credentials, error) {
-	return thread.NewSignedCredsFromBytes(c.ThreadID, c.PubKey, c.Signature)
+func getAuth(header *pb.Header, body proto.Message) (*thread.Auth, error) {
```
asutula: Yea, curious how this would look as request metadata

db/collection.go
```diff
@@ -148,66 +149,100 @@ func (c *Collection) AddIndex(config IndexConfig) error {
...
-func (c *Collection) Save(vs ...[]byte) error {
+// Save saves changes of an instance in the collection.
+func (c *Collection) Save(v []byte, opts ...TxnOption) error {
+	return c.WriteTxn(func(txn *Txn) error {
+		return txn.Save(v)
+	}, opts...)
+}
+
+// SaveMany saves changes of multiple instances in the collection.
+func (c *Collection) SaveMany(vs [][]byte, opts ...TxnOption) error {
```
asutula: This doesn't seem worth it to me to have these different flavors of create and save for single and many. Can we just make the auth an explicit param and not use options and stick with varadic args?

db/manager.go
```diff
@@ -97,19 +90,19 @@ func NewManager(network net.Net, opts ...Option) (*Manager, error) {
 }
 
 // NewDB creates a new db and prefixes its datastore with base key.
-func (m *Manager) NewDB(ctx context.Context, creds thread.Credentials, collections ...CollectionConfig) (*DB, error) {
-	id := creds.ThreadID()
+func (m *Manager) NewDB(ctx context.Context, id thread.ID, opts ...NewManagedDBOption) (*DB, error) {
```
asutula: I'd actually love to see the `id thread.ID` param moved to options. If you specify one as an option it's used, otherwise we'll generate a thread id for you.

db/manager.go
```diff
@@ -120,26 +113,30 @@ func (m *Manager) NewDB(ctx context.Context, creds thread.Credentials, collectio
 // NewDBFromAddr creates a new db from address and prefixes its datastore with base key.
 // Unlike NewDB, this method takes a list of collections added to the original db that
 // should also be added to this host.
-func (m *Manager) NewDBFromAddr(ctx context.Context, creds thread.Credentials, addr ma.Multiaddr, key thread.Key, collections ...CollectionConfig) (*DB, error) {
-	id := creds.ThreadID()
+func (m *Manager) NewDBFromAddr(ctx context.Context, addr ma.Multiaddr, key thread.Key, opts ...NewManagedDBOption) (*DB, error) {
```
asutula: Nice, this is much cleaner without having to deal with credentials and the thread id (that sent me for a loop when integrating credentials into folder sync)

api/client/client.go
```diff
@@ -16,60 +17,63 @@ import (
...
-// Client provides the client api
+// Client provides the client api.
 type Client struct {
 	c    pb.APIClient
 	conn *grpc.ClientConn
 }
 
-// NewClient starts the client
+// Instances is a list of collection instances.
+type Instances []interface{}
```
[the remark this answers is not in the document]
sanderpick: It is optional though. Some threads will not require any auth, e.g., the ACL is wide open.

db/collection.go
sanderpick: Is there a downside other than just having another method? Mongo does this and it feels pretty intuitive as a user IMO.

db/manager.go
sanderpick: I was thinking the same, but on the client side (for this DB API) once we have client-side ACL builders that involve packing the ACL hash into the ID.

api/service.go
sanderpick: Nice, good call... I totally forgot about metadata amidst all the other changes.

db/db.go
sanderpick: Yep, I'll just return the error.

db/db.go
sanderpick: Nope, removing.

api/client/client.go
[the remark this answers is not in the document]
asutula: Ok sounds good

api/service.go
asutula: 👍

db/collection.go
asutula: Ok, I can live with that

db/manager.go
asutula: ok. maybe we can explore client side sugar for this in the future

sanderpick: Totally

core/app/app.go
```diff
@@ -0,0 +1,213 @@
+package app
```
sanderpick: All the app and connector stuff now lives here. This "app" abstraction feels great.

core/app/app.go
```diff
@@ -0,0 +1,213 @@
...
+func (l *LocalEventListener) Discard() {
+	l.listener.Discard()
+}
+
+// Net adds the ability to connect an app to a thread.
+type Net interface {
+	net.Net
+
+	// ConnectApp returns an app<->thread connector.
+	ConnectApp(App, thread.ID) (*Connector, error)
```
sanderpick: The net layer simply implements this one method. I'll highlight it below.

core/app/app.go
```diff
@@ -0,0 +1,213 @@
...
+var log = logging.Logger("app")
+
+const (
+	busTimeout        = time.Second * 10
+	addRecordTimeout  = time.Second * 10
+	fetchEventTimeout = time.Second * 15
+)
+
+// App provides a bidirectional hook for thread-based apps.
+type App interface {
```
sanderpick: An app implements these two methods. Basically, it must provide a hook for threads to handle new events, and it must itself handle new records.

core/app/app.go
```diff
@@ -0,0 +1,213 @@
...
+
+	lock   sync.Mutex
+	closed bool
+}
+
+// Connection receives new thread records, which are pumped to the app.
+type Connection func(context.Context, map[thread.ID]struct{}) (<-chan net.ThreadRecord, error)
+
+// NewConnector creates bidirectional connection between an app and a thread.
+func NewConnector(app App, net Net, tinfo thread.Info, conn Connection) (*Connector, error) {
```
sanderpick: The app calls `ConnectApp`, which must return a connector. The net layer implements this func, which uses this constructor to create the connection. It simply needs to return the thread that's being connected, and a connection func that is used to notify the app of new records.

core/app/app.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	a := &Connector{
+		app:       app,
+		net:       net,
+		threadID:  tinfo.ID,
+		threadKey: tinfo.Key,
+		logID:     lg.ID,
+		closeChan: make(chan struct{}),
+	}
+	var wg sync.WaitGroup
```
sanderpick: We can just start the connector, no need to call the old adapter start method.

core/app/app.go
```diff
@@ -0,0 +1,213 @@
...
+		select {
+		case <-c.closeChan:
+			log.Debugf("closing thread-to-app flow on thread %s", c.threadID)
+			return
+		case rec, ok := <-sub:
+			if !ok {
+				log.Debug("notification channel closed, not listening to external changes anymore")
+				return
+			}
+			if err = c.app.HandleNetRecord(rec, c.threadKey, c.logID, fetchEventTimeout); err != nil {
```
sanderpick: This method now takes a thread key, which is now baked into the connection. There's no longer a need to fetch the thread here.

db/db.go
```diff
@@ -152,9 +155,11 @@ func newDB(n net.Net, creds thread.Credentials, config *Config) (*DB, error) {
 		}
 	}
 
-	adapter := newSingleThreadAdapter(d, n, creds)
-	d.adapter = adapter
-	adapter.Start()
+	connector, err := n.ConnectApp(d, id)
```
sanderpick: Connect the db to the net layer.

net/net.go
```diff
@@ -622,21 +628,29 @@ func (t *net) Subscribe(ctx context.Context, opts ...core.SubOption) (<-chan cor
 	return channel, nil
 }
 
+func (n *net) ConnectApp(a app.App, threadID thread.ID) (*app.Connector, error) {
```
sanderpick: Here's where net implements app.Net. We just need to fetch the thread (can do w/o auth here), and return a connection.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
+package thread
```
sanderpick: Alrighty then, take a looks at the changes here to `thread.Auth`.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/gogo/status"
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"google.golang.org/grpc/codes"
+)
+
+type Identity interface {
```
sanderpick: @carsonfarmer This provides the ability to use a generic "identity" that can sign. Most of the time this would just be a private key, which implements this interface.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	Sign([]byte) ([]byte, error)
+	GetPublic() crypto.PubKey
+}
+
+type Auth struct {
+	pk  crypto.PubKey
+	sig []byte
+}
+
+type Method byte
```
sanderpick: `Method` provides a tight scoping of the message being signed, since some API methods take the same args.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	AddRecord            = 0x08
+	GetRecord            = 0x09
+	Subscribe            = 0x0A
+)
+
+func NewAuth(id Identity, method Method, args ...encoding.BinaryMarshaler) (auth Auth, err error) {
+	if id == nil {
+		return
+	}
+	msg, err := marshalBinary(method, args)
```
sanderpick: This rolls up the args for the given method into a signable message.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"google.golang.org/grpc/codes"
+)
+
+type Identity interface {
+	Sign([]byte) ([]byte, error)
+	GetPublic() crypto.PubKey
+}
+
+type Auth struct {
```
sanderpick: Auth simply wraps a pub key and signature.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return
+	}
+	return Auth{
+		pk:  id.GetPublic(),
+		sig: sig,
+	}, nil
+}
+
+func NewAuthFromMD(ctx context.Context) (auth Auth, err error) {
```
sanderpick: On the server-side, we can create an Auth object from request metadata.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	if a.pk == nil {
+		return ""
+	}
+	pkb, err := crypto.MarshalPublicKey(a.pk)
+	if err != nil {
+		panic(err)
+	}
+	keystr := base64.StdEncoding.EncodeToString(pkb)
+	sigstr := base64.StdEncoding.EncodeToString(a.sig)
+	return fmt.Sprintf("Signature keyId=%s,signature=%s", keystr, sigstr)
```
sanderpick: This spits out the authorization string used in the metadata, which is an HTTP header. I'm not sure it really matters, but here I'm using the standard: https://tools.ietf.org/html/draft-cavage-http-signatures-11#section-2.1 for including a signature for authorization. You can see the keyId is not exactly that same as just passing the key, which I'm doing. Part of me wants to use something close to a standard, rather then make up something totally new, like a custom header. Opinions welcome.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (a Auth) Verify(method Method, args ...encoding.BinaryMarshaler) (crypto.PubKey, error) {
+	if a.pk == nil {
+		return nil, nil
+	}
+	msg, err := marshalBinary(method, args)
```
sanderpick: On the verification end, we can create the exact same message _from any of the APIs_... not just the gRPC ones since the message does not involve protos.

core/thread/auth.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, err
+		}
+		bytes = append(bytes, d...)
+	}
+	return bytes, nil
+}
+
+type ctxKey string
+
+func NewAuthContext(ctx context.Context, auth Auth) context.Context {
```
sanderpick: Below here is just gRPC metadata handling.

net/api/client/client.go
```diff
@@ -59,153 +56,197 @@ func (c *Client) GetHostID(ctx context.Context) (peer.ID, error) {
...
+func (c *Client) CreateThread(ctx context.Context, id thread.ID, opts ...core.NewThreadOption) (info thread.Info, err error) {
+	args := &core.NewThreadOptions{}
 	for _, opt := range opts {
 		opt(args)
 	}
 	keys, err := getThreadKeys(args)
 	if err != nil {
 		return
 	}
+	auth, err := thread.NewAuth(args.Identity, thread.CreateThread, id, args)
```
sanderpick: `NewAuth` at work.

net/api/client/client_test.go
```diff
@@ -210,8 +210,7 @@ func TestClient_AddRecord(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	creds := thread.NewDefaultCreds(id, thread.WithPrivKey(authorSk))
-	_, err = client.CreateThread(context.Background(), creds, core.WithThreadKey(tk), core.WithLogKey(logPk))
+	_, err = client.CreateThread(context.Background(), id, core.WithThreadKey(tk), core.WithLogKey(logPk), core.WithNewThreadIdentity(authorSk))
```
sanderpick: Here's how a client would use their own identity provider with a request. If they aren't comfortable trusting the client code to not suck up their private key, they could implements a wrapper that also implements `Identity`, and keep private key private. In any case, private key is sent anywhere.

net/api/service.go
```diff
@@ -61,15 +61,20 @@ func (s *Service) GetHostID(_ context.Context, _ *pb.GetHostIDRequest) (*pb.GetH
...
 	if err != nil {
-		return nil, err
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 	opts, err := getKeyOptions(req.Keys)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
-	info, err := s.net.CreateThread(ctx, creds, opts...)
+	auth, err := thread.NewAuthFromMD(ctx)
```
sanderpick: Here we can pull out the auth from the metadata and simply pass to the internal API.

net/net.go
```diff
@@ -127,47 +128,57 @@ func (t *net) Close() (err error) {
...
+	identity, err := args.Auth.Verify(thread.CreateThread, id, args)
+	if err != nil {
 		return
 	}
+	if identity != nil {
+		iid, err := peer.IDFromPublicKey(identity)
+		if err != nil {
+			return info, err
+		}
+		log.Debugf("👋 %s", iid)
```
sanderpick: Finally, we can wave hello to the caller.

sanderpick: (This would be checked against the thread ACL)

sanderpick:
Okay, I've updated `net/api` with the new auth handling:
- Different signing method
- Using gRPC metadata instead of custom pb objects for authorization

I'll leave it here until we can reach consensus on the methods, then I can update `api` with the same logic.

net/net.go
jsign: Liked my proposed _App_ name, seems to fit quite well for a first idea :P

carsonfarmer:
I'm certainly a fan of the generic Sign and Verify interface here, nice to see that. I do have the same question #1 that @jsign has here about _what_ we are signing. Recognizing that we're trying to avoid signing protos (thank you!), is there any alternative here that might be more 'standard'. @jsign I'm interested in hearing your thoughts here in particular?

As for @jsign's second question, I'm not sure I understand what you are asking here? Is the issue that, since the client can have an arbitrary key type for signing, there's no way for the server to _know_ how to verify? Since the client is sending along the public key, the server should be able to determine how to verify with/from that, no?

jsign:
> I do have the same question # 1 that @jsign has here about _what_ we are signing. Recognizing that we're trying to avoid signing protos (thank you!), is there any alternative here that might be more 'standard'. @jsign I'm interested in hearing your thoughts here in particular?

Well, my last point wasn't about what's being signed. From what I understood is a _metod+args_ custom encoding of the _intention_. My point was chatting about how that scheme matters when making changes in the API and compatibility.

I'm not sure why signing protos was discarded... but even in that case, I think I'd make the same points of how changing protos might impact client and server compatibility. I remember my main point in the last review round was more about some strange behavior of `Sign`... but a lot happened in this PR, so maybe missed something. 
 
> As for @jsign's second question, I'm not sure I understand what you are asking here? Is the issue that, since the client can have an arbitrary key type for signing, there's no way for the server to _know_ how to verify? Since the client is sending along the public key, the server should be able to determine how to verify with/from that, no?

What I mean is that the verification is a public signature (defined). It can't be a _generic verify_ function. I'm OK with that since I don't an idea how that can be different... but I'm sure you mentioned something of that? The `Identity` interface doesn't have a `Verify` method. (And I repeat, I can't imagine how that could be without introducing more complex ideas).
I'm just mentioning this since I had the _sensation_ you were interesting in something of this... if that's isn't the case, just dismiss this. :P

carsonfarmer:
Just to add on to my previous comment. I hate to do a long back and forth here, but this is a pretty critical interface that we need to get right early on. I'd much rather us argue a bit and come to a nice consensus as @sanderpick has suggested, than just move along with something that is _almost_ right (keeping in mind of course, that it might _be_ right already!).

carsonfarmer:
As for avoiding using the protos directly. This is primarily because making clients deal with proto objects is not a great DX tbh. The protos are an over the wire thing, but we never expose these in JS, Java, etc. (or at least, we really shouldn't). So having to sign the _payload_ directly isn't ideal in that case. Having said that, it _is_ possible to abstract this away from the callers, so if that makes things easier from an API point of view, let's think about that?

carsonfarmer:
Side note: I hate to go back and forth on this a ton, but it _is_ kind of a critical interface, particularly for external integrations, so I think the extra discussion overhead now is worth it.

sanderpick:
I agree, the back and forth is good. I'm glad to see the input for these complex and user facing design decisions. 

We can't use the protos for signing because there are a number of entry points to the internal net API that are not from the gRPC service. So, in that case, what would be the message the signature against? 

This message should in some way describe the intention of the user. This is easy with network-based APIs like HTTP / gRPC since you can just use the already binary marshaled network request. But what to do for a normal API, like `net/net`? That's why I landed on the `thread.Method` + args approach. I really can't think of another way forward, but I certainly may be missing an obvious solution.

api/service.go
```diff
@@ -66,11 +68,15 @@ func (s *Service) Close() error {
 func (s *Service) NewDB(ctx context.Context, req *pb.NewDBRequest) (*pb.NewDBReply, error) {
 	log.Debugf("received new db request")
 
-	creds, err := getCredentials(req.Credentials)
+	id, err := thread.Cast(req.Body.DbID)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	auth, err := getAuth(req.Header, req.Body)
```
asutula: Thought I read your comments about moving to metadata, but I guess not yet?

sanderpick:
Another point that is at play here: The ACL stuff is _part of the thread protocol_. So, the verification must happen inside `net/net`, not at the gRPC level. This may sound arbitrary, but what would be the meaning of an ACL that's baked into a `thread.ID` if the protocol can't enforce the rules on it's own? It shouldn't need a surrounding wrapper for that, right?

carsonfarmer:
Very good points @sanderpick, and this helps a lot in understanding your approach! On that front, and to address @jsign's concern + suggestion, what about marshalling that thread.Method + Args approach with a semver prefix?

sanderpick:
I agree @jsign, the `Identity` owner should not be able to verify anything without knowing the input... seems that would defeat the purpose.

sanderpick:
I like it.. we don't have any version stuff in here at all, so I guess now would be a good time to add it.

Adding the arg lengths also sounds good.

db/collection.go
```diff
@@ -148,66 +149,100 @@ func (c *Collection) AddIndex(config IndexConfig) error {
...
 		return err
-	})
+	}, opts...)
 	return
 }
 
-// Create creates instances in the collection.
-func (c *Collection) Create(vs ...[]byte) (ids []core.InstanceID, err error) {
+// Create creates an instance in the collection.
+func (c *Collection) Create(v []byte, opts ...TxnOption) (ids []core.InstanceID, err error) {
```
asutula: This should just return a single InstanceID

db/db.go
asutula: very nice

api/service.go
sanderpick: You can see the approach in the `net/api`. I wanted to get some confirmation on my signing approach before making changes to `api`. Plus, I ran out of time before bed.

db/collection.go
sanderpick: Good catch 👍

jsign:
To be clear, most of my points above are descriptions of the current solution and some things to share and be mindful about... like the versioning thing. That's not an argument to look for a better solution.

I think also that unpacking an auth-scheme idea from a code review rather than a spec it's a bit hard, at least for me, and be sure to be understood correctly. Not to say if there's some back and forth.

Without having the ACL implemented, it's hard to know if this solution is the best or not or if there's a more simple one. For example, the `verifyRequest` is just checking the signature is valid for the public-key. But I guess there's a missing part where that public-key is checked to be an allowed one? Sounds to me that will land in the ACL part. I guess there're other characteristics that might not make sense now, but maybe later.

jsign:
An extra point in sharing ideas:
Request replays. We should cover cases where replaying a request might be undesirable or dangerous? Or our methods are replay-safe? 
Not sure about the answers, just sharing a flashing thought.

sanderpick:
I considered adding an expiry to each request to limit the reply attack, but then, based on my research last week, a replay attack is not possible with SSL. So, seems that we can just rely on that. Otherwise, we'd have to cover a bunch of other vectors also covered by SSL.

jsign:
> I considered adding an expiry to each request to limit the reply attack, but then, based on my research last week, a replay attack is not possible with SSL. So, seems that we can just rely on that. Otherwise, we'd have to cover a bunch of other vectors also covered by SSL.

But if I take the raw bytes of the request  (after the SSL unwrapping; the proto), and send those bytes a new request to other peer, that check only relies on our logic. Or maybe I'm confused?

sanderpick:
See https://security.stackexchange.com/a/20106.

> The encryption provided by SSL/TLS certainly prevents an eavesdropper from seeing that application request, and thus from replaying it with their own separate SSL/TLS connection.

> However, SSL/TLS on its own doesn't necessarily prevent the legitimate initial user from replaying a request. Protocols and applications that require this additional level of protection tend to have nonce-based mechanisms at the application level to address this problem (which is rather independent of SSL/TLS, although preventing eavesdropping helps).

This caveat is important to note, but not really a concern for us as far as I can tell. If a legit user wants to replay their own request, go for it.

jsign:
Yep, I described just that case above.
So if we don't have signed requests shared by different peers, that shouldn't be a problem.

sanderpick:
> But if I take the raw bytes of the request (after the SSL unwrapping; the proto), and send those bytes a new request to other peer, that check only relies on our logic. Or maybe I'm confused?

That's true, but how would you get those bytes? Are you referring to if the service itself send the same request to another service?

sanderpick:
Ah okay, I see... comment delay!

jsign:
> That's true, but how would you get those bytes? Are you referring to if the service itself send the same request to another service?

Yeah, just mentioning if we have some case in which a signed message is sent to another peer.
I don't have this PR so fresh in my mind, so if that isn't the case... nothing to worry about. But thinking about the ACL stuff, I thought we might have those cases.

sanderpick:
I should have shared notes on the ACL thinking. I'll try to summarize that...

For each new thread, a JSON ACL doc can be built by the client lib:

```
{
  "$w": ["a", "b"], // these keys can add records
  "$r": ["*"], // these keys can get thread, get record, pull thread, and subscribe
  "$d": ["a"], // these keys can delete the thread
  "$s": ["*"], // these keys can add services like a replicator
}
```

`*` is anyone.

- The hash of the ACL can be derived from the static `thread.ID`. This means that the ACL and it's rules can be independently verified  / enforced by anyone who has the ACL and `thread.ID`.
- Since threads does not concern itself with pinning, I've been thinking that these ACL docs can just be sent allow with push log requests. This would ensure that everyone has the ACL who needs it.
- In order to enforce the rules, a peer needs _some proof_ of the caller's identity. Does the caller control the associated private key? If so, check the rules... can they write? read? etc. Is the rule a * rule? If so, it doesn't matter.

**This PR is enabling the last piece.**

sanderpick:
Issue https://github.com/textileio/go-threads/issues/295 created referencing the ACL implementation.

api/service.go
asutula: Cool, I like what I saw in net

jsign:
Cool, that would have helped a bit from the start. :P Good to know that I landed to the same conclusion by the code+discussion.

It seems that the fundamental requirement that explodes into this signing/verifying thing, is wanting to have arbitrary public keys in the ACL definition. In particular, not being the public/private keys used by the libp2p peer. If that was the case, we can rely on the libp2p check of identity that is done on the connection handshake.

Is there some ground to explore about using PeerID in the ACL, or similar? Or maybe manually define the PeerID of the peer to the same key? (if that's possible) Or maybe we want the original owner to take actions in any libp2p peer which doesn't allow to rely on PeerID?

Just making this question since that solution sounds to good to be true. But want to confirm that that was intentional.

sanderpick:
> Is there some ground to explore about using PeerID in the ACL, or similar? Or maybe manually define the PeerID of the peer to the same key? (if that's possible) Or maybe we want the original owner to take actions in any libp2p peer which doesn't allow to rely on PeerID?

Yep, the use-cases for an identity are pretty broad, basically any PKI. So, eventually we have to reach beyond `go-libp2p-core/crypto` keys, but for now I'm just using the verify method from that lib.

For example, fleek.co want to use keybase keys. I believe those are https://nacl.cr.yp.to/ type. In order to support verifying those, we could include the `algorithm` type in the HTTP authorization header, as mentioned here: https://tools.ietf.org/html/draft-cavage-http-signatures-11.

carsonfarmer:
Just another mention of the JWS standard for including algorithm information in signed messages. I know the spec specifies using base64 encoding strings for the signed bytes, which is a bummer. But at least there are plenty of tools (across various languages) to handle these.

sanderpick:
Okay guys, this is getting a little insane, kicked off by the fact the the above approach does not work at the DB layer*, I've come up with _another_ (👨 🔫 ) approach that is actually a bit closer to my original signing style of just using the `thread.ID` bytes as the identity "challenge". 

*Requests often involve more then one thread operation. 

First, taking a step back, amidst all the related log verification stuff in `net` (the p2p API)** and all the feedback here, I forgot what the original intention was. 

**To summarize the _related but different work here_: The verification of peers is needed because I don't know how to access the underlying the libp2p `peer.ID` that's masked by the higher level `net.Conn` (we can probably figure that out at some point). This is why we sign the protos at this level, which _is not the same_ as the record verification, which needs to be possible when the records come from the non-author, hence the new `AuthorSig` in `core.Record`. In other words, the peer host is not always the record author.

So, that's _two different_ signing related things going on here, both of which are all about normal message verification. However, what's needed for ACL checks is not message verification, it's proof of identity, i.e., "Does this entity control the inferred private key?". That's the question we're trying to answer here. 

Back to the original idea of just signing `thread.ID`. In fact, this approach is sufficient to answer this question (I blame myself for getting derailed from this goal). Whether or not this proof is replayed to the same method call (user intention) or even a different one does not matter. But, what we're doing here is effectively forcing the user to _login_ for every request, which is cumbersome and not needed.

The solution here is super common and it's called a user session / token :) The user token can be used with different requests. In the case of the thread API, the token "claims" are already backed into the thread (the ACL), so it doesn't even have to carry around that information. 

The basic idea is that the peer host can issue JWTs just like a normal server (not the other way around for encoding message verifications, which was suggested above). In order for this to work, the user must complete a signing challenge (much like `go-textile`). After successful completion, they get a JWT token scoped to their public key. 

With gRPC bidi steams, it's even simpler than what we used to do, since the stream is _secure_, we don't have to deal with client or server nonces (as far as I can tell, please tell me otherwise).

I'll make some comments in a bit to help the (hopefully last big) review.

net/net.go
```diff
@@ -305,18 +349,21 @@ func (t *net) pullThread(ctx context.Context, id thread.ID) error {
...
 	for _, lg := range info.Logs {
-		has, err := t.bstore.Has(lg.Head)
-		if err != nil {
-			return err
+		var has bool
+		if lg.Head.Defined() {
+			has, err = n.bstore.Has(lg.Head)
+			if err != nil {
+				return err
+			}
```
sanderpick, line 366: @asutula here's the bug that causes the bloom filter error.

api/service.go
```diff
@@ -47,60 +50,128 @@ func NewService(network net.Net, conf Config) (*Service, error) {
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	collections := make([]db.CollectionConfig, len(req.Collections))
+	for i, c := range req.Collections {
+		cc, err := collectionConfigFromPb(c)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		collections[i] = cc
+	}
```
sanderpick, line 140: Collecting the collections... noticed they were missing for `NewDB`.

core/net/net.go
```diff
@@ -30,38 +30,41 @@ type API interface {
 	// GetHostID returns the host's peer id.
 	GetHostID(ctx context.Context) (peer.ID, error)
 
+	// GetToken returns a signed token representing an identity.
+	GetToken(ctx context.Context, identity thread.Identity) (thread.Token, error)
```
sanderpick, line 34: This is the "login" method that returns a stateless token (JWT) representing an "identity".

core/net/options.go
```diff
@@ -25,24 +26,54 @@ func WithThreadKey(key thread.Key) KeyOption {
...
-func WithLogKey(key crypto.Key) KeyOption {
-	return func(args *KeyOptions) {
+func WithLogKey(key crypto.Key) NewThreadOption {
+	return func(args *NewThreadOptions) {
 		args.LogKey = key
 	}
 }
 
+// WithNewThreadToken provides authorization for creating a new thread.
+func WithNewThreadToken(t thread.Token) NewThreadOption {
```
sanderpick, line 36: All these options get much more simplified now. The API methods only ever need to take the token string.

core/thread/token.go
```diff
@@ -0,0 +1,189 @@
+package thread
```
sanderpick, line 1: This package contains the token handling.

core/thread/token.go
```diff
@@ -0,0 +1,189 @@
...
+func NewLibp2pIdentity(key crypto.PrivKey) Identity {
+	return &Libp2pIdentity{PrivKey: key}
+}
+
+func (p *Libp2pIdentity) GetPublic() PubKey {
+	return NewLibp2pPubKey(p.PrivKey.GetPublic())
+}
+
+// Pubkey can be anything that provides a verify method.
+type PubKey interface {
```
sanderpick, line 43: This is meant to allow for all kinds of pub keys, with an implementation for libp2p pubkeys below. This also has built in string methods that should encode / decode with base64 for use in a JWT.

core/thread/token.go
```diff
@@ -0,0 +1,189 @@
...
+	"google.golang.org/grpc/codes"
+)
+
+// Identity represents an entity capable of signing a message
+// and returning the associated public key for verification.
+// In many cases, this will just be a private key, but callers
+// can use any setup that suits their needs.
+type Identity interface {
+	Sign([]byte) ([]byte, error)
+	GetPublic() PubKey
```
sanderpick, line 25: Identity pubkey can be of the more generic type.

core/thread/token.go
```diff
@@ -0,0 +1,189 @@
...
+func NewToken(issuer crypto.PrivKey, key PubKey) (tok Token, err error) {
+	var ok bool
+	issuer, ok = issuer.(*crypto.Ed25519PrivateKey)
+	if !ok {
+		log.Fatal("issuer must be an Ed25519PrivateKey")
+	}
+	claims := jwt.StandardClaims{
+		Subject:  key.String(),
+		Issuer:   NewLibp2pIdentity(issuer).GetPublic().String(),
+		IssuedAt: time.Now().Unix(),
```
sanderpick, line 99: We could add an expiry here in the future to help limit the downside of a stolen token.

core/thread/token.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		if tok == nil {
+			return nil, ErrTokenNotFound
+		} else {
+			return nil, ErrInvalidToken
+		}
+	}
+
+	key := &Libp2pPubKey{}
+	if err = key.UnmarshalString(claims.Subject); err != nil {
```
sanderpick, line 134: The subject of the token is the user's pubkey. Here's how to pull it out.

db/manager.go
```diff
@@ -96,20 +89,25 @@ func NewManager(network net.Net, opts ...Option) (*Manager, error) {
...
 		return nil, fmt.Errorf("db %s already exists", id)
 	}
-	if _, err := m.network.CreateThread(ctx, creds); err != nil {
-		return nil, err
+	args := &NewManagedDBOptions{}
+	for _, opt := range opts {
+		opt(args)
 	}
-	if err := m.putDBCreds(id, creds); err != nil {
+	if _, err := m.network.CreateThread(ctx, id, net.WithNewThreadToken(args.Token)); err != nil {
```
sanderpick, line 106: The tokens can easily be passed through from up in the DB layer.

db/manager.go
```diff
@@ -96,20 +89,25 @@ func NewManager(network net.Net, opts ...Option) (*Manager, error) {
 	return m, nil
 }
 
+// GetToken provides access to thread network tokens.
+func (m *Manager) GetToken(ctx context.Context, identity thread.Identity) (thread.Token, error) {
```
sanderpick, line 93: Manager provides it's own avenue to getting a token, which just calls the method on the network layer. Note: the normal `db.go` APIs can also take a token. In these cases, the user would own the network layer, and they can call `GetToken` there directly.

jwt/jwt.go
```diff
@@ -0,0 +1,75 @@
+package jwted25519
```
sanderpick, line 1: This pulls in the old JWT ed25519 method I wrote for `go-textile`.

net/net.go
```diff
@@ -127,48 +131,71 @@ func (t *net) Close() (err error) {
...
-func (t *net) GetHostID(_ context.Context) (peer.ID, error) {
-	return t.host.ID(), nil
+func (n *net) GetHostID(_ context.Context) (peer.ID, error) {
+	return n.host.ID(), nil
 }
 
-func (t *net) CreateThread(_ context.Context, creds thread.Credentials, opts ...core.KeyOption) (info thread.Info, err error) {
-	id := creds.ThreadID()
-	if err = t.ensureUnique(id); err != nil {
+func (n *net) GetToken(_ context.Context, identity thread.Identity) (tok thread.Token, err error) {
```
sanderpick, line 166: Here's the start of the interesting bits. This method takes an _identity_ interface which allows it to sign a random message payload, or challenge. You might be thinking, but shouldn't the signing happen on the client? It does, because this is just exposing an opaque request for a signature. We can't actually get private key here.

net/net.go
```diff
@@ -127,48 +131,71 @@ func (t *net) Close() (err error) {
...
+	msg := make([]byte, tokenChallengeBytes)
+	if _, err = rand.Read(msg); err != nil {
+		return
+	}
+	sig, err := identity.Sign(msg)
+	if err != nil {
 		return
 	}
+	key := identity.GetPublic()
+	if ok, err := key.Verify(msg, sig); !ok || err != nil {
```
sanderpick, line 176: Again, it might look weird to sign and verify right away, but remember `Identity` above is not the private key, just a method to ask for a signature. This will be more obvious in the gRPC service below.

net/net.go
```diff
@@ -127,48 +131,71 @@ func (t *net) Close() (err error) {
...
+		return tok, fmt.Errorf("bad signature")
+	}
+	return thread.NewToken(n.getPrivKey(), key)
```
sanderpick, line 179: Finally, return a newly minted token, signed by the host's private key. This is saying, "Okay, as far as I'm concerned, you are who you say you are, and you can use this verifiable token in the future to remind me".

net/net.go
```diff
@@ -127,48 +131,71 @@ func (t *net) Close() (err error) {
...
+}
 
-	args := &core.KeyOptions{}
+func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThreadOption) (info thread.Info, err error) {
+	args := &core.NewThreadOptions{}
 	for _, opt := range opts {
 		opt(args)
 	}
+	// @todo: Check identity key against ACL.
+	identity, err := thread.ValidateToken(n.getPrivKey(), args.Token)
```
sanderpick, line 188: Validate the token (if present). This is where the ACL will get checked.

net/api/client/client.go
```diff
@@ -59,153 +57,206 @@ func (c *Client) GetHostID(ctx context.Context) (peer.ID, error) {
 	return peer.IDFromBytes(resp.PeerID)
 }
 
-func (c *Client) CreateThread(ctx context.Context, creds thread.Credentials, opts ...core.KeyOption) (info thread.Info, err error) {
-	signed, err := signCreds(creds)
+func (c *Client) GetToken(ctx context.Context, identity thread.Identity) (tok thread.Token, err error) {
```
sanderpick, line 60: Here's a client requesting a token with an `Identity`. 
1. Send the remote a pubkey
2. Wait for a challenge
3. Sign the challenge and send back the signature
4. Wait for the final token

net/api/service.go
```diff
@@ -58,18 +57,85 @@ func (s *Service) GetHostID(_ context.Context, _ *pb.GetHostIDRequest) (*pb.GetH
...
+		return nil, status.Error(codes.InvalidArgument, "Signature is required")
+	}
+	return sig, nil
+}
+
+func (i *remoteIdentity) GetPublic() thread.PubKey {
+	return i.pk
+}
+
+func (s *Service) GetToken(server pb.API_GetTokenServer) error {
```
sanderpick, line 92: On the remote-side:
1. Save the received pubkey (it gets included in the JWT)
2. Send back a random challenge (server-side nonce). Again, I don't think we need a client-side nonce here since there can be no man in the middle (SSL). However, it would be super easy to add in the original message send to the remote.
3. `net.GetToken` takes an implementation of `Identity`. We can build one with the bidi server and sent pubkey, and let that method finish the job of minting a token.

net/api/service.go
```diff
@@ -58,18 +57,85 @@ func (s *Service) GetHostID(_ context.Context, _ *pb.GetHostIDRequest) (*pb.GetH
 	}, nil
 }
 
+type remoteIdentity struct {
```
sanderpick, line 60: Here's this gRPC-based `Identity` implementation... more about that below.

sanderpick:
(To clarify the point about getting the caller's peer info from the libp2p connection, I'm referring to the `net.Conn` implementation in github.com/libp2p/go-libp2p-gostream that I'm using as a gRPC dial option. Looking now, it might not be too bad... will look more after this PR.

jsign:
> (To clarify the point about getting the caller's peer info from the libp2p connection, I'm referring to the `net.Conn` implementation in github.com/libp2p/go-libp2p-gostream that I'm using as a gRPC dial option. Looking now, it might not be too bad... will look more after this PR.

To see if I'm understanding it correctly, are you trying to do something similar of what I mentioned above? (auto-quote below):
> Is there some ground to explore about using PeerID in the ACL, or similar? Or maybe manually define the PeerID of the peer to the same key? (if that's possible) Or maybe we want the original owner to take actions in any libp2p peer which doesn't allow to rely on PeerID?
> 
> Just making this question since that solution sounds to good to be true. But want to confirm that that was intentional.

Asking since it seems that the reply at that moment seems that wasn't a good idea, but what you're mentioning now sounds quite similar. Are we talking about the same?

---

This _token generation_ protocol sounds good to me; its aligned with what identity protocols usually do; so sounds good.  I will take a look now.

net/net.go
jsign: I'll take advantage of making a future-related question.

The ACL covers constraints on two dimensions, read and write.
Sounds to me that the _auth_ scheme of this _token_ generation etc, is to cover more the _read_ side of the ACL. So, to answer the following questions:
- I received a request, is this identity known in the ACL? If not, kick out. (So, just general identity check)
- OK, this identity is at least known. He is trying to pull the thread... is he allowed to? (So, read-related actions)

So now comes the important part of the question (at least for me and my understanding, I think @carsonfarmer  was interested in this topic too):
The _write_ side of ACL checks are tied to the _AutorSig_ at the record-level, right?
This because something you insisted about (which is important and true), that received records to be added to a log could be received by another peer than the author. This means that write-related ACL verifications can't be done at the _auth/token_ "layer"... so not here in this `ValidateToken`. Correct?

So, I'd like to confirm with you @sanderpick if that's the same mental model you have in mind.

I've other questions about _AuthSig_ at the record level (as I kind of mentioned in some previous review), but better leaving that for later.

net/net.go
```diff
@@ -127,48 +131,71 @@ func (t *net) Close() (err error) {
...
 
-func (t *net) CreateThread(_ context.Context, creds thread.Credentials, opts ...core.KeyOption) (info thread.Info, err error) {
-	id := creds.ThreadID()
-	if err = t.ensureUnique(id); err != nil {
+func (n *net) GetToken(_ context.Context, identity thread.Identity) (tok thread.Token, err error) {
+	msg := make([]byte, tokenChallengeBytes)
+	if _, err = rand.Read(msg); err != nil {
+		return
+	}
+	sig, err := identity.Sign(msg)
```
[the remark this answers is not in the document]
jsign: Clarif: Mostly mentioned to be _issued_ to be done later, up to you. But important to track of.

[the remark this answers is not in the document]
sanderpick: Nice, good call 👍 I'll make an issue.

sanderpick:
> To see if I'm understanding it correctly, are you trying to do something similar of what I mentioned above? (auto-quote below):
> 
> > Is there some ground to explore about using PeerID in the ACL, or similar? Or maybe manually define the PeerID of the peer to the same key? (if that's possible) Or maybe we want the original owner to take actions in any libp2p peer which doesn't allow to rely on PeerID?
> > Just making this question since that solution sounds to good to be true. But want to confirm that that was intentional.
> 
> Asking since it seems that the reply at that moment seems that wasn't a good idea, but what you're mentioning now sounds quite similar. Are we talking about the same?

Not exactly, but definitely related. The `net` (where records/logs are pushed/pulled) requests don't carry any information about _who the sender is_. It would be nice to pull that info out of the underlying libp2p stream, but it wasn't obvious to me how to do that. So, I added that information to the actual requests. But, to prevent peers from lying to each other about who they are, I added the request signing+verification. Now we can see who is sending requests (calling `PushRecord`, etc.), but to an earlier point, this peer might not be the author. 

To further complicate things, now I'm 80% sure now there is no reason to care about who the sender is at this level (the net p2p level). This information was originally going to be used to restrict who can update log addresses. I thought we should only allow the actual log host (libp2p peer) to update their own addresses. So, in that case we need to know who's calling. Not to get into to much detail of that issue... but I think we'll need to eventually embed address updates in the records so they can be proxied through replicators (like records). So, we can probably remove these verifications, but I'd like to leave them for now until that work starts to be sure. 

Back to the peer not being the author, this is the role of the AuthorSig, which can be used to verify the users _identity_, not peer ID. More on that in response to your other question coming...

net/net.go
sanderpick: > The write side of ACL checks are tied to the AuthorSig at the record-level, right?

Yep, that's right. But the `core.Net` level _can_ do some write validation in `CreateRecord`/`AddRecord`. But this isn't enough on its own—you may have taken an existing thread that you don't have write permission in, hacked your node at the `core.Net` level to ignore outbound write checks (in `CreateRecord` or `AddRecord`), and then added a replicator. Now you send me an invite pointing to your replicator, which starts pushing me records. Based on the thread ID and `AuthorSig` I can verify that in fact you do not have write permission and I can ignore your records.

sanderpick: One last point here, _read_ permission at the lower net p2p level is currently only handled by the _service key_, since reads come from a peer, not an identity. So, if you have the service key, you will be able to get logs and records. There are two problems with this:
1. Whether or not somebody is allowed to have the read key is not baked into the thread ACL rules.
2. I _think_ a peer could trick another peer into delivering them a service key by somehow getting them to use an invite that points back to their host address. However, this would require the invite user to already know the service key. So, not sure if this is a real problem. 

Consider these open questions!

sanderpick: (edited above... I had mentioned a read key but that should have been service)

sanderpick: Hmm, _whether or not somebody is allowed to have the service key_ could be baked into the ACL by using the remote's private key as identity. But still, a node could be hacked to ignore this _outbound_ rule. The ACL rules really only work on the receiver side.

api/client/client.go
```diff
@@ -141,97 +207,100 @@ func collectionConfigToPb(c db.CollectionConfig) (*pb.CollectionConfig, error) {
...
 	return err
 }
 
-// Create creates new instances of objects
-func (c *Client) Create(ctx context.Context, creds thread.Credentials, collectionName string, items ...interface{}) ([]string, error) {
-	signed, err := signCreds(creds)
-	if err != nil {
-		return nil, err
+// Create creates new instances of objects.
+func (c *Client) Create(ctx context.Context, dbID thread.ID, collectionName string, instances Instances, opts ...db.TxnOption) ([]string, error) {
```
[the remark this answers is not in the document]
sanderpick: Oh, good question. I like the variadic style more. It's just that on the server is no longer possible in a clean way. So, toss up: Is it more important to have the APIs match, or to have a simpler gRPC client?

net/net.go
jsign: Considering the above comments, and the other at the PR level... I'd consider we're on the same page... so sounds good!

Now the other big question: _AuthSig_. Trying to explain my question might be long, so I'll take another approach and propose something and see your response to it. I hope that can answer many questions that would be long to do. :P

What if when a log is created, the first record is the public key of the owner (a pubkey in the ACL). This would mean that the log is saying that any record signed by the log keys are also indirectly signed by the author. (Because when the log was created, it said: _The author of this log is X_, and logs are single-writers obviously). Would that allow to avoid using the _AuthSig_ field in *all* the records of the log, and only leaving the _LogSig_ as is now?

net/net.go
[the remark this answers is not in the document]
sanderpick: Seems like it would be good to avoid further API changes, so I just did this now.

net/net.go
sanderpick: Hmm, clever! That's much cleaner. I like it 💯 

Let me take a quick look, maybe I can tag that on quickly. It would be nice to not have `AuthorSig` in there ASAP if we don't need it.

sanderpick: So, when the ACL is in place, the _receiver of a log_ would have to then unpack the first record and verify that the author has write access (check ACL write rule)? If that's what you mean, wouldn't that imply that the receiver _has to have the read key_ in order to do the unpacking?

api/client/client.go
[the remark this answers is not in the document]
asutula: But this isn't varadic style... `Instances` is just []interface{}`. The varadic style also isn't possible here because we're using `TxnOption` as the varadic last parameter. So if I read this right, what we have on the remote is (using `Create` as an example):

`CreateMany(list)`
`CreateOne(single)`

and on the client

`Create(list)`

It makes sense that `Create(list)` is the most general, and we could actually just do that on the remote as well and get rid of the single and many flavors. Just to clarify, varadic instances are not used anywhere anymore.

But since you introduced the single and many flavors, I actually like that because it makes the api easier to use in the most common use case of creating one new object at a time. 

I'd vote for consistency between the remote and client, and I think I'd say that your single and many flavors are a nice way to go.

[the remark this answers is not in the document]
sanderpick: Oh, right, of course... Okay, sounds good. I'm gonna kick the can on updating the client in this PR, but I can create an issue.

net/net.go
jsign: First question: yes.

Second question: Uhm, no?. Isn't the read key needed for the Events? These signatures are on the Record. Am I confused?

api/client/client.go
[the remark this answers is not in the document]
asutula: Sounds good

net/net.go
sanderpick: Oh, I thought you meant write a record that has a body (event) that contains the author's public key. Do you mean, keep `AuthorSig` on record (so readable by service key), but only on the very first record? Or remove `AuthorSig` and for the very first record, use the author's key to create `Sig`, instead of the log key?

jsign: Well, actually didn't thought about a particular implementation.

Both sounds reasonable. One looks "cleaner" than the other, but it has a field only used once.
What are your thoughts? Not a clear-cut winner for me.

jsign: cc @carsonfarmer since we're going down the path for an ACL related thing.

sanderpick: As an aside, I realized here just having the author sig is not enough, we also need the pubkey. So, I'm glad we're having this convo ;)

Okay cool. Ideally, even the first record has the log signature as well. How about this: 

```
type record struct {
	Block     cid.Cid
	Sig       []byte
	AuthorKey []byte
	Prev      cid.Cid `refmt:",omitempty"`
}
```

This may be a little wild, but since we can control the key type of the logs, we could use the asymmetric encryption methods we already have in `crypto` to _encrypt the author sig in `Sig`_ for the first record. On the other end, verification would mean _decrypting `Sig` with log private key, and then verifying the author using the result as a signature. 

Alternatively, we just include a second signature in the first record :)

jsign: I think I understood if I only changed _with log private key_ with _with log public key_, was that the intention? So you're leaning towards "reusing" the field and making the first record a special case. Right?

carsonfarmer: I'm finding it really hard to track this conversation, but I thinks its making sense to me. My only concern with making the 'first' record special is that it means a client can't really trust or do anything until they have synced a full log, right? So assuming they come in 'late' to the game, this could be an issue, yes?

jsign: Uhm, and what's `AuthorKey` used for in that idea?

jsign: > I'm finding it really hard to track this conversation, but I thinks its making sense to me. My only concern with making the 'first' record special is that it means a client can't really trust or do anything until they have synced a full log, right? So assuming they come in 'late' to the game, this could be an issue, yes?

Yes, that's true. Seems like a tradeoff between covering that case, and having some overhead in all the headers of the log. Maybe we can include the info every 200 records or have some middle ground? Just a flashing thought.

sanderpick: Ah, scratch that encryption idea... I did mean private logs key, but I forgot (duh) that the receivers won't (necessarily) have the private key for decryption. 

`AuthorKey` is used to indicate to the receiver who the author is. If they just had a signature, you wouldn't be able to tell who it is or verify.

sanderpick: Hrm, that's a good point @carsonfarmer. Though, every record _could_ carry a link to the first.

jsign: The first record could have:
- `pubkey`, a new field which should map to some pubkey in the ACL. Indicates author pub key. Only has value in first record.
- `Sig` field would be the signature of that pubkey of the record. So this is the only case `Sig` isn't a signature made with the log-private-key.
- No `AuthorSig` field.

What I'm missing in this reasoning?

sanderpick: Funny, I was just typing out... "the sig could just be the sig of the author's pubkey", but now I see that's what you meant :P

Okay, so I'll change the logic and `AuthorKey` -> `PubKey`.

sanderpick: Nice, I like it. I updated the signing+verification to take the first record into account. For now, I'm just including the pubkey in every record, even though only the first one is verifiable. Once we jump on the ACL and the _log record height_ issue (#137), we can adds these links... sound okay? I've already ballooned this PR a good bit.

jsign: SGTM!

PR #296, opened 2020-04-01
Fix concurrency bugs

jsign:
Trying to reproduce some problems @asutula is targeting regarding `go-threads` concurrency bugs, I realized we're not using the `-race` flag to detect these kinds of problems in tests.

So, I ran with the race detector and found some bugs that I've fixed.
There's a particular one that is pretty bad, and other that I'll prefer @sanderpick  opinion since I'm not sure about the original intention of the logic.


.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           go get -v -t -d ./...
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
jsign, line 34: More details [here](https://golang.org/doc/articles/race_detector.html).
We should always enable this flag in CI tests.

db/db_test.go
```diff
@@ -277,10 +278,13 @@ func runListenersComplexUseCase(t *testing.T, los ...ListenOption) []Action {
 	l, err := d.Listen(los...)
 	checkErr(t, err)
 	var actions []Action
+	var wg sync.WaitGroup
```
jsign, line 281: Important since is a test, but nothing core-critical.

net/net.go
```diff
@@ -336,7 +337,9 @@ func (t *net) pullThreadUnsafe(ctx context.Context, id thread.ID) error {
 				log.Error(err)
 				return
 			}
+			lock.Lock()
```
jsign, line 340: This is a very important one. Logs are pulled in parallel and they were appended to a shared slice that wasn't guarded. That can lead to unexpected behavior (loose pulled things, corruption, and wild things).

net/pubsub.go
```diff
@@ -140,7 +140,9 @@ func (s *PubSub) watch(ctx context.Context, id thread.ID, topic *topic) {
 // subscribe to a topic for thread updates.
 func (s *PubSub) subscribe(ctx context.Context, id thread.ID, topic *topic) {
 	var err error
+	s.Lock()
 	topic.s, err = topic.t.Subscribe()
+	s.Unlock()
```
jsign, line 145: I'd like @sanderpick  opinion about this one.

First and foremost, this lock is necessary since `Remove()` reads `topic.s` too. `subscribe()` is called from `Add()`, so on `Add()` and `Remove()` concurrent calls, there is a data-race on `topic.s`. So, with this lock that gets fixed.

Now, as `Add` comments says `Add()` can be called multiple times. And and `Add` call, will lead to a `subscribe()` call, which here looks like it's assigning `topic.s` which is an instance-wide attribute. Is that OK? Seems that an `Add` call will override a previous `Add` call. Mentioning this to be sure that was intentional.

jsign, line 145: This is necessary since `subscribe()` runs on a fired goroutine from `Add()`.
`Add()` gets the lock, adds it to the topic map and release the lock, but the goroutine of `subscribe()` writes to `topic.s` without locking it.

The problem lies on a concurrent call of `Remove()`. Even if `Remove()` is guarded, it finds the `topic` that `Add()` added, and does `topic.s.Remove()`. But The `subscribe()` call is assigning the same `topic.s` without owning it.

It might sound confusing... but this change fixes it. 

@sanderpick , maybe this assignment can be moved to the `Add` function and forget about this lock at the `subscribe()` level. But maybe it was here since `.Subscribe()` may be slow?

net/net.go
sanderpick: Hrm, good catch. It doesn't look like the interior maps will collide, but the appending to the slice will definitely cause issues. Looks like it's from your original concurrency fixes: https://github.com/textileio/go-threads/commit/9deb080fa44326df16a695954f43d96c99c748c4

jsign: Possibly, which solved a ton of other ones.

net/pubsub.go
sanderpick: Ah, good catch! Yep, it can take a bit of time to subscribe, which is why it lives here as you said.

jsign: Great, so this sounds good to me then.

net/net.go
sanderpick: Indeed! The whole thing was hot off the press :) Anyway, not pointing fingers... you just mentioned you were curious about the original intention of the logic.

PR #297, opened 2020-04-01
Foldersync Integration Test

asutula:
Implement what was in go-foldersync as an integration test in here. Fixes some bugs found along the way.


asutula:
Ok I think the tests failures are legit, something we need to figure out the cause of.

asutula:
Summarizing from Slack convo:

 Seeing this sort of result which looks suspicious:
```
Tree of client 0
        d1e1c919-73eb-465c-977c-da1233b8bd47 client0
        0ff57879-16aa-41b3-923a-44f83dddb29d client1
        c21ef22a-054d-464e-8ad2-29278e043c8e client2
        8810e82e-0be6-40a0-a463-a6cc61ff610e client3
        94c8df6e-7b5a-459b-84f8-735447dde55d client4
Tree of client 1
        d1e1c919-73eb-465c-977c-da1233b8bd47 client0
        0ff57879-16aa-41b3-923a-44f83dddb29d client1
Tree of client 2
        d1e1c919-73eb-465c-977c-da1233b8bd47 client0
        c21ef22a-054d-464e-8ad2-29278e043c8e client2
Tree of client 3
        d1e1c919-73eb-465c-977c-da1233b8bd47 client0
        8810e82e-0be6-40a0-a463-a6cc61ff610e client3
Tree of client 4
        d1e1c919-73eb-465c-977c-da1233b8bd47 client0
        94c8df6e-7b5a-459b-84f8-735447dde55d client4
```
The flow is:
1. client 0 is created
2. invite info is created from client 0's db
3. clients 1, 2 , 3 and 4 are created using that invite info
4. each client adds their folder instance to the shared thread

So it's interesting that client 0 does see everyone's updates, but all the other clients only see their own and client 0's

asutula:
Question from @sanderpick:

> Hmm, in the that case, what if you breakpoint the number of logs that are returns to each client in server.go/GetLogs ?

Answer from me:

> it sends 1 log 3 times, once for each additional peer

asutula:
Any strong options on whether or not to include the cli for running a foldersyc setup manually? I left it out for now, but could add it. Seemed more useful when foldersync was meant to be an example, but not sure now that it's meant to be a test.

core/net/net.go
```diff
@@ -48,6 +48,9 @@ type API interface {
 	// DeleteThread with credentials.
 	DeleteThread(ctx context.Context, creds thread.Credentials) error
 
+	// GetThreadAddresses returns the host addresses augmented with peerID and threadID components
+	GetThreadAddresses(ctx context.Context, creds thread.Credentials) ([]ma.Multiaddr, error)
```
asutula, line 52: Added this new method to the `Net` that returns addresses for a specified thread.

db/db.go
```diff
@@ -285,10 +285,26 @@ func (d *DB) Reduce(events []core.Event) error {
...
+	if err != nil {
+		return thread.ID{}, nil, thread.Key{}, err
+	}
+	tinfo, err := d.adapter.net.GetThread(ctx, creds)
+	if err != nil {
+		return thread.ID{}, nil, thread.Key{}, err
+	}
+	return d.adapter.threadCreds.ThreadID(), addresses, tinfo.Key, nil
+}
+
```
asutula, line 302: Then the `DB` can use the two methods from `Net` to build the information needed to invite another peer to a thread.

integrationtests/foldersync/foldersync_test.go
```diff
@@ -0,0 +1,385 @@
...
+
+func TestMain(m *testing.M) {
+	logging.SetLogLevel("foldersync", "info")
+	// logging.SetLogLevel("store", "debug")
+	// logging.SetLogLevel("threads", "debug")
+	// logging.SetLogLevel("threadstore", "debug")
+	os.Exit(m.Run())
+}
+
+func TestSimple(t *testing.T) {
```
asutula, line 32: Just to eliminate the complexity of the bootstrap test below, this test is super simple. Creates multiple DBs sharing a thread, adds data to them and checks for data synchronization

logstore/logstore.go
```diff
@@ -211,6 +213,10 @@ func (ls *logstore) GetLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
 	ls.RLock()
 	defer ls.RUnlock()
 
+	return ls.getLog(id, lid)
+}
+
+func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err error) {
```
asutula, line 219: Bug fix I chased for days

asutula: Thx @jsign !

net/net.go
```diff
@@ -401,6 +406,22 @@ func (t *net) deleteThread(ctx context.Context, id thread.ID) error {
 	return t.store.DeleteThread(id) // Delete logstore keys, addresses, and heads
 }
 
+func (t *net) GetThreadAddresses(ctx context.Context, creds thread.Credentials) ([]ma.Multiaddr, error) {
```
asutula, line 409: Implementation of new `GetThreadAddresses` method

asutula:
@sanderpick and @jsign even though this is still draft and failing tests, I added you to review so we can put our heads together on what the problem might be.

integrationtests/foldersync/foldersync_test.go
```diff
@@ -0,0 +1,385 @@
...
+			for i := 1; i < tt.totalCorePeers; i++ {
+				client, clean := createJoinerClient(t, fmt.Sprintf("client%d", i), tid, addr, key)
+				defer clean()
+				clients = append(clients, client)
+			}
+
+			for i := tt.totalCorePeers; i < tt.totalClients; i++ {
+				index := i % tt.totalCorePeers
+				fmt.Printf("using index %v for client %v\n", index, i)
+				tid, addr, key, err := clients[i%tt.totalCorePeers].getInviteInfo()
```
asutula, line 178: Not sure who originally wrote this part, but can @sanderpick or @jsign explain what the intention of non-core peers is where this modulo is used to choose the peer to get the invite from? Thing is, in all these tests, that always evaluates to 0.

jsign: That's because I think someone commented other tests setups where `totalCorePeers` are greater than 1. If you uncomment those, it evaluates to other values than 0.

The idea is to make peers connect to `totalCorePeers` in invitations. Cycling thought them.

asutula: Sorry, I mis-spoke... there is one instance where it evaluates to 1, but the rest of the time it's 0.

jsign: @asutula, I think we both commented at the same time.
See above just in case you missed it.

jsign: To be clear, that `i%tt.totalCorePeers` is just selecting a core peer for some non-core client. If its 0, or 1, or whatever shouldn't be important. 

It's just distributing invitations evenly on core-peers, that's the intention of the code.

asutula: Ok thanks for the explanation. My statement about it evaluating to 1 in a single case is with all the tests enabled here locally. Since it was non-0 in only that one instance, it made me question if something was wrong, but I get what you're saying... it's a general rule that is applied to every test no matter what the test configuration is.

jsign: Exactly!

core/app/app.go
```diff
@@ -102,8 +102,9 @@ type Net interface {
 
 // Connector connects an app to a thread.
 type Connector struct {
+	Net Net
```
asutula, line 105: Need to make `Net` public so it can be access from the `DB`

core/thread/id.go
```diff
@@ -301,11 +301,12 @@ func (s IDSlice) Len() int           { return len(s) }
...
-// Info holds thread logs and keys.
+// Info holds thread logs, keys and addresses.
 type Info struct {
-	ID   ID
-	Key  Key
-	Logs []LogInfo
+	ID    ID
+	Key   Key
+	Logs  []LogInfo
+	Addrs []ma.Multiaddr
```
asutula, line 309: Added `Addrs` here to represent the thread addresses. It is the `Net` that is responsible for setting this value when appropriate.

db/db.go
asutula: Little update since above comment... This now just calls the underlying `net.GetThread` and returns the useful parts of the result.

asutula: This is were DB needs access to the net connector, and some other places below as well.

db/db.go
```diff
@@ -51,7 +51,6 @@ var (
 type DB struct {
 	io.Closer
 
-	net       app.Net
```
asutula, line 54: This was actually not being set in latest master. We don't need it anyway since `connector.Net` is now accessible.

db/options.go
```diff
@@ -16,11 +16,11 @@ const (
 	defaultDatastorePath = "eventstore"
 )
 
-// Option takes a Config and modifies it.
-type Option func(*Config) error
+// NewDBOption takes a Config and modifies it.
+type NewDBOption func(*NewDBOptions) error
```
asutula, line 20: Needed to rename the options for creating DBs so decided to make them consistent with new option naming conventions.

db/options.go
```diff
@@ -149,3 +149,18 @@ func WithManagedDBToken(t thread.Token) ManagedDBOption {
...
+
+// InviteInfoOption specifies a managed db option.
+type InviteInfoOption func(*InviteInfoOptions)
+
+// WithInviteInfoToken provides authorization for accessing DB invite info.
+func WithInviteInfoToken(t thread.Token) InviteInfoOption {
+	return func(args *InviteInfoOptions) {
+		args.Token = t
+	}
+}
```
asutula, line 166: Above are new options for fetching db invite info. This is basically just a proxy for the thread options that are needed to call GetThread in the net layer.

integrationtests/foldersync/client.go
```diff
@@ -0,0 +1,346 @@
+package foldersync
```
asutula, line 1: All the folder sync code is mostly ported (cleaned up a bit along the way) or copy/pasted/updated from the old folder sync repo.

asutula: I put it in this integration tests directory. We can move other things that are actually integration tests in here as well.

net/api/client/client.go
```diff
@@ -396,10 +396,18 @@ func threadInfoFromProto(reply *pb.ThreadInfoReply) (info thread.Info, err error
...
+		}
+	}
 	return thread.Info{
-		ID:   threadID,
-		Key:  k,
-		Logs: logs,
+		ID:    threadID,
+		Key:   k,
+		Logs:  logs,
+		Addrs: addrs,
```
asutula, line 410: In the net client, be sure to actually use the address data we get from the service.

net/client.go
```diff
@@ -2,6 +2,7 @@ package net
 
 import (
 	"context"
+	"errors"
```
asutula, line 5: These are all changes @sanderpick found to fix multi peer data sync bugs we were seeing.

net/net.go
```diff
@@ -305,11 +305,32 @@ func (n *net) GetThread(_ context.Context, id thread.ID, opts ...core.ThreadOpti
 	if _, err = thread.ValidateToken(n.getPrivKey(), args.Token); err != nil {
 		return
 	}
-	return n.store.GetThread(id)
+	return n.getThreadWithAddrs(id)
 }
 
-func (n *net) getThread(_ context.Context, id thread.ID) (thread.Info, error) {
-	return n.store.GetThread(id)
+func (n *net) getThreadWithAddrs(id thread.ID) (info thread.Info, err error) {
```
asutula, line 311: It seemed you'd want to populate `thread.Info` results with address information in a few cases... thread create, add, and get. In those cases above, we just call this helper method that gets the `thread.Info` from the logstore then augments it with address information.

net/net.go
```diff
@@ -718,7 +739,7 @@ func (n *net) subscribe(ctx context.Context, filter map[thread.ID]struct{}) (<-c
 }
 
 func (n *net) ConnectApp(a app.App, threadID thread.ID) (*app.Connector, error) {
-	info, err := n.getThread(context.Background(), threadID)
+	info, err := n.getThreadWithAddrs(threadID)
```
asutula, line 742: Chose to use the address-augments `thread.Info` results here as well, not sure it matters.

net/net.go
```diff
@@ -606,7 +627,7 @@ func (n *net) AddRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 		return err
 	}
 	if logpk == nil {
-		return fmt.Errorf("log not found")
+		return lstore.ErrLogNotFound
```
asutula, line 630: @sanderpick change for bug fixes

net/net.go
```diff
@@ -764,7 +785,7 @@ func (n *net) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 		return nil
 	}
 	// Get or create a log for the new rec
-	lg, err := n.getLog(id, lid)
+	lg, err := n.store.GetLog(id, lid)
```
asutula, line 788: Then here and below are all @sanderpick bug fixes as well

net/net_test.go
```diff
@@ -161,6 +167,9 @@ func TestNet_AddThread(t *testing.T) {
 	if len(info3.Logs) != 2 {
 		t.Fatalf("expected 2 logs got %d", len(info3.Logs))
 	}
+	if len(info3.Addrs) == 0 {
+		t.Fatalf("expected more than 0 addresses got %d", len(info3.Addrs))
+	}
```
asutula, line 172: Updated these tests to check for the presence of address information in `thread.Info`

asutula:
Seeing some flakiness in the tests. Ironically, not the folder sync tests. I can actually disable the folder sync tests and still get failures locally every few times I run the remaining tests. Seems to fail more reliably on the CI server.

api/pb/api.proto
```diff
@@ -195,7 +195,7 @@ service API {
     rpc GetToken(stream GetTokenRequest) returns (stream GetTokenReply) {}
     rpc NewDB(NewDBRequest) returns (NewDBReply) {}
     rpc NewDBFromAddr(NewDBFromAddrRequest) returns (NewDBReply) {}
-    rpc GetDBInfo(GetDBInfoRequest) returns (GetDBInfoReply) {}
+    rpc GetInviteInfo(GetInviteInfoRequest) returns (GetInviteInfoReply) {}
```
asutula, line 198: This change will require a tiny client tweak downstream cc @carsonfarmer and @andrewxhill

sanderpick:
> Any strong options on whether or not to include the cli for running a foldersyc setup manually? I left it out for now, but could add it. Seemed more useful when foldersync was meant to be an example, but not sure now that it's meant to be a test.

+1 for leaving it out. 

Related, I'd vote for removing all the existing examples and starting fresh with some that use the clients.

api/client/client.go
```diff
@@ -215,16 +215,32 @@ func collectionConfigToPb(c db.CollectionConfig) (*pb.CollectionConfig, error) {
 	}, nil
 }
 
-// GetDBInfo retrives db addresses and keys.
-func (c *Client) GetDBInfo(ctx context.Context, dbID thread.ID, opts ...db.ManagedDBOption) (*pb.GetDBInfoReply, error) {
+// GetInviteInfo retrives db addresses and keys.
+func (c *Client) GetInviteInfo(ctx context.Context, dbID thread.ID, opts ...db.ManagedDBOption) ([]ma.Multiaddr, thread.Key, error) {
```
sanderpick, line 219: Could this be `GetDBInvite`, so that it's clear it returns a DB invite, as apposed to a collection invite or something else?

Also, you might use this for getting the key.. as in, not when you need an invite. That's why I had it more generic: `GetDBInfo`

db/db.go
```diff
@@ -289,10 +288,25 @@ func (d *DB) Reduce(events []core.Event) error {
...
 }
 
+// GetInviteInfo returns the addresses and key that can be used to join the DB thread
+func (d *DB) GetInviteInfo(opts ...InviteInfoOption) ([]ma.Multiaddr, thread.Key, error) {
+	options := &InviteInfoOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	tinfo, err := d.connector.Net.GetThread(context.Background(), d.connector.ThreadID(), net.WithThreadToken(options.Token))
```
sanderpick, line 298: 👍

db/options.go
sanderpick: Sweeeeet

net/net.go
```diff
@@ -764,7 +785,7 @@ func (n *net) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 		return nil
 	}
 	// Get or create a log for the new rec
```
sanderpick, line 787: Not a PR change here, but this should probably read "Get the log" (nothing about create), or just be removed

api/client/client.go
asutula: Ah didn't realize there were other use cases. In that case, I'll just change it back to GetDBInfo.

net/net.go
asutula: Done

asutula:
Ok looks like 3 peer folder sync tests works! Merging...

PR #300, opened 2020-04-02
Use author pub key as first rec sig payload

sanderpick:
I forgot to push this last change for https://github.com/textileio/go-threads/pull/293.


net/client.go
```diff
@@ -269,6 +269,10 @@ func (s *server) getRecords(ctx context.Context, id thread.ID, lid peer.ID, offs
 						log.Error(err)
 						return
 					}
+					if err = rec.Verify(lg.PubKey); err != nil {
```
sanderpick, line 272: Something to note... records from `getRecords` weren't being verified 😨

PR #302, opened 2020-04-03
Bug fix picks

asutula:
These are the fixes we identified in the course of working on my folder sync branch, but here, applied to master.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -606,7 +606,7 @@ func (n *net) AddRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 		return err
 	}
 	if logpk == nil {
-		return fmt.Errorf("log not found")
+		return lstore.ErrLogNotFound
 	}
 
 	knownRecord, err := n.bstore.Has(rec.Cid())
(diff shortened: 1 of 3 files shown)


net/net.go
```diff
@@ -855,9 +854,6 @@ func (n *net) getLocalRecords(ctx context.Context, id thread.ID, lid peer.ID, of
 	if err != nil {
 		return nil, err
 	}
-	if lg.PubKey == nil {
-		return nil, fmt.Errorf("log not found")
-	}
```
jsign, line 860: In what situation we could have a Log but without its public key?
I ask since this check is removed, but I don't see any other change regarding `GetLog` logic?
cc @sanderpick

asutula:
One thing to keep in mind here... the tests passed, but there is a chance there is something wrong with these changes and the existing tests (on master) are pushing things hard enough with multiple peers to see any failures. I'm a little suspicious of that because the tests in my folder sync branch are much more unstable (of course there could be some other change in that branch that is causing problems)

net/net.go
sanderpick: > In what situation we could have a Log but without its public key?

There wouldn't be a situation like that anymore since GetLog now returns an error if it's not found. This was the old way to check if it was "empty", since info is not a pointer receiver.

jsign: Nice, much better then!

PR #304, opened 2020-04-06
ulid for instance id

### LINKED ISSUE #280: Determine instanceID policy
JSON mode originally had a strict uuid policy where non-json mode allowed anything. For now we're allowing anything, but let's discuss what policy should be applied. Some notes from slack convo:

@sanderpick: For example, ObjectIDs (in mongo) have a prefixed time component

@jsign: Maybe we can add an extra flag for ID validations... that might help for flexibility, and as a free solution to tests too

asutula:
Pretty simple change since we're allowing clients to use any string id if they want.

closes #280


core/db/db.go
```diff
@@ -16,18 +19,19 @@ type InstanceID string
 
 // NewInstanceID generates a new identity for an instance.
 func NewInstanceID() InstanceID {
-	return InstanceID(uuid.New().String())
+	t := time.Now().UTC()
+	entropy := ulid.Monotonic(rand.New(rand.NewSource(t.UnixNano())), 0)
+	id := ulid.MustNew(ulid.Timestamp(t), entropy)
+	return InstanceID(id.String())
```
asutula, line 25: The api for this go ulid lib is kinda ugly. I'm not quite following what the docs [here](https://github.com/oklog/ulid#usage) mean about the thread safety issue. The way I've implemented this method, it's creating a new `rand.Rand` each time this method is called. Input welcome here.

core/db/db.go
```diff
@@ -16,18 +19,19 @@ type InstanceID string
...
 func (e InstanceID) String() string {
 	return string(e)
 }
 
+// IsValidInstanceID checks if an id is valid
 func IsValidInstanceID(instanceID string) bool {
-	// ToDo: Decide how we want to apply instance id requirements
 	return len(instanceID) > 0
-	// _, err := uuid.Parse(instanceID)
-	// return err == nil
```
asutula, line 30: We can always add more to this in the future, but this is the only check needed for now since we're allowing clients to specify any string as an id.

db/collection.go
```diff
@@ -451,18 +459,18 @@ func (t *Txn) Discard() {
 	t.discarded = true
 }
 
-func getInstanceID(t []byte) core.InstanceID {
+func getInstanceID(t []byte) (core.InstanceID, error) {
 	partial := &struct{ ID *string }{}
 	if err := json.Unmarshal(t, partial); err != nil {
-		log.Fatalf("error when unmarshaling json instance: %v", err)
+		return core.EmptyInstanceID, err
```
asutula, line 465: `log.Fatal` seems strange to me. I'm not sure why we'd ever want to actually just kill the process when there is an error we can return to the caller. I made this change here since I was working around/in this code, but this `log.Fatal` is used other places as well. If there is a good reason to revert this, let me know.

asutula:
Not sure what's up with the tests. There is a folder sync test failing because the client trees aren't equal, but they look perfectly equal in the log output. And it passes for me locally 😞

core/db/db.go
sanderpick: I think he's referring to the `math/rand` package specifically as opposed to `crypto/rand`. Looking at the tests, you could do something simpler like:

```
id := ulid.MustNew(ulid.Now(), rand.Reader)
```

Where `rand` is from `crypto/rand`. Guessing `ulid.Now()` returns now in millis.

db/collection.go
sanderpick: There are cases where fatal makes sense. For example, "I've encountered an error that indicates a fatal flaw on logic and I should not be used anymore". There are a number of places where fatal is used like that in here. This one around getting the instance ID _could_ be like that. The system allowed the ID to enter, if it can't get it back out, something bad happened that can't be recovered from. It's nice to have a distinction in the code the separates these kinds of things. But, I'm not super partial to one direction for this one.

sanderpick: It would kill their daemon, hopefully they would report the fatal bug.

PR #307, opened 2020-04-07
Run foldersync tests separately

asutula:
Even though the folder sync tests themselves were running serially, I think they were still running in parallel with all the other tests. Here we try to run them in a completely separate CI step.


asutula:
Ok, this seems much more stable this way. I was able to re-enable the 3 peer folder sync tests, which is crucial.

asutula:
Yea, good thought. It's an easy change if/when we add some other integration tests.

PR #308, opened 2020-04-11
Textile integration nits

### LINKED ISSUE #306: libp2p host key vs ACL
In https://github.com/textileio/go-threads/pull/300 you've added the libp2p host public key to the first record in the log to match the log and the peer that own it. And according to the discussion in the https://github.com/textileio/go-threads/pull/293 it looks like you are going to implement thread ACL based on this node's pub key

So here is come the question. How are you going to handle __1 user__ – __multiple devices__ in that case? 

libp2p is not supposed to share identity between nodes: https://discuss.ipfs.io/t/can-multiple-nodes-share-the-same-peer-id/1204

Scenario:
- Thread owner grants the user permission to read and write to the thread
- Later this user want to access this thread from another client/device he/she owns

Questions:
- Who will be responsible for adding the other user device host ID to the ACL? 
- How to indicate that both logs/pubkeys belongs to the same user. Looks like it will be the must-have scenario for applications
- What if the user doesn't have ACL modify permission (e.g. can't add one more user with write permission)?

TLDR
- libp2p key, used in the 1st log's record looks like a device key, not a user key
- on the other hand looks like ACL should rely on the user itself, not on the device(libp2p host) key

At @anytypeio we've implemented the second layer cbornode on top of Record to store both device and user pubkey&signature. But it will be better to implement on the level of go-threads Logs. I think this is a pretty common use-case.

Maybe think to add the non-ephemeral user pubkey/signature alongside with the libp2p host key?

sanderpick:
- Closes #306 
- Updates to Go 1.14.2
- Changes gRPC service names to be more explicit, which is less confusing in other services (textile)


api/pb/api.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package api.pb;
+package threads.pb;
```
carsonfarmer, line 2: Will this have an impact on the exported Typescript/Javascript libraries?

net/api/pb/api.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package net.api.pb;
+package threads.net.pb;
```
carsonfarmer, line 2: Same question here.

PR #310, opened 2020-04-14
Don't wait while pushing

### LINKED ISSUE #309: CreateRecords waits for push requests to be finished
Here at `pushRecord`, you are waiting for all record push requests to the log addresses  to be done:
https://github.com/textileio/go-threads/blob/e1b3ffe98cb4c525c3ce3de99a4ba26f9e926223/net/client.go#L390

It can take a while when you have replicators for your log. For me it looks like CreateRecord execution time shouldn't depend on the number of addresses and their response time. 

And because it doesn't pass and return any error from the actual push request it looks like it waitgroup may be removed without any bad effects.

sanderpick:
Closes #309 

Also pushes records added by `AddRecord`.

### DIFF
--- a/net/client.go
+++ b/net/client.go
@@ -316,11 +316,8 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec 
 	}
 
 	// Push to each address
-	wg := sync.WaitGroup{}
 	for _, addr := range addrs {
-		wg.Add(1)
 		go func(addr ma.Multiaddr) {
-			defer wg.Done()
 			p, err := addr.ValueForProtocol(ma.P_P2P)
 			if err != nil {
 				log.Error(err)
@@ -337,7 +334,7 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec 
 
 			log.Debugf("pushing record to %s...", p)
 
-			cctx, cancel := context.WithTimeout(ctx, reqTimeout)
+			cctx, cancel := context.WithTimeout(context.Background(), reqTimeout)
 			defer cancel()
 			conn, err := s.dial(cctx, pid, grpc.WithInsecure())
 			if err != nil {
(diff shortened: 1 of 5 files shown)


net/client.go
```diff
@@ -386,8 +383,6 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
 	if err = s.ps.Publish(ctx, id, req); err != nil {
 		log.Errorf("error publishing record: %s", err)
 	}
-
-	wg.Wait()
```
sanderpick, line 390: I can't remember why we waited here... probably out of an abundance of caution while developing.

net/net.go
```diff
@@ -819,10 +822,10 @@ func (n *net) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 
 		log.Debugf("put record %s (thread=%s, log=%s)", r.Cid(), id, lg.ID)
 
-		if err = n.bus.SendWithTimeout(NewRecord(r, id, lg.ID), notifyTimeout); err != nil {
+		if err = n.store.SetHead(id, lg.ID, r.Cid()); err != nil {
 			return err
 		}
-		if err = n.store.SetHead(id, lg.ID, r.Cid()); err != nil {
+		if err = n.bus.SendWithTimeout(NewRecord(r, id, lg.ID), notifyTimeout); err != nil {
```
sanderpick, line 828: Just flipping the order here. Seems safer to set the head before broadcasting.

net/net.go
```diff
@@ -642,7 +642,10 @@ func (n *net) AddRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
 	if err = rec.Verify(logpk); err != nil {
 		return err
 	}
-	return n.PutRecord(ctx, id, lid, rec)
+	if err = n.PutRecord(ctx, id, lid, rec); err != nil {
+		return err
+	}
+	return n.server.pushRecord(ctx, id, lid, rec)
```
sanderpick, line 648: Surely this must have been a bug... why not push out records from `AddRecord` as well as `CreateRecord`?

jsign: Yeah. Eventually, this can be different if we plan to batch pushes or similar, but not the case now.
SGTM.

sanderpick: Nice yeah, good call

PR #319, opened 2020-04-21
Use _id as instance identifier key

sanderpick:
It's all JSON now, so we may as well use JSON-like keys. The underscore in `_id` reduces conflict with user chosen keys. Thoughts?


carsonfarmer:
Yup, love this. In reality, the majority of usage of these APIs is going to be from clients, not Go, so this feels like a good tradeoff IMO (says the guy who won't ever use the APIs from Go) ;)

sanderpick:
Agreed, it's a downside on the Go side, but offers more flex in the end. Let's see how it feels.

PR #324, opened 2020-04-28
Update ipfs-lite, test data races, and other implications

jsign:
- Update `ipfs-lite` which cascaded to a lot of other important updates
- Fix all broken things from that update


common/common.go
```diff
@@ -62,20 +60,14 @@ func DefaultNetwork(repoPath string, opts ...NetOption) (NetBoostrapper, error)
...
 
 	ctx, cancel := context.WithCancel(context.Background())
-	pstore, err := pstoreds.NewPeerstore(ctx, litestore, pstoreds.DefaultOpts())
-	if err != nil {
-		litestore.Close()
-		cancel()
-		return nil, err
-	}
 	priv := util.LoadKey(filepath.Join(ipfsLitePath, "key"))
 	h, d, err := ipfslite.SetupLibp2p(
```
jsign, line 64: IPFS-lite now doesn't need an explicit peerstore, but requires a batching datastore.

db/index.go
```diff
@@ -88,7 +88,7 @@ func indexUpdate(baseKey ds.Key, path string, index Index, tx ds.Txn, key ds.Key
 		return nil
```
jsign, line 88: Indexes where relying on the _prefix query will contain non-child keys_ assumption.

db/query_more_test.go
```diff
@@ -94,7 +94,7 @@ var (
```
jsign, line 94: Test ran in parallel, but `createCollectionWithData` was modifying the global variable!
Now that method returns the modified dataset from the sample data, so parallel runs use their own data sample.

db/query_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
```
jsign, line 6: Same as file above.

go.mod
```diff
@@ -18,35 +18,35 @@ require (
...
-	github.com/ipfs/go-log v1.0.0
-	github.com/ipfs/go-merkledag v0.2.3
-	github.com/libp2p/go-libp2p v0.4.2
-	github.com/libp2p/go-libp2p-connmgr v0.1.1
-	github.com/libp2p/go-libp2p-core v0.3.0
+	github.com/ipfs/go-ipld-cbor v0.0.4
+	github.com/ipfs/go-ipld-format v0.2.0
+	github.com/ipfs/go-log v1.0.4
+	github.com/ipfs/go-merkledag v0.3.2
+	github.com/libp2p/go-libp2p v0.8.2
```
jsign, line 37: From v0.4.2 to v0.8.2

go.mod
```diff
@@ -18,35 +18,35 @@ require (
 	github.com/gogo/status v1.1.0
 	github.com/golang/protobuf v1.3.3
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.0
-	github.com/hashicorp/go-multierror v1.0.0
-	github.com/hashicorp/golang-lru v0.5.3
-	github.com/hsanjuan/ipfs-lite v0.1.8
+	github.com/hashicorp/go-multierror v1.1.0
+	github.com/hashicorp/golang-lru v0.5.4
+	github.com/hsanjuan/ipfs-lite v1.1.11
```
jsign, line 23: The innocent trigger...

go.mod
```diff
@@ -18,35 +18,35 @@ require (
...
 	github.com/improbable-eng/grpc-web v0.12.0
 	github.com/ipfs/go-block-format v0.0.2
-	github.com/ipfs/go-blockservice v0.1.2
+	github.com/ipfs/go-blockservice v0.1.3
 	github.com/ipfs/go-cid v0.0.5
-	github.com/ipfs/go-datastore v0.3.1
-	github.com/ipfs/go-ds-badger v0.2.0
-	github.com/ipfs/go-ipfs-blockstore v0.1.1
+	github.com/ipfs/go-datastore v0.4.4
+	github.com/ipfs/go-ds-badger v0.2.4
```
jsign, line 29: More deep ones. Having them in v0.x is a problem for semver...

logstore/lstoreds/keybook.go
```diff
@@ -175,18 +175,17 @@ func (kb *dsKeyBook) ClearKeys(t thread.ID) error {
 
 // ClearLogKeys deletes all keys under a log.
 func (kb *dsKeyBook) ClearLogKeys(t thread.ID, p peer.ID) error {
-	if err := kb.clearKeys(dsLogKey(t, p, kbBase).Child(privSuffix)); err != nil {
```
jsign, line 178: Relied on the assumption that prefix will contain the exact key match.
Switched to simply deleting directly.

db/query_more_test.go
sanderpick: 😨

go.mod
sanderpick: People are addicted to the idea of a v0 release

PR #327, opened 2020-04-29
revert to explicit peerstore for persistence

### LINKED ISSUE #326: libp2p peerstore now in-memory
After #324 by @jsign was merged, now go-threads will rely on the default libp2p peerstore, which is in-memory: https://github.com/libp2p/go-libp2p/blob/master/defaults.go#L48

But it looks like you assume that it is permanents: https://github.com/textileio/go-threads/blob/master/net/net.go#L528

Also, in-memory peerstore may be especially painful for mobile clients, which can be unloaded pretty often and will need to do discovery again and again

jsign:
Closes #326

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -198,6 +198,9 @@ func (tsb *netBoostrapper) Close() error {
 	if err := tsb.host.Close(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


jsign:
Thanks @asutula !

PR #331, opened 2020-05-01
fix(daemon): replaces fatal with error in collections methods

andrewxhill:


db/collection.go
```diff
@@ -456,13 +456,13 @@ func getInstanceID(t []byte) core.InstanceID {
 		ID *string `json:"_id"`
 	}{}
 	if err := json.Unmarshal(t, partial); err != nil {
-		log.Fatalf("error when unmarshaling json instance: %v", err)
+		log.Errorf("error when unmarshaling json instance: %v", err)
```
sanderpick, line 459: Instead of logging, should this return an error to the user? Thoughts @jsign?

jsign: I double checked reverting the `Fatal` and it seems that is ok since this is user input, and not unmarshaling presumably valid data that was already stored (which sounds to me like a fatal thing... should never happen). (Mentioning this since this comes from other discussion thread).

Regardin what you mention: Agree on that error be returned. (shouldn't move on)

sanderpick:
Closing this in favor of https://github.com/textileio/go-threads/pull/338

PR #338, opened 2020-05-05
collections: allow missing id on create

### LINKED ISSUE #337: threadsd crashes when inserting an instance without an ID attribute
**Current Behaviour**
When creating a model instance using the `@textitle/threads-client` js client, and I try to create a new instance in my documentation like so:
```
client.create(threadId, collectionName, [
      {
        slug: 'First slug',
        hash: 'QmWuvwp71MFr7vVkyC7teUJoGV6oDDoz9DD2Uk8uiXZHoF',
      },
    ]);
```
My `threadsd` crashes because an instance `ID` is not provided.

**Expected Behaviour**
`threadsdb` should not crash, instead it should return an error or some sort of failure, that the client can handle.

**Other Information**
Environment: Running go_threads v0.1.14
Error Stacktrace:
```
2020-05-05T15:07:22.370+0100	FATAL	db	db/collection.go:460	invalid instance: doesn't have an ID attribute
github.com/textileio/go-threads/db.getInstanceID
	/home/runner/work/go-threads/go-threads/db/collection.go:460
github.com/textileio/go-threads/db.(*Txn).Create
	/home/runner/work/go-threads/go-threads/db/collection.go:303
github.com/textileio/go-threads/db.(*Collection).CreateMany.func1
	/home/runner/work/go-threads/go-threads/db/collection.go:191
github.com/textileio/go-threads/db.(*DB).writeTxn
	/home/runner/work/go-threads/go-threads/db/db.go:422
github.com/textileio/go-threads/db.(*Collection).WriteTxn
	/home/runner/work/go-threads/go-threads/db/collection.go:159
github.com/textileio/go-threads/db.(*Collection).CreateMany
	/home/runner/work/go-threads/go-threads/db/collection.go:190
github.com/textileio/go-threads/api.(*Service).processCreateRequest
	/home/runner/work/go-threads/go-threads/api/service.go:659
github.com/textileio/go-threads/api.(*Service).Create
	/home/runner/work/go-threads/go-threads/api/service.go:284
github.com/textileio/go-threads/api/pb._API_Create_Handler
	/home/runner/work/go-threads/go-threads/api/pb/api.pb.go:2454
google.golang.org/grpc.(*Server).processUnaryRPC
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.25.1/server.go:1007
google.golang.org/grpc.(*Server).handleStream
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.25.1/server.go:1287
google.golang.org/grpc.(*Server).serveStreams.func1.1
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.25.1/server.go:722
```

sanderpick:
Closes #337 

Allows a `nil` `_id` attribute on instances that are being created.

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -153,6 +153,26 @@ func TestClient_Create(t *testing.T) {
 			t.Fatalf("failed to create collection: %v", err)
 		}
 	})
+
+	t.Run("test collection create with missing id", func(t *testing.T) {
+		id := thread.NewIDV1(thread.Raw, 32)
+		err := client.NewDB(context.Background(), id)
+		checkErr(t, err)
+		err = client.NewCollection(context.Background(), id, db.CollectionConfig{Name: collectionName, Schema: util.SchemaFromSchemaString(schema)})
+		checkErr(t, err)
+
+		ids, err := client.Create(context.Background(), id, collectionName, Instances{&PersonWithoutID{
+			FirstName: "Adam",
+			LastName:  "Doe",
+			Age:       21,
+		}})
+		if err != nil {
+			t.Fatalf("failed to create collection: %v", err)
+		}
+		if len(ids) != 1 {
+			t.Fatal("expected a new id, got none")
+		}
+	})
 }
 
 func TestClient_Save(t *testing.T) {
@@ -693,7 +713,6 @@ const (
 	"$schema": "http://json-schema.org/draft-07/schema#",
 	"title": "` + collectionName + `",
 	"type": "object",
-	"required": ["_id"],
 	"properties": {
 		"_id": {
 			"type": "string",
@@ -718,6 +737,12 @@ const (
 
 type Person struct {
 	ID        string `json:"_id"`
+	FirstName string `json:"firstName,omitempty"`
+	LastName  string `json:"lastName,omitempty"`
+	Age       int    `json:"age,omitempty"`
+}
+
+type PersonWithoutID struct {
 	FirstName string `json:"firstName,omitempty"`
 	LastName  string `json:"lastName,omitempty"`
 	Age       int    `json:"age,omitempty"`
--- a/db/collection.go
+++ b/db/collection.go
@@ -458,15 +458,15 @@ func (t *Txn) Discard() {
 	t.discarded = true
 }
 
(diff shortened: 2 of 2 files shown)


api/client/client_test.go
```diff
@@ -693,7 +713,6 @@ const (
 	"$schema": "http://json-schema.org/draft-07/schema#",
 	"title": "` + collectionName + `",
 	"type": "object",
-	"required": ["_id"],
```
sanderpick, line 696: I removed this from the schema since it shouldn't be required of the user to demand that this is "required", but also since it was preventing the issue from being tested, since the schema validation would return an error before reaching the fatal error.

db/collection.go
```diff
@@ -451,20 +458,17 @@ func (t *Txn) Discard() {
...
 		ID *string `json:"_id"`
 	}{}
 	if err := json.Unmarshal(t, partial); err != nil {
-		log.Fatalf("error when unmarshaling json instance: %v", err)
+		return id, fmt.Errorf("error unmarshaling json instance: %v", err)
 	}
 	if partial.ID == nil {
-		log.Fatal("invalid instance: doesn't have an _id attribute")
-	}
-	if *partial.ID != "" && !core.IsValidInstanceID(*partial.ID) {
```
sanderpick, line 464: This looked to be an impossible condition: _an empty string and an invalid instance_, since the only way an instance is invalid is if it's not an empty string! I think the validity check used to check for UUID type or something.

sanderpick:
Thanks @carsonfarmer - I'll wait for a peek from @asutula since he was the last one that touched these lines.

db/collection.go
```diff
@@ -451,20 +458,17 @@ func (t *Txn) Discard() {
...
-		log.Fatal("invalid instance: invalid _id value")
+		return id, errMissingInstanceID
```
asutula, line 469: Think we should return `core.EmptyInstanceID` explicitly here since that's what is tested for to decide if we need to generate an id for the new instance?

asutula: Here and a couple lines above as well

sanderpick: Ah nice, good call. Just pushed the change.

PR #339, opened 2020-05-06
Retain client connections

### LINKED ISSUE #276: Retain connections with net peers
Currently, a new connection is created every time a `net` dial is needed. We should maintain a map of host connection.

sanderpick:
Closes #276 

@requilence, here's a quick stab at fixing the leaky client issue. Can you give it a spin to see if we're in the right direction?


net/client.go
```diff
@@ -107,13 +106,12 @@ func (s *server) pushLog(ctx context.Context, id thread.ID, lg thread.LogInfo, p
...
 
-	cctx, cancel := context.WithTimeout(ctx, reqTimeout)
-	defer cancel()
-	conn, err := s.dial(cctx, pid, grpc.WithInsecure())
+	client, err := s.dial(pid)
 	if err != nil {
 		return fmt.Errorf("dial %s failed: %s", pid, err)
 	}
-	client := pb.NewServiceClient(conn)
+	cctx, cancel := context.WithTimeout(ctx, reqTimeout)
```
requilence, line 113: maybe you can export reqTimeout(and maybe split PushTime/PullTimeout) so we can play with it?

requilence:
According to the max number of connections, I believe it should be handled by libp2p's low and high watermarks. This way we shouldn't care a lot about this, as this should be a very little overhead to store a pointer to a closed connection.

But I believe you need to handle grpc conn state in case libp2p closes the connection: https://github.com/textileio/go-threads/pull/339#discussion_r420626771

net/client.go
```diff
@@ -387,9 +383,21 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
 // dial attempts to open a gRPC connection over libp2p to a peer.
-func (s *server) dial(ctx context.Context, peerID peer.ID, dialOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	opts := append([]grpc.DialOption{s.getDialOption()}, dialOpts...)
-	return grpc.DialContext(ctx, peerID.Pretty(), opts...)
+func (s *server) dial(peerID peer.ID) (pb.ServiceClient, error) {
+	s.Lock()
+	defer s.Unlock()
+	conn, ok := s.conns[peerID]
+	if ok {
+		return pb.NewServiceClient(conn), nil
```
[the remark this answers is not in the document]
sanderpick: Yep, nice... I wasn't sure how those connections are managed. I assume they have it working pretty magically, but probably same to assume that if the state is `Shutdown`, we can close it. Will do.

sanderpick:
Sounds good, thanks for the 👀 . I'll open it up later today... headed out for a bit this morning.

net/client.go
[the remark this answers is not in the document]
sanderpick: Back in here now. Looks like "TransientFailure indicates the ClientConn has seen a failure but expects to recover.". So, I'm not sure we need to handle that case.

integrationtests/foldersync/foldersync_test.go
```diff
@@ -296,37 +295,7 @@ func EqualFiles(c1 *client, f1 file, c2 *client, f2 file) bool {
...
-		}
-		b2, err := ioutil.ReadAll(r2)
-		if err != nil {
-			panic(err)
-		}
-		if !bytes.Equal(b1, b2) {
-			return false
-		}
-	} else {
+	if f1.IsDirectory {
```
jsign, line 298: I think the whole `EqualFiles` could be avoided... but this change is enough to reduce the noise.
Eventually we'll take out all this files example, so good.

net/client.go
```diff
@@ -387,17 +384,35 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
...
+	if ok {
+		if conn.GetState() == connectivity.Shutdown {
+			if err := conn.Close(); err != nil {
+				log.Errorf("error closing connection: %v", err)
+			}
+		} else {
+			return pb.NewServiceClient(conn), nil
+		}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), RequestTimeout)
```
jsign, line 400: Nit: Better to call this `DialTimeout`?

sanderpick: Nice, that's better... will do.

PR #343, opened 2020-05-09
util: fix typo in func name DefaultBootstrapPeers()

benjaminbollen:
While going over the `examples/chat/main.go` I noticed a innocent typo in the function name of `util/DefaultBootstrapPeers()` (was missing a t). If it's helpful, I've made the name change; if you feel it breaks interface and currently not helpful, no problem either!

PS: we started a new project and will be building on threads, so hopefully I'll be around with more in-depth contributions. thanks for the work!


benjaminbollen:
fails at checking out code. closing for now.

PR #344, opened 2020-05-12
Make thread.ID just a string alias

asutula:
Instead of a struct. The inner `str` property was private and therefore, not serializable to json. This cropped up as a problem when @Schwartz10 started trying to store thread ids in a collection. This will likely be a common thing to do.

The options were to make `str` public `Str` or to just make `thread.ID` an alias for `string`. I went with the later because it's more simple and consistent with our instance id type.

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -165,7 +165,7 @@ func TestIt(t *testing.T) {
 
 	_, err = client.Create(context.Background(), id, collectionName, Instances{owner})
 	if err != nil {
-		t.Fatalf("failed to create collection: %v", err)
+		t.Fatalf("failed to create instance: %v", err)
 	}
 }
 
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -22,6 +22,7 @@ import (
 	. "github.com/textileio/go-threads/api/client"
 	pb "github.com/textileio/go-threads/api/pb"
 	"github.com/textileio/go-threads/common"
+	core "github.com/textileio/go-threads/core/db"
 	"github.com/textileio/go-threads/core/thread"
 	"github.com/textileio/go-threads/db"
 	"github.com/textileio/go-threads/util"
@@ -134,6 +135,38 @@ func TestClient_NewCollection(t *testing.T) {
 			t.Fatalf("failed add new collection: %v", err)
 		}
 	})
+}
+
+func TestIt(t *testing.T) {
+	type Owner struct {
+		ID       core.InstanceID `json:"_id"`
+		ThreadID thread.ID       `json:"threadId"`
+	}
+	collectionName := "owners"
+
+	client, done := setup(t)
+	defer done()
+
+	id := thread.NewIDV1(thread.Raw, 32)
+	err := client.NewDB(context.Background(), id)
+	checkErr(t, err)
+	sch := util.SchemaFromInstance(Owner{}, false)
+	schBytes, err := json.MarshalIndent(sch, "", "  ")
+	checkErr(t, err)
+	schStr := string(schBytes)
+	t.Logf("\n%v", schStr)
+	err = client.NewCollection(context.Background(), id, db.CollectionConfig{Name: collectionName, Schema: sch})
+	checkErr(t, err)
+
+	owner := &Owner{
+		ID:       "",
+		ThreadID: id,
+	}
+
+	_, err = client.Create(context.Background(), id, collectionName, Instances{owner})
+	if err != nil {
+		t.Fatalf("failed to create collection: %v", err)
+	}
 }
 
 func TestClient_Create(t *testing.T) {
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -150,7 +150,7 @@ func TestIt(t *testing.T) {
 	id := thread.NewIDV1(thread.Raw, 32)
 	err := client.NewDB(context.Background(), id)
 	checkErr(t, err)
-	sch := util.SchemaFromInstance(Owner{}, false)
+	sch := util.SchemaFromInstance(Owner{}, false, true)
 	schBytes, err := json.MarshalIndent(sch, "", "  ")
 	checkErr(t, err)
 	schStr := string(schBytes)
@@ -163,9 +163,21 @@ func TestIt(t *testing.T) {
 		ThreadID: id,
 	}
 
-	_, err = client.Create(context.Background(), id, collectionName, Instances{owner})
+	ids, err := client.Create(context.Background(), id, collectionName, Instances{owner})
 	if err != nil {
 		t.Fatalf("failed to create instance: %v", err)
+	}
+
+	another := &Owner{}
+
+	err = client.FindByID(context.Background(), id, collectionName, ids[0], another)
+	if err != nil {
+		t.Fatalf("failed to find instance: %v", err)
+	}
+	one := owner.ThreadID.String()
+	two := another.ThreadID.String()
+	if one != two {
+		t.Fatalf("thread id %v != %v", one, two)
 	}
 }
 
--- a/core/thread/id.go
+++ b/core/thread/id.go
@@ -86,7 +86,7 @@ var Undef = ID("")
 // Calling any other methods on an undefined ID will result in
 // undefined behavior.
 func (i ID) Defined() bool {
-	return i != Undef
+	return i != ""
 }
 
 // Decode parses an ID-encoded string and returns an ID object.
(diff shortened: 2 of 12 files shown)


asutula:
Ah yea good call. I'll do that.

core/thread/id.go
```diff
@@ -294,6 +294,19 @@ func (i ID) Loggable() map[string]interface{} {
 	}
 }
 
+func (i ID) MarshalJSON() ([]byte, error) {
+	return i.Bytes(), nil
```
asutula, line 298: @sanderpick I know you mentioned somehow using the multibase methods in here, but I thought I'd try to just use the existing methods that convert `ID` to and from `[]byte`. When having `encoding/json` marshal a `Person` in the test below, I get this error:
```json: error calling MarshalJSON for type thread.ID: invalid character '\x01' looking for beginning of value```
Ideas?

Comment (asutula) [reply core/thread/id.go:298]: To clarify, `json/encoding` chokes on the bytes returned from here and gives that error.

Comment (sanderpick) [reply core/thread/id.go:298]: Nice, good call on using existing methods. Not sure what's up with that. Taking a look.

Comment (sanderpick) [reply core/thread/id.go:298]: Looking here @asutula - Do you know what the original error was that led to this work? Running your new test without the custom marshalers seems to be fine. It looks like the `json` package falls back to using `TextMarshaler`, which is already implemented on `thread.ID`. I copied that from the `CID` implementation... maybe they included that to support JSON.

I changed the test to:
```
func TestJSON(t *testing.T) {
	i := NewIDV1(Raw, 32)
	t.Logf("New ID: %s", i.String())

	data, err := json.Marshal(i)
	if err != nil {
		t.Errorf("failed to marshal ID %s: %s", i.String(), err)
	}
	t.Logf("JSON Encoded ID: %s", string(data))

	i2 := ID{}
	err = json.Unmarshal(data, &i2)
	if err != nil {
		t.Errorf("failed to unmarshal ID: %s", err)
	}

	if i.str != i2.str {
		t.Errorf("ID string %s != %s", i.str, i2.str)
	}
}
```

Which just uses the `json` package directly.

Comment (asutula) [reply core/thread/id.go:298]: Ah yea, brain is foggy today it seems. I think this has diverged from the original problem. I'll push a test that illustrates the original problem which has to do with json schemas that include thread ID.

File: api/client/client_test.go [hunk 2]
```diff
@@ -136,6 +137,38 @@ func TestClient_NewCollection(t *testing.T) {
 	})
 }
 
+func TestIt(t *testing.T) {
+	type Owner struct {
+		ID       core.InstanceID `json:"_id"`
+		ThreadID thread.ID       `json:"threadId"`
```
asutula, line 143: Ok @sanderpick back to the original problem that my original PR of making thread id a string alias would have fixed...

Here we have a model object with a `thread.ID` property. We're going to create a collection and try to save an instance of it.

api/client/client_test.go
```diff
@@ -136,6 +137,38 @@ func TestClient_NewCollection(t *testing.T) {
...
+	defer done()
+
+	id := thread.NewIDV1(thread.Raw, 32)
+	err := client.NewDB(context.Background(), id)
+	checkErr(t, err)
+	sch := util.SchemaFromInstance(Owner{}, false)
+	schBytes, err := json.MarshalIndent(sch, "", "  ")
+	checkErr(t, err)
+	schStr := string(schBytes)
+	t.Logf("\n%v", schStr)
```
asutula, line 157: Here is the output of this log statement, displaying the json schema string:
```
        {
          "$schema": "http://json-schema.org/draft-04/schema#",
          "$ref": "#/definitions/Owner",
          "definitions": {
            "ID": {
              "additionalProperties": false,
              "type": "object"
            },
            "Owner": {
              "required": [
                "_id",
                "threadId"
              ],
              "properties": {
                "_id": {
                  "type": "string"
                },
                "threadId": {
                  "$schema": "http://json-schema.org/draft-04/schema#",
                  "$ref": "#/definitions/ID"
                }
              },
              "additionalProperties": false,
              "type": "object"
            }
          }
        }
```
The `ID` type that represents thread id looks strange. It is an `object` but has no properties, presumably because the `str` property of `thread.ID` is internal/private.

api/client/client_test.go
```diff
@@ -136,6 +137,38 @@ func TestClient_NewCollection(t *testing.T) {
...
+	err = client.NewCollection(context.Background(), id, db.CollectionConfig{Name: collectionName, Schema: sch})
+	checkErr(t, err)
+
+	owner := &Owner{
+		ID:       "",
+		ThreadID: id,
+	}
+
+	_, err = client.Create(context.Background(), id, collectionName, Instances{owner})
```
asutula, line 166: When you try to save a `Owner` here, you get the error:
```
failed to create collection: rpc error: code = Unknown desc = instance doesn't correspond to schema
```

api/client/client_test.go
```diff
@@ -136,6 +137,38 @@ func TestClient_NewCollection(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create instance: %v", err)
```
asutula, line 168: When you try to save a `Owner` here, you get the error:
```
failed to create instance: rpc error: code = Unknown desc = instance doesn't correspond to schema
```

asutula: This is a bit mysterious, but it seems has something to do with the nested private `str string` inside `thread.ID` and the way that interacts with json schema and/or json marshaling/unmarshaling

api/client/client_test.go
sanderpick: Ah, I see. Okay, let me take a look here in a sec... juggling a few things ATM.

sanderpick:
👀

api/client/client_test.go
```diff
@@ -136,6 +137,50 @@ func TestClient_NewCollection(t *testing.T) {
...
+	defer done()
+
+	id := thread.NewIDV1(thread.Raw, 32)
+	err := client.NewDB(context.Background(), id)
+	checkErr(t, err)
+	sch := util.SchemaFromInstance(Owner{}, false, true)
+	schBytes, err := json.MarshalIndent(sch, "", "  ")
+	checkErr(t, err)
+	schStr := string(schBytes)
+	t.Logf("\n%v", schStr)
```
asutula, line 157: Ok here's the current output of the schema. The compromise/workaround here is that the schema must specify `additionalProperties=true` for any property that is a `thread.ID` or it must specify the `str` property inside a `properties` object. The library we use to create json schemas from an instance using reflection makes it easy to make `additionalProperties=true` for all properties, so that's what I'm doing for now. I'd like to find out of there's some way to force the presence of the `properties` and `str` property definition or at least only specify `additionalProperties=true` just for `thread.ID` properties, but it wasn't clear if that is possible with the lib we use.
```
        {
          "$schema": "http://json-schema.org/draft-04/schema#",
          "$ref": "#/definitions/Owner",
          "definitions": {
            "ID": {
              "additionalProperties": true,
              "type": "object"
            },
            "Owner": {
              "required": [
                "_id",
                "threadId"
              ],
              "properties": {
                "_id": {
                  "type": "string"
                },
                "threadId": {
                  "$schema": "http://json-schema.org/draft-04/schema#",
                  "$ref": "#/definitions/ID"
                }
              },
              "additionalProperties": true,
              "type": "object"
            }
          }
        }
```

core/thread/id.go
```diff
@@ -276,6 +290,13 @@ func (i ID) MarshalText() ([]byte, error) {
 	return []byte(i.String()), nil
 }
 
+func (i ID) MarshalJSON() ([]byte, error) {
+	data := map[string]string{
+		"str": i.String(),
+	}
+	return json.Marshal(data)
```
asutula, line 297: Other part of this solution is you actually have to return json data from `MarshalJSON` (i.e. a `map` in go). If you return bytes encoding some string value, the schema validation fails because the marshaled json you are validating includes a `string` value where the schema says the value must be an `object`. Returning bytes encoding a map here makes sure that is the case.

api/client/client_test.go
sanderpick: Hmm, I think I'm still leaning toward the string approach. This extra cruft doesn't buy us validation or anything like that, and makes the resulting JSON a bit weird, ie, `{"str": ID}`... that's a thread ID? Seems confusing to me.

asutula: Ok cool, just wanted to explore that to be sure. String it is.

sanderpick: Yeah totally, good to know how that's all working.

asutula:
This was resolved in https://github.com/textileio/go-threads/pull/346

PR #346, opened 2020-05-12
Copy of Make thread.ID just a string alias

sanderpick:
Sorry @asutula, I didn't grasp that the problem was related to JSON Schemas. So, adding this PR to reduce work. Feel free to take it over. 

Some notes:
- There doesn't seem to be a way to reflect a private attribute. 
- +1 to your original solution.
- This branch has been `reset --hard` to when before you reverted the change, plus a small nits I added. I wasn't able to easily cherry-pick your `better error message` commit, so I made the change manually. 

So, the downside to this, and why I was resistant, is that not there's nothing stopping me from doing `myID := ID("foo")`, which is not a valid `ID`. Calling `String()` on `myID` would panic if we weren't hardcoding the response from `Version()`, which we should probably not be doing anymore since now it's possible to create a version other than `V1`. Maybe we can add a `Valid()` or similar method that could be called to ensure `String()` won't panic when you're unsure.

### DIFF
--- a/core/thread/id.go
+++ b/core/thread/id.go
@@ -136,23 +136,10 @@ func ExtractEncoding(v string) (mbase.Encoding, error) {
 // Please use decode when parsing a regular ID string, as Cast does not
 // expect multibase-encoded data. Cast accepts the output of ID.Bytes().
 func Cast(data []byte) (ID, error) {
-	vers, n := binary.Uvarint(data)
-	if err := uvError(n); err != nil {
+	if err := validateIDData(data); err != nil {
 		return Undef, err
 	}
-
-	if vers != 1 {
-		return Undef, fmt.Errorf("expected 1 as the id version number, got: %d", vers)
-	}
-
-	_, cn := binary.Uvarint(data[n:])
-	if err := uvError(cn); err != nil {
-		return Undef, err
-	}
-
-	id := data[n+cn:]
-
-	return ID(data[0 : n+cn+len(id)]), nil
+	return ID(data), nil
 }
 
 // FromAddr returns ID from a multiaddress if present.
@@ -184,6 +171,47 @@ func uvError(read int) error {
 	}
 }
 
+func (i ID) Valid() bool {
+	data := i.Bytes()
+	err := validateIDData(data)
+	return err == nil
+}
+
+func getVersion(data []byte) (uint64, int, error) {
+	vers, n := binary.Uvarint(data)
+	if err := uvError(n); err != nil {
+		return 0, 0, err
+	}
+	return vers, n, nil
+}
+
+func validateIDData(data []byte) error {
+	vers, n, err := getVersion(data)
+	if err != nil {
+		return err
+	}
+
+	if vers != V1 {
+		return fmt.Errorf("expected 1 as the id version number, got: %d", vers)
+	}
+
+	variant, cn := binary.Uvarint(data[n:])
+	if err := uvError(cn); err != nil {
+		return err
+	}
+
+	if variant != uint64(Raw) && variant != uint64(AccessControlled) {
+		return fmt.Errorf("expected Raw or AccessControlled as the id variant, got: %d", variant)
+	}
+
+	id := data[n+cn:]
+	if len(id) == 0 {
+		return fmt.Errorf("expected random id bytes but there are none")
+	}
+
+	return nil
+}
+
 // UnmarshalBinary is equivalent to Cast(). It implements the
 // encoding.BinaryUnmarshaler interface.
 func (i *ID) UnmarshalBinary(data []byte) error {
@@ -208,7 +236,11 @@ func (i *ID) UnmarshalText(text []byte) error {
 
 // Version returns the ID version.
 func (i ID) Version() uint64 {
-	return V1
+	version, _, err := getVersion(i.Bytes())
+	if err != nil {
+		panic("error getting version: " + err.Error())
+	}
+	return version
 }
 
 // Variant returns the variant of an ID.
@@ -231,7 +263,7 @@ func (i ID) String() string {
 
 		return mbstr
 	default:
-		panic("not possible to reach this point")
+		panic("unknown thread id version")
 	}
 }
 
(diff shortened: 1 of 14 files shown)


sanderpick:
@asutula I updated the Android protoc installer action... seems to be good now.

asutula:
Awesome, thanks for doing that. I'm going to go through this one more time in the morning and make sure I also feel right about it... you got me second guessing myself haha. Mainly, I want to step down into some of the json schema code and understand why it thinks the instance doesn't match the schema in that test case. There's probably a good reason for it, but it isn't apparent to me just yet.

sanderpick:
Mongo's `ObjectID` could provide some inspiration: https://godoc.org/go.mongodb.org/mongo-driver/bson/primitive#ObjectID. Instead of a string or struct, it's a byte array. This has the same problem as I described above, where it allows a user to create an invalid ID. However, we can easily add some extra validation in the string decoding method, as they have done in [`ObjectIDFromHex`](https://godoc.org/go.mongodb.org/mongo-driver/bson/primitive#ObjectIDFromHex).

Some additional thoughts:
- Regarding validation, no matter what we do, the schema won't be able to really validate the ID, which requires something custom (maybe that doesn't matter, as a bad ID would/should cause hiccups in a user's app logic)
- In any case, https://github.com/qri-io/jsonschema#custom-validators looks interesting, although it would only be useful in the Go realm

sanderpick:
To clarify, I'm not suggesting with use a byte array, just that seeing their design makes me feel better about not "protecting" the inner string.

asutula:
I think I'm close to finding out if we can keep the original private string... think it might be possible. Stand by.

core/thread/id.go
```diff
@@ -136,23 +136,10 @@ func ExtractEncoding(v string) (mbase.Encoding, error) {
...
-
-	_, cn := binary.Uvarint(data[n:])
-	if err := uvError(cn); err != nil {
-		return Undef, err
-	}
-
-	id := data[n+cn:]
-
-	return ID{string(data[0 : n+cn+len(id)])}, nil
+	return ID(data), nil
```
asutula, line 142: We can simply validate the bytes then return an `ID`

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
 	}
 }
 
+func (i ID) Valid() bool {
```
asutula, line 174: Made this return `bool`, but I'm thinking `func (i ID) Validate() error` might be cleaner to use. Opinion?

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
...
+
+func validateIDData(data []byte) error {
+	vers, n, err := getVersion(data)
+	if err != nil {
+		return err
+	}
+
+	if vers != V1 {
+		return fmt.Errorf("expected 1 as the id version number, got: %d", vers)
+	}
```
asutula, line 196: First we get the version then make sure it is a known version (currently just `V1`)

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
...
+
+	variant, cn := binary.Uvarint(data[n:])
+	if err := uvError(cn); err != nil {
+		return err
+	}
+
+	if variant != uint64(Raw) && variant != uint64(AccessControlled) {
+		return fmt.Errorf("expected Raw or AccessControlled as the id variant, got: %d", variant)
+	}
```
asutula, line 205: Then we get the variant and make sure it's a known `Variant`

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
...
+
+	id := data[n+cn:]
+	if len(id) == 0 {
+		return fmt.Errorf("expected random id bytes but there are none")
+	}
```
asutula, line 210: Last we check that there are some random bytes

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
...
 }
 
 // Version returns the ID version.
 func (i ID) Version() uint64 {
-	return V1
+	version, _, err := getVersion(i.Bytes())
+	if err != nil {
+		panic("error getting version: " + err.Error())
+	}
+	return version
```
asutula, line 243: This returns a version `uint64` as long as a `uint64` of the right size was read from the data. It doesn't validate that it is a known version (that's what `Validate()` is for).

core/thread/id.go
```diff
@@ -184,84 +171,138 @@ func uvError(read int) error {
...
 }
 
 // Variant returns the variant of an ID.
 func (i ID) Variant() Variant {
-	_, n := uvarint(i.str)
-	variant, _ := uvarint(i.str[n:])
+	_, n := uvarint(string(i))
+	variant, _ := uvarint(string(i)[n:])
```
asutula, line 249: This seems like it could be similar to `Version()` and panic if there were an error reading the variant byte, but it uses this `uvarint()` method I don't really understand. Just tell me if we should leave this or make it more like `Version()`.

This also doesn't verify it's a known `Variant`, but calling `Valid()` does check that.

core/thread/id.go
sanderpick: Yeah, I guess error would allow for finer-grained feedback

asutula:
Ok @sanderpick I pushed some changes here.

- Adds `Valid() bool` method to `thread.ID`
- Refactors code in `thread.ID` to be DRY as possible
- Two methods on `thread.ID` (`String()` and `Version()`) might panic if they encounter an unsupported state because of an invalid `thread.ID`, so I updated all production calls (not tests) to call `thread.ID.Valid()` and handle the invalid case before calling either of those

Left some comments and questions in the code. Let me know what you think.

core/thread/id.go
sanderpick: Yeah let's just leave this for now. Maybe we can revisit when adding the ACL stuff to ID.

net/net.go
```diff
@@ -276,6 +276,10 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
 		}
 	}
 
+	if !id.Valid() {
```
sanderpick, line 279: Should this be checked at the top of these public API methods?

asutula: Yea good call. So in this file, any method that accepts or derives a thread id from data coming from the api user, I'll call `Validate`

PR #362, opened 2020-05-27
Collection and index updates and fixes

### LINKED ISSUE #291: Check that index fields are valid schema fields
Here is the original todo: https://github.com/textileio/go-threads/blob/master/db/db.go#L232

### LINKED ISSUE #359: Allow collections to be updated

### LINKED ISSUE #360: Add a remove collection method

### LINKED ISSUE #361: Validate unique index is possible
When a unique index is created at some time after collection creation, a full scan will have to be run in order to ensure that it's actually unique.

sanderpick:
Closes #291 
Closes #359 
Closes #360 
Closes #361

### DB
- Adds `UpdateCollection` and associated API+client methods
- Adds `DeleteCollection` and associated API+client methods
- Adds `GetCollectionIndexes` and associated API+client methods
- Adds `dropIndex` to `Collection` method
- Adds index validation:
  - Index path must hit a valid type in the schema
  - Supported types: string, number, integer, boolean
  - For unique indexes, ensure collection does not contain multiple instances with the same value at path
- Better schema validation errors

----

Some things to consider:
- Updating a collection schema _could_ cause old data to be unreadable if the particular API doesn't have access to the raw JSON results. This can happen when I field is renamed.
- Along those lines, there's still work to do here on _migrations_. A migration config could go along with a collection update request, which would handle field re-mapping. [Realm](https://realm.io/docs/swift/latest/#local-migrations) could be some inspiration here: https://github.com/textileio/go-threads/issues/363
- Back-filling / re-indexing is still not a thing: https://github.com/textileio/go-threads/issues/286


api/client/client.go
```diff
@@ -274,6 +273,59 @@ func (c *Client) NewCollection(ctx context.Context, dbID thread.ID, config db.Co
...
+	}
+	_, err := c.c.DeleteCollection(ctx, &pb.DeleteCollectionRequest{
+		DbID: dbID.Bytes(),
+		Name: name,
+	})
+	return err
+}
+
+// GetCollectionIndexes returns an existing collection's indexes.
+func (c *Client) GetCollectionIndexes(ctx context.Context, dbID thread.ID, name string, opts ...db.ManagedOption) ([]db.Index, error) {
```
carsonfarmer, line 307: Ok, so we'll want to expose this on the remote clients as well then huh?

api/client/client_test.go
```diff
@@ -136,6 +136,83 @@ func TestClient_NewCollection(t *testing.T) {
...
+
+	id := thread.NewIDV1(thread.Raw, 32)
+	err := client.NewDB(context.Background(), id)
+	checkErr(t, err)
+	err = client.NewCollection(context.Background(), id, db.CollectionConfig{Name: collectionName, Schema: util.SchemaFromSchemaString(schema)})
+	checkErr(t, err)
+
+	t.Run("test update collection", func(t *testing.T) {
+		err = client.UpdateCollection(context.Background(), id, db.CollectionConfig{
+			Name:   collectionName,
```
carsonfarmer, line 152: Nice, so looks like anything in collectionconfig is fair game for updating. Super.

db/collection_test.go
```diff
@@ -111,32 +164,251 @@ func TestNewCollection(t *testing.T) {
 			t.Fatal("the collection name should be invalid")
 		}
 	})
+	t.Run("Fail/BadIndexPath", func(t *testing.T) {
```
carsonfarmer, line 167: Nice

db/collection_test.go
```diff
@@ -111,32 +164,251 @@ func TestNewCollection(t *testing.T) {
...
+		checkErr(t, err)
+	})
+	t.Run("AddFieldsAndIndexes", func(t *testing.T) {
+		t.Parallel()
+		db, clean := createTestDB(t)
+		defer clean()
+		_, err := db.NewCollection(CollectionConfig{
+			Name:    "Dog",
+			Schema:  util.SchemaFromInstance(&Dog{}, false),
+			Indexes: []Index{{Path: "Name", Unique: true}},
```
carsonfarmer, line 227: Nice API

db/db.go
```diff
@@ -168,154 +169,193 @@ func newDB(n app.Net, id thread.ID, options *NewDBOptions) (*DB, error) {
...
+		}
+		var indexes map[string]Index
 		index, err := d.datastore.Get(dsDBIndexes.ChildString(name))
 		if err == nil && index != nil {
-			_ = json.Unmarshal(index, &indexes)
-		}
-
-		indexValues := make([]IndexConfig, len(indexes))
-		for _, value := range indexes {
-			indexValues = append(indexValues, value)
```
sanderpick, line 201: This was a bug that resulted in the values being appended _to the end_ of the slice which already has length > 0. Looks like it was resulting in some blank indexes being persisted... not too harmful, but something to keep an eye on when we reload an existing repo with the new code.

db/collection_test.go
```diff
@@ -111,32 +164,251 @@ func TestNewCollection(t *testing.T) {
...
 		})
-		t.Run("AddIDIndex", func(t *testing.T) {
-			err := collection.AddIndex(IndexConfig{Path: "ID", Unique: true})
-			checkErr(t, err)
+		checkErr(t, err)
+		c, err := db.UpdateCollection(CollectionConfig{
+			Name:    "Dog",
+			Schema:  util.SchemaFromInstance(&Dog2{}, false),
+			Indexes: []Index{{Path: "FullName", Unique: true}, {Path: "Toys.Favorite"}},
```
carsonfarmer, line 233: Love the symmetry

db/collection_test.go
```diff
@@ -146,36 +418,36 @@ func TestCreateInstance(t *testing.T) {
...
-			assertPersonInCollection(t, collection, newPerson)
+			assertPersonInCollection(t, c, newPerson)
 		})
 	})
 	t.Run("Multiple", func(t *testing.T) {
 		t.Parallel()
 		db, clean := createTestDB(t)
 		defer clean()
-		collection, err := db.NewCollection(CollectionConfig{
+		c, err := db.NewCollection(CollectionConfig{
```
carsonfarmer, line 450: I've been seeing a lot of these changes lately, what is the reasoning behind it? Aesthetics? Go best practices?

db/db.go
```diff
@@ -148,7 +149,7 @@ func newDB(n app.Net, id thread.ID, options *NewDBOptions) (*DB, error) {
 		datastore:           options.Datastore,
 		dispatcher:          newDispatcher(options.Datastore),
 		eventcodec:          options.EventCodec,
-		collectionNames:     make(map[string]*Collection),
+		collections:         make(map[string]*Collection),
```
carsonfarmer, line 152: Nice

db/db.go
carsonfarmer: Good catch!

sanderpick: Okay, playing on the safe side... I added a check here during re-hydration: https://github.com/textileio/go-threads/pull/362/commits/014b3363ade213f843cb54bbe70a3421a35fd26b

db/index.go
```diff
@@ -11,120 +11,213 @@ import (
...
-	Indexes() map[string]Index //[indexname]indexFunc
-}
+	indexPrefix = ds.NewKey("_index")
+	indexTypes  = []string{"string", "number", "integer", "boolean"}
+)
 
-// Index is a function that returns the indexable, encoded bytes of the passed in bytes
+// Index defines an index.
 type Index struct {
-	IndexFunc func(name string, value []byte) (ds.Key, error)
```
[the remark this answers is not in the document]
carsonfarmer: Fair enough, though it would be nice to be able to index certain field types differently? I guess we could encompass that in a single function still...

db/index.go
```diff
@@ -11,120 +11,213 @@ import (
...
+	return c.saveIndexes()
 }
 
-// removes an item from the index
-// be sure to pass the data from the old record, not the new one
-func indexDelete(indexer Indexer, tx ds.Txn, key ds.Key, originalData []byte) error {
-	indexes := indexer.Indexes()
+// dropIndex drops the index at path.
+// @todo: Handle token auth
+func (c *Collection) dropIndex(pth string, opts ...Option) error {
```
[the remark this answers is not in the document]
carsonfarmer: Yes, I quite like this.

api/service.go
```diff
@@ -288,6 +288,74 @@ func (s *Service) NewCollection(ctx context.Context, req *pb.NewCollectionReques
...
+	token, err := thread.NewTokenFromMD(ctx)
+	if err != nil {
+		return nil, err
+	}
+	collection, err := s.getCollection(ctx, req.Name, id, token)
+	if err != nil {
+		return nil, err
+	}
+	indexes := collection.GetIndexes()
+	var pbindexes []*pb.Index
```
jsign, line 347: Nit: preallocate slice.

db/collection.go
```diff
@@ -4,22 +4,25 @@ import (
...
 	"github.com/textileio/go-threads/core/thread"
-	"github.com/tidwall/gjson"
 	"github.com/xeipuuv/gojsonschema"
 )
 
 var (
-	// ErrNotFound indicates that the specified instance doesn't
-	// exist in the collection.
-	ErrNotFound = errors.New("instance not found")
+	// ErrInvalidSchemaPath indicates path does not resolve to a schema type.
```
jsign, line 18: `s/ErrInvalidSchemaPath/ErrInvalidCollectionSchemaPath`

db/db.go
```diff
@@ -168,154 +169,195 @@ func newDB(n app.Net, id thread.ID, options *NewDBOptions) (*DB, error) {
...
+		return nil, err
 	}
+	return c, nil
+}
+
+// UpdateCollection updates an existing db collection with a new config.
+// Indexes to new paths will be created.
+// Indexes to removed paths will be dropped.
+// @todo: Handle token auth
+func (d *DB) UpdateCollection(config CollectionConfig, opts ...Option) (*Collection, error) {
```
jsign, line 268: Nit: possible improvement for the future, validate all actions, then do actions. Just to minimize the possibility of partial applied changes. Not sure if it may worth it since the client could retry knowing with the "remaining" part (which might be a bit hard to know which one, but anyway).

db/db.go
```diff
@@ -168,154 +169,195 @@ func newDB(n app.Net, id thread.ID, options *NewDBOptions) (*DB, error) {
...
+	c, ok := d.collections[name]
+	if !ok {
+		return ErrCollectionNotFound
+	}
+	if err := d.datastore.Delete(dsDBIndexes.ChildString(c.name)); err != nil {
+		return err
+	}
+	if err := d.datastore.Delete(dsDBSchemas.ChildString(c.name)); err != nil {
+		return err
+	}
```
jsign, line 351: Nit: could leverage datastore txn.

api/service.go
sanderpick: `indexes` is a map. I suppose a count could be tracked in addition the for loop that is used to correctly assign indexes. Will do.

sanderpick: Hmm, better yet, make `GetIndexes` return a slice.

api/client/client_test.go
```diff
@@ -136,6 +136,83 @@ func TestClient_NewCollection(t *testing.T) {
...
+		checkErr(t, err)
+
+		err = client.DeleteCollection(context.Background(), id, collectionName)
+		if err != nil {
+			t.Fatalf("failed to delete collection: %v", err)
+		}
+		_, err = client.Create(context.Background(), id, collectionName, Instances{createPerson()})
+		if err == nil {
+			t.Fatal("failed to delete collection")
+		}
```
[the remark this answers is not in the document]
sanderpick: There's also tests on the internal collection, so a bit light here :)

api/service.go
jsign: @sanderpick , `len(indexes)` works for maps too.

db/db.go
sanderpick: Nice yeah, I had a similar thought. I'll create an issue.

api/service.go
sanderpick: Right, I'm just saying you wouldn't be able to do the normal pattern `indexes[i] = `, since iterating a map doesn't reveal an index. But tracking count would allow that, just a little extra code.

api/client/client.go
sanderpick: Yep yep

api/service.go
jsign: Oh, I see what you mean. 👍

db/collection_test.go
sanderpick: I just like the shorter names where readability isn't paramount

db/index.go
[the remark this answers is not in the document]
sanderpick: Yep yep

db/db.go
sanderpick: Good call, will do

sanderpick:
Thanks for the reviews guys 🍺

PR #372, opened 2020-06-09
Allow specifying which log to replicate from clients

carsonfarmer:
This is the simplest solution to ensuring that a (remote) client can request a (remote) daemon/peer to replicate a log that it does not 'own' (in the sense of having access to the private key). By requiring the client to specify _which_ log to replicate, we allow cases where there are multiple clients using the remote daemon/peer to 'proxy' their logs, without actually sharing their private LogKey with the remote. This minimal change avoids adding any new concepts at the net level, and maintains backward compatibility with the existing behavior (defaulting to the daemon's own log). Not that AddReplicator is the only place that this behavior (getting the daemons own log via the private key) exists. The only other place that the private key 'ownership' is checked is in `thread.Info.GetOwnLog`, but its usage is limited to app usage, which _is_ entirely appropriate.


core/net/net.go
```diff
@@ -53,7 +53,7 @@ type API interface {
 
 	// AddReplicator with credentials.
 	// The thread service key and all records will be pushed to paddr.
-	AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiaddr, opts ...ThreadOption) (peer.ID, error)
+	AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiaddr, lid peer.ID, opts ...ThreadOption) (peer.ID, error)
```
carsonfarmer, line 56: I also considered having this as a TheadOption, but from a client's perspective, this felt messy. It was nicer for 'optional' behavior, but I think the explicit (which log do you want to replicate) behavior is more intuitive

examples/chat/main.go
```diff
@@ -522,7 +522,7 @@ func addReplicatorCmd(id thread.ID, addrStr string) (out string, err error) {
 	if err != nil {
 		return
 	}
-	pid, err := net.AddReplicator(ctx, id, addr)
+	pid, err := net.AddReplicator(ctx, id, addr, peer.ID(""))
```
carsonfarmer, line 525: Using an invalid peer id/log id like this just defaults to the current behavior.

net/api/client/client.go
```diff
@@ -197,15 +197,17 @@ func (c *Client) DeleteThread(ctx context.Context, id thread.ID, opts ...core.Th
 	return err
 }
 
-func (c *Client) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiaddr, opts ...core.ThreadOption) (pid peer.ID, err error) {
+func (c *Client) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiaddr, lid peer.ID, opts ...core.ThreadOption) (pid peer.ID, err error) {
 	args := &core.ThreadOptions{}
 	for _, opt := range opts {
 		opt(args)
 	}
+	lidb, _ := lid.Marshal()
```
carsonfarmer, line 205: If this is invalid, that's ok...

net/api/service.go
```diff
@@ -252,11 +252,13 @@ func (s *Service) AddReplicator(ctx context.Context, req *pb.AddReplicatorReques
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// This can produce an invalid Log ID, which is ok...
+	lid, _ := peer.IDFromBytes(req.LogID)
```
carsonfarmer, line 256: We can avoid the error check here in favor of allowing invalid log ids later on down the line

net/net.go
```diff
@@ -1037,15 +1044,17 @@ func (n *net) getOwnLog(id thread.ID) (info thread.LogInfo, err error) {
 			return n.store.GetLog(id, lid)
 		}
 	}
-	return info, nil
+	return info, lstore.ErrLogNotFound
```
carsonfarmer, line 1047: This felt safer (thanks to discussion with @jsign), and doesn't break anything else... though without it, we _can_ end up in a state where we have an invalid log (log that is just the zero state).

PR #374, opened 2020-06-10
Flag locally created logs as direct

### LINKED ISSUE #373: Invalid log information from getOwnLog if peer doesn't 'own' the log
Creating a Thread on a (remote) peer/daemon doesn't require the client to provide the private key for it to be created. This is great, as it supports using the remote daemon/peer as a 'proxy' peer for doing 'trustless' log updates for a given remote client peer. Unfortunately, some of our existing internal logic assumes 'ownership' of a log is represented by having the private key. This is a reasonable assumption in many cases, but doesn't support the use case listed above. To remedy this, some more general concept of 'ownership' should be considered. In practice, this might look like 'flagging' logs created by the daemon/peer directly with a 'Direct' bool flag, indicating that this log is managed directly by this peer. To keep backward compatibility, we could still check for 'ownership' in the strict sense via the existence of a private key.

carsonfarmer:
Fixes #373, and a better alternative to #372.

The approach here is to simply take advantage of the existing metadata store on the logstore to set a flag as to whether a given log is "managed?" by the local peer. Directly "managed" logs are those logs created or directly added by the local peer. Otherwise, unmanaged logs are those from other peers in a given thread. Related, "owned" logs are those logs for which the peer has the private key. There can only be _one_ "owned" log, versus potentially many "managed" logs.

This should support multiple use cases:
* The local peer is the client peer. One "owned" log is used to interact with a given thread (single-tenant, single-writer setup)
* The local peer is (also) a relay peer. Maybe multiple "managed" logs on behalf of other clients, and zero or one "owned" log to also participate in the thread on its own (multi-tenant, multi-writer setup)
* The local peer is (also) a proxy-writer. Maybe multiple "managed" logs on behalf of other clients, and one "owned" log that is written to on behalf of other 'agents' (multi-tenant, single-writer setup)

### DIFF
--- a/core/logstore/logstore.go
+++ b/core/logstore/logstore.go
@@ -22,6 +22,9 @@ var ErrThreadNotFound = fmt.Errorf("thread not found")
 // ErrLogNotFound indicates a requested log was not found.
 var ErrLogNotFound = fmt.Errorf("log not found")
 
+// ErrLogExists indicates a requested log already exists.
+var ErrLogExists = fmt.Errorf("log already exists")
+
 // Logstore stores log keys, addresses, heads and thread meta data.
 type Logstore interface {
 	Close() error
@@ -49,8 +52,8 @@ type Logstore interface {
 	// GetLog returns info about a log.
 	GetLog(thread.ID, peer.ID) (thread.LogInfo, error)
 
-	// GetOwnLogs returns info about locally managed logs.
-	GetOwnLogs(thread.ID) ([]thread.LogInfo, error)
+	// GetManagedLogs returns info about locally managed logs.
+	GetManagedLogs(thread.ID) ([]thread.LogInfo, error)
 
 	// DeleteLog deletes a log.
 	DeleteLog(thread.ID, peer.ID) error
--- a/core/logstore/logstore.go
+++ b/core/logstore/logstore.go
@@ -49,6 +49,9 @@ type Logstore interface {
 	// GetLog returns info about a log.
 	GetLog(thread.ID, peer.ID) (thread.LogInfo, error)
 
+	// GetOwnLogs returns info about locally managed logs.
+	GetOwnLogs(thread.ID) ([]thread.LogInfo, error)
+
 	// DeleteLog deletes a log.
 	DeleteLog(thread.ID, peer.ID) error
 }
@@ -61,11 +64,17 @@ type ThreadMetadata interface {
 	// PutInt64 stores an int value under key.
 	PutInt64(t thread.ID, key string, val int64) error
 
-	// GetString retrieves an int value under key.
+	// GetString retrieves a string value under key.
 	GetString(t thread.ID, key string) (*string, error)
 
 	// PutString stores a string value under key.
 	PutString(t thread.ID, key string, val string) error
+
+	// GetBool retrieves a boolean value under key.
+	GetBool(t thread.ID, key string) (*bool, error)
+
+	// PutBool stores a boolean value under key.
+	PutBool(t thread.ID, key string, val bool) error
 
 	// GetBytes retrieves a byte value under key.
 	GetBytes(t thread.ID, key string) (*[]byte, error)
--- a/core/thread/id.go
+++ b/core/thread/id.go
@@ -366,10 +366,16 @@ func (i Info) GetOwnLog() *LogInfo {
 
 // LogInfo holds log keys, addresses, and heads.
 type LogInfo struct {
-	ID      peer.ID
-	PubKey  crypto.PubKey
+	// ID is the log's identifier.
+	ID peer.ID
+	// PubKey is the log's public key.
+	PubKey crypto.PubKey
+	// PrivKey is the log's private key.
 	PrivKey crypto.PrivKey
-	Addrs   []ma.Multiaddr
-	Head    cid.Cid
-	Direct  bool // Whether log is 'directly' managed (owned). Indirect logs are 'peer' logs.
+	// Addrs are the addresses associated with the given log.
+	Addrs []ma.Multiaddr
+	// Head is the log's current head.
+	Head cid.Cid
+	// Managed logs are any logs directly added/created by the host, and/or logs for which we have the private key
+	Managed bool
 }
--- a/core/thread/id.go
+++ b/core/thread/id.go
@@ -110,7 +110,7 @@ func Decode(v string) (ID, error) {
 	return Cast(data)
 }
 
-// Extract the encoding from an ID. If Decode on the same string did
+// ExtractEncoding from an ID. If Decode on the same string did
 // not return an error neither will this function.
 func ExtractEncoding(v string) (mbase.Encoding, error) {
 	if len(v) < 2 {
@@ -171,6 +171,7 @@ func uvError(read int) error {
 	}
 }
 
+// Validate the ID.
 func (i ID) Validate() error {
 	data := i.Bytes()
 	return validateIDData(data)
@@ -269,7 +270,7 @@ func (i ID) String() string {
 	}
 }
 
-// String returns the string representation of an ID
+// StringOfBase returns the string representation of an ID
 // encoded is selected base.
 func (i ID) StringOfBase(base mbase.Encoding) (string, error) {
 	if err := i.Validate(); err != nil {
@@ -353,6 +354,7 @@ type Info struct {
 }
 
 // GetOwnLog returns the first log found with a private key.
+// This is a strict owership check, vs returning all directly 'managed' logs.
 func (i Info) GetOwnLog() *LogInfo {
 	for _, lg := range i.Logs {
 		if lg.PrivKey != nil {
@@ -369,4 +371,5 @@ type LogInfo struct {
 	PrivKey crypto.PrivKey
 	Addrs   []ma.Multiaddr
 	Head    cid.Cid
+	Direct  bool // Whether log is 'directly' managed (owned). Indirect logs are 'peer' logs.
 }
--- a/db/options.go
+++ b/db/options.go
@@ -42,6 +42,7 @@ type NewOptions struct {
 	EventCodec  core.EventCodec
 	LowMem      bool
 	Debug       bool
+	ThreadKey   thread.Key
 }
 
 // NewOption specifies a new db option.
@@ -65,6 +66,13 @@ func WithNewRepoPath(path string) NewOption {
 func WithNewToken(t thread.Token) NewOption {
 	return func(o *NewOptions) {
 		o.Token = t
+	}
+}
+
+// WithNewThreadKey provides control over thread keys to use with a db.
+func WithNewThreadKey(key thread.Key) NewOption {
+	return func(o *NewOptions) {
+		o.ThreadKey = key
 	}
 }
 
--- a/logstore/logstore.go
+++ b/logstore/logstore.go
@@ -203,6 +203,11 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
 			return err
 		}
 	}
+	if lg.Direct {
+		if err = ls.PutBool(id, lg.ID.Pretty(), true); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
@@ -234,6 +239,13 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err 
 	if err != nil {
 		return
 	}
+	direct, err := ls.GetBool(id, lid.Pretty())
+	if err != nil {
+		return
+	}
+	if direct != nil {
+		info.Direct = *direct
+	}
 	info.ID = lid
 	info.PubKey = pk
 	info.PrivKey = sk
@@ -242,6 +254,27 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err 
 		info.Head = heads[0]
 	}
 	return
+}
+
+// GetOwnLogs returns the logs the host is directly managing under the given thread.
+func (ls *logstore) GetOwnLogs(id thread.ID) ([]thread.LogInfo, error) {
+	logs, err := ls.LogsWithKeys(id)
+	if err != nil {
+		return nil, err
+	}
+	var direct []thread.LogInfo
+	for _, lid := range logs {
+		lg, err := ls.GetLog(id, lid)
+		if err != nil {
+			return nil, err
+		}
+		// We are 'managing' any direct logs or logs for which we have the private key
+		if lg.Direct || lg.PrivKey != nil {
+			direct = append(direct, lg)
+			continue
+		}
+	}
+	return direct, nil
 }
 
 // DeleteLog deletes a log.
--- a/logstore/logstore.go
+++ b/logstore/logstore.go
@@ -1,6 +1,7 @@
 package logstore
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"sync"
@@ -91,12 +92,34 @@ func (ls *logstore) AddThread(info thread.Info) error {
 	if info.Key.Service() == nil {
 		return fmt.Errorf("a service-key is required to add a thread")
 	}
-	if err := ls.AddServiceKey(info.ID, info.Key.Service()); err != nil {
+	sk, err := ls.ServiceKey(info.ID)
+	if err != nil {
 		return err
 	}
-	if info.Key.CanRead() {
-		if err := ls.AddReadKey(info.ID, info.Key.Read()); err != nil {
+	if sk == nil {
+		if err := ls.AddServiceKey(info.ID, info.Key.Service()); err != nil {
 			return err
+		}
+	} else {
+		// Ensure keys are the same
+		if !bytes.Equal(info.Key.Service().Bytes(), sk.Bytes()) {
+			return fmt.Errorf("service-key mismatch")
+		}
+	}
+	if info.Key.CanRead() {
+		rk, err := ls.ReadKey(info.ID)
+		if err != nil {
+			return err
+		}
+		if rk == nil {
+			if err := ls.AddReadKey(info.ID, info.Key.Read()); err != nil {
+				return err
+			}
+		} else {
+			// Ensure keys are the same
+			if !bytes.Equal(info.Key.Read().Bytes(), rk.Bytes()) {
+				return fmt.Errorf("read-key mismatch")
+			}
 		}
 	}
 	return nil
--- a/logstore/logstore.go
+++ b/logstore/logstore.go
@@ -11,6 +11,8 @@ import (
 	"github.com/textileio/go-threads/core/thread"
 )
 
+var managedSuffix = "/managed"
+
 // logstore is a collection of books for storing thread logs.
 type logstore struct {
 	sync.RWMutex
@@ -186,14 +188,17 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
 	ls.Lock()
 	defer ls.Unlock()
 
+	if lg.PrivKey != nil {
+		if pk, _ := ls.PrivKey(id, lg.ID); pk != nil {
+			return core.ErrLogExists
+		}
+		if err := ls.AddPrivKey(id, lg.ID, lg.PrivKey); err != nil {
+			return err
+		}
+	}
 	err := ls.AddPubKey(id, lg.ID, lg.PubKey)
 	if err != nil {
 		return err
-	}
-	if lg.PrivKey != nil {
-		if err = ls.AddPrivKey(id, lg.ID, lg.PrivKey); err != nil {
-			return err
-		}
 	}
 	if err = ls.AddAddrs(id, lg.ID, lg.Addrs, pstore.PermanentAddrTTL); err != nil {
 		return err
@@ -203,8 +208,9 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
 			return err
 		}
 	}
-	if lg.Direct {
-		if err = ls.PutBool(id, lg.ID.Pretty(), true); err != nil {
+	// By definition 'owned' logs are also 'managed' logs.
+	if lg.Managed || lg.PrivKey != nil {
+		if err = ls.PutBool(id, lg.ID.Pretty()+managedSuffix, true); err != nil {
 			return err
 		}
 	}
@@ -239,12 +245,12 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err 
 	if err != nil {
 		return
 	}
-	direct, err := ls.GetBool(id, lid.Pretty())
+	managed, err := ls.GetBool(id, lid.Pretty()+managedSuffix)
 	if err != nil {
 		return
 	}
-	if direct != nil {
-		info.Direct = *direct
+	if managed != nil {
+		info.Managed = *managed
 	}
 	info.ID = lid
 	info.PubKey = pk
@@ -256,25 +262,24 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err 
 	return
 }
 
-// GetOwnLogs returns the logs the host is directly managing under the given thread.
-func (ls *logstore) GetOwnLogs(id thread.ID) ([]thread.LogInfo, error) {
+// GetManagedLogs returns the logs the host is 'managing' under the given thread.
+func (ls *logstore) GetManagedLogs(id thread.ID) ([]thread.LogInfo, error) {
 	logs, err := ls.LogsWithKeys(id)
 	if err != nil {
 		return nil, err
 	}
-	var direct []thread.LogInfo
+	var managed []thread.LogInfo
 	for _, lid := range logs {
 		lg, err := ls.GetLog(id, lid)
 		if err != nil {
 			return nil, err
 		}
-		// We are 'managing' any direct logs or logs for which we have the private key
-		if lg.Direct || lg.PrivKey != nil {
-			direct = append(direct, lg)
+		if lg.Managed || lg.PrivKey != nil {
+			managed = append(managed, lg)
 			continue
 		}
 	}
-	return direct, nil
+	return managed, nil
 }
 
 // DeleteLog deletes a log.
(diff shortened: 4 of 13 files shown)


core/logstore/logstore.go
```diff
@@ -49,6 +49,9 @@ type Logstore interface {
 	// GetLog returns info about a log.
 	GetLog(thread.ID, peer.ID) (thread.LogInfo, error)
 
+	// GetOwnLogs returns info about locally managed logs.
+	GetOwnLogs(thread.ID) ([]thread.LogInfo, error)
```
carsonfarmer, line 53: New top-level API on logstore. Only used in one place for now, but could be handy. Could alternatively be called something else, but this seems simple enough.

core/logstore/logstore.go
```diff
@@ -61,12 +64,18 @@ type ThreadMetadata interface {
...
 
-	// GetString retrieves an int value under key.
+	// GetString retrieves a string value under key.
 	GetString(t thread.ID, key string) (*string, error)
 
 	// PutString stores a string value under key.
 	PutString(t thread.ID, key string, val string) error
 
+	// GetBool retrieves a boolean value under key.
+	GetBool(t thread.ID, key string) (*bool, error)
```
carsonfarmer, line 74: Adding here as a useful 'flag' API, but a more dedicated/bespoke API could be used instead to create the 'direct' flag elsewhere?

core/thread/id.go
```diff
@@ -110,7 +110,7 @@ func Decode(v string) (ID, error) {
 	return Cast(data)
 }
 
-// Extract the encoding from an ID. If Decode on the same string did
+// ExtractEncoding from an ID. If Decode on the same string did
```
carsonfarmer, line 113: Linter complaints 🤷

logstore/logstore.go
```diff
@@ -234,6 +239,13 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
 	if err != nil {
 		return
 	}
+	direct, err := ls.GetBool(id, lid.Pretty())
```
carsonfarmer, line 242: This is maybe not as 'safe' as having a dedicated Direct flag API (because we're just using the log id as the generic key here). I could be convinced to create a more bespoke API for the Direct flag if that seems more appropriate?

logstore/logstore.go
```diff
@@ -244,6 +256,27 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
 	return
 }
 
+// GetOwnLogs returns the logs the host is directly managing under the given thread.
+func (ls *logstore) GetOwnLogs(id thread.ID) ([]thread.LogInfo, error) {
```
carsonfarmer, line 260: Moved here from `net/net.go`

logstore/logstore.go
```diff
@@ -244,6 +256,27 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
...
+		return nil, err
+	}
+	var direct []thread.LogInfo
+	for _, lid := range logs {
+		lg, err := ls.GetLog(id, lid)
+		if err != nil {
+			return nil, err
+		}
+		// We are 'managing' any direct logs or logs for which we have the private key
+		if lg.Direct || lg.PrivKey != nil {
```
carsonfarmer, line 272: We return the log if its directly managed by this peer, or if we have the private key. In theory, if private key is available, then it _should_ also be direct.

net/net.go
```diff
@@ -534,12 +534,14 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
...
 		return
 	}
-	ownlg, err := n.getOwnLog(info.ID)
+	logs, err := n.store.GetOwnLogs(info.ID)
 	if err != nil {
 		return
 	}
-	if err = n.store.AddAddr(info.ID, ownlg.ID, addr, pstore.PermanentAddrTTL); err != nil {
-		return
+	for _, lg := range logs {
```
carsonfarmer, line 541: Send out all directly managed logs.

net/net.go
```diff
@@ -557,14 +559,17 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
...
 		if err = n.server.pushLog(ctx, info.ID, l, pid, info.Key.Service(), nil); err != nil {
-			if err := n.store.SetAddrs(info.ID, ownlg.ID, ownlg.Addrs, pstore.PermanentAddrTTL); err != nil {
-				log.Errorf("error rolling back log address change: %s", err)
+			// Rollback all potentially failed log(s)
+			for _, lg := range logs {
+				if err := n.store.SetAddrs(info.ID, lg.ID, lg.Addrs, pstore.PermanentAddrTTL); err != nil {
+					log.Errorf("error rolling back log address change: %s", err)
+				}
 			}
 			return
```
carsonfarmer, line 568: Noticed this early out here. Do we want to fail and roll-back if _any_ of the logs fail to push?

net/net.go
```diff
@@ -1022,37 +1032,22 @@ func (n *net) startPulling() {
 	}
 }
 
-// getOwnLoad returns the log owned by the host under the given thread.
-func (n *net) getOwnLog(id thread.ID) (info thread.LogInfo, err error) {
```
carsonfarmer, line 1026: moved to logstore

net/net.go
```diff
@@ -1022,37 +1032,22 @@ func (n *net) startPulling() {
...
-		if sk != nil {
-			return n.store.GetLog(id, lid)
-		}
-	}
-	return info, nil
-}
-
-// getOrCreateOwnLoad returns the log owned by the host under the given thread.
-// If no log exists, a new one is created under the given thread.
+// getOrCreateOwnLoad returns the first log 'owned' by the host under the given thread.
```
carsonfarmer, line 1035: Note the updated docs here.

net/api/pb/threadsnet.proto
```diff
@@ -49,6 +49,7 @@ message LogInfo {
     bytes privKey = 3;
     repeated bytes addrs = 4;
     bytes head = 5;
+    bool direct = 6;
```
carsonfarmer, line 52: Went with direct here, that way the zero value is false, making it backward compatible.

core/thread/id.go
```diff
@@ -369,4 +371,5 @@ type LogInfo struct {
 	PrivKey crypto.PrivKey
 	Addrs   []ma.Multiaddr
 	Head    cid.Cid
+	Direct  bool // Whether log is 'directly' managed (owned). Indirect logs are 'peer' logs.
```
carsonfarmer, line 374: Went with Direct here, that way the zero value is false, making it backward compatible.

core/thread/id.go
```diff
@@ -353,6 +354,7 @@ type Info struct {
 }
 
 // GetOwnLog returns the first log found with a private key.
+// This is a strict owership check, vs returning all directly 'managed' logs.
```
carsonfarmer, line 357: Updated docs here, but leaving the behavior the same. Essentially, this is a more 'strict' check on ownership, where you _do_ need to the have private key locally.

net/net.go
```diff
@@ -1022,37 +1032,22 @@ func (n *net) startPulling() {
...
+// If no log exists, a new one is created under the given thread and returned.
+// This is a strict 'owership' check, vs returning the first directly 'managed' log.
 func (n *net) getOrCreateOwnLog(id thread.ID) (info thread.LogInfo, err error) {
-	info, err = n.getOwnLog(id)
+	lgs, err := n.store.GetOwnLogs(id)
 	if err != nil {
 		return info, err
 	}
-	if info.PubKey != nil {
```
carsonfarmer, line 1050: Is this right? PubKey here? Seems like it should have been PrivKey (see update), but its possible I'm missing something and PubKey is sufficient.

core/logstore/logstore.go
sanderpick: `GetDirectLogs`?

core/logstore/logstore.go
sanderpick: SGTM!

core/thread/id.go
sanderpick: Maybe I'll see below, but is this still needed? Now that there _could_ be multiple private keys, this could start returning different things at different times.

core/thread/id.go
sanderpick: Sweet

logstore/logstore.go
sanderpick: What about making the key `<logID>/direct` or similar?

logstore/logstore.go
sanderpick: Just based on the comments you have in code explaining "direct", maybe "managed" is a better word? Might not take as much explaining... just a thought.

net/api/pb/threadsnet.proto
sanderpick: Is this needed here? The value of direct should be different for different peers... maybe I'll see more about it below.

net/net.go
sanderpick: If one of the _direct_ logs makes it to the replicator, _technically it does have a new address_, so maybe that shouldn't be rolled back, but still return the error from `pushLog` (as it is now). In other words, I'd vote for rolling back just the failed pushes of _direct_ logs. Sound right?

net/net.go
```diff
@@ -588,9 +593,10 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
 			if pid.String() == n.host.ID().String() {
 				return
 			}
-
-			if err = n.server.pushLog(ctx, info.ID, ownlg, pid, nil, nil); err != nil {
-				log.Errorf("error pushing log %s to %s", ownlg.ID, p)
+			for _, lg := range logs {
```
sanderpick, line 596: 👍

net/net.go
```diff
@@ -720,14 +727,17 @@ func NewRecord(r core.Record, id thread.ID, lid peer.ID) (core.ThreadRecord, err
 	return &Record{Record: r, threadID: id, logID: lid}, nil
 }
 
+// Value returns the underlying value for the record.
```
sanderpick, line 730: I generally leave off docstrings that are from implemented methods, since it's redundant... if you make a change then you have to make it in multiple places.

net/net.go
sanderpick: `getOwnLog` used to return an empty struct if own log wasn't found, so a check for pubkey check if it's _really_ a log or not.

sanderpick: Does `getOwnLog` always return an inflated struct now?

net/net.go
```diff
@@ -1022,37 +1032,22 @@ func (n *net) startPulling() {
...
-	info, err = n.getOwnLog(id)
+	lgs, err := n.store.GetOwnLogs(id)
 	if err != nil {
 		return info, err
 	}
-	if info.PubKey != nil {
-		return
+	for _, info = range lgs {
+		if info.PrivKey != nil {
+			return info, nil
```
sanderpick, line 1045: Seems like we need to ensure somewhere (maybe this function _is_ that assurance?) that a thread does not have multiple private keys (I'm not if we're missing out on some use case_. Above I mentioned something about having multiple private keys, but maybe not relevant considering this...

net/net.go
carsonfarmer: Yes, and it errors now with NotFound if it doesn't find any logs. So this additional check is now redundant. Will update.

net/net.go
carsonfarmer: Yeh, in theory, it shouldn't be possible to get into a state where there are multiple private keys, because as you've identified, this function essentially checks that. I'm going to double check, but I think this will _always_ be called any time a 'managed' log is being created, so this check should be sufficient.

core/logstore/logstore.go
carsonfarmer: Yup, I like that, easy update.

net/net.go
sanderpick: Cool, sounds good! I don't remember, but you might be able to call `AddThread` multiple times for the same thread... and both times (or just the second time) supply your _own_ private log key.

sanderpick: Looked... of course, you can't add the same thread twice. So, as it stands now, a thread can only have _one direct/managed log_. That makes sense actually, since in the multitenant setup (single writer), multiple identities will just use the same log.

carsonfarmer: Ah yes of course. But what about the situation where we have multiple remote clients treating a single remote peer as essentially a relay node? In that case, if one client is already using the remote to interact with a thread, the second client can't also use it. The 'single writer' setup isn't going to cover that use case, is it?

carsonfarmer: (That msg is more about thinking through how we might support this use-case, which is actually what motivated this PR/Issue in the first place)

carsonfarmer: Note that the answer might also be, we're not going to support that use-case directly, at this time...?

sanderpick: Hmm, yeah good point. I guess we should just allow it via `CreateThread` / `AddThread`. I think you'd have to ensure info isn't overwritten in `logstore.AddThread` when it's called multiple times for the same thread (probably okay, but just occurred to me as something to look at). Also, `ensureUnique` might have to get smarter, taking the logkey into account.

carsonfarmer: Phrew, ok, those points are the same conclusions I came to as well! That's fine then. I think totally doable in this PR update tomorrow.

core/thread/id.go
carsonfarmer: Opting for Managed as suggested elsewhere. Much better for explanation.

core/logstore/logstore.go
```diff
@@ -49,6 +52,9 @@ type Logstore interface {
 	// GetLog returns info about a log.
 	GetLog(thread.ID, peer.ID) (thread.LogInfo, error)
 
+	// GetManagedLogs returns info about locally managed logs.
+	GetManagedLogs(thread.ID) ([]thread.LogInfo, error)
```
carsonfarmer, line 56: Changed based on discussion with @sanderpick

core/thread/id.go
carsonfarmer: This is actually now ok, as we have the distinction between "owned" (have the private key) and "managed" (have only the public key)

core/thread/id.go
```diff
@@ -364,9 +366,16 @@ func (i Info) GetOwnLog() *LogInfo {
...
+	PubKey crypto.PubKey
+	// PrivKey is the log's private key.
 	PrivKey crypto.PrivKey
-	Addrs   []ma.Multiaddr
-	Head    cid.Cid
+	// Addrs are the addresses associated with the given log.
+	Addrs []ma.Multiaddr
+	// Head is the log's current head.
+	Head cid.Cid
+	// Managed logs are any logs directly added/created by the host, and/or logs for which we have the private key
```
carsonfarmer, line 379: Went with Managed, and this should be a pretty clear explanation of what that means.

logstore/logstore.go
```diff
@@ -186,15 +188,18 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
 	ls.Lock()
 	defer ls.Unlock()
 
-	err := ls.AddPubKey(id, lg.ID, lg.PubKey)
-	if err != nil {
-		return err
-	}
 	if lg.PrivKey != nil {
-		if err = ls.AddPrivKey(id, lg.ID, lg.PrivKey); err != nil {
+		if pk, _ := ls.PrivKey(id, lg.ID); pk != nil {
```
carsonfarmer, line 192: There can only be one "owned" log, with the private key. Fail here if we're trying to add a new PrivateKey. The alternative would be to not fail, and overwrite, but that seems dangerous.

logstore/logstore.go
```diff
@@ -186,15 +188,18 @@ func (ls *logstore) AddLog(id thread.ID, lg thread.LogInfo) error {
...
+			return core.ErrLogExists
+		}
+		if err := ls.AddPrivKey(id, lg.ID, lg.PrivKey); err != nil {
 			return err
 		}
 	}
+	err := ls.AddPubKey(id, lg.ID, lg.PubKey)
```
carsonfarmer, line 199: Moved below Private Key check

logstore/logstore.go
```diff
@@ -244,6 +262,26 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
 	return
 }
 
+// GetManagedLogs returns the logs the host is 'managing' under the given thread.
+func (ls *logstore) GetManagedLogs(id thread.ID) ([]thread.LogInfo, error) {
```
carsonfarmer, line 266: Renamed

logstore/logstore.go
```diff
@@ -244,6 +262,26 @@ func (ls *logstore) getLog(id thread.ID, lid peer.ID) (info thread.LogInfo, err
...
+	if err != nil {
+		return nil, err
+	}
+	var managed []thread.LogInfo
+	for _, lid := range logs {
+		lg, err := ls.GetLog(id, lid)
+		if err != nil {
+			return nil, err
+		}
+		if lg.Managed || lg.PrivKey != nil {
```
carsonfarmer, line 277: This check is essentially redundant, but keeps us backward compatible

net/net.go
```diff
@@ -218,6 +218,7 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
 	if !info.Key.Defined() {
 		info.Key = thread.NewRandomKey()
 	}
+	// This may overwrite ThreadKey information if there is a mismatch
```
carsonfarmer, line 221: Note that some existing tests relied on the ability to overwrite read and service keys, so no check or error is thrown when this happens. This _seems_ ok, because it would support things like key rotation etc... but I've added this comment as a reminder and perhaps this should be documented in the public APIs?

net/net.go
```diff
@@ -235,13 +236,18 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
...
+	// Check if thread exists.
+	thrd, err := n.store.GetThread(id)
 	if !errors.Is(err, lstore.ErrThreadNotFound) {
 		return err
 	}
+	if err == lstore.ErrThreadNotFound {
+		return nil
+	}
+	// Ensure we don't already have our 'own' log.
+	if thrd.GetOwnLog().PrivKey != nil {
```
carsonfarmer, line 248: GetOwnLog will return a nil ThreadInfo, so we still want to check for PrivKey

net/net.go
```diff
@@ -265,6 +271,7 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
 		return
 	}
 
+	// This may overwrite ThreadKey information if there is a mismatch
```
carsonfarmer, line 274: See above

net/net.go
```diff
@@ -534,12 +541,14 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
 	if err != nil {
 		return
 	}
-	ownlg, err := n.getOwnLog(info.ID)
+	logs, err := n.store.GetManagedLogs(info.ID)
```
sanderpick, line 544: Nit: `managedLogs`?

net/net.go
```diff
@@ -706,13 +722,13 @@ func (n *net) getRecord(ctx context.Context, id thread.ID, rid cid.Cid) (core.Re
...
 }
 
+// Record implements core.Record. The most basic component of a Log.
 type Record struct {
 	core.Record
 	threadID thread.ID
 	logID    peer.ID
 }
 
-// NewRecord returns a record with the given values.
```
sanderpick, line 715: Nit: not redundant, right?

net/net.go
```diff
@@ -1022,37 +1038,23 @@ func (n *net) startPulling() {
...
-		}
-	}
-	return info, nil
-}
-
-// getOrCreateOwnLoad returns the log owned by the host under the given thread.
-// If no log exists, a new one is created under the given thread.
+// getOrCreateOwnLoad returns the first log 'owned' by the host under the given thread.
+// If no log exists, a new one is created under the given thread and returned.
+// This is a strict 'owership' check vs returning managed logs.
```
sanderpick, line 1043: Nit: type `owership`

net/net.go
```diff
@@ -1022,37 +1038,23 @@ func (n *net) startPulling() {
...
 func (n *net) getOrCreateOwnLog(id thread.ID) (info thread.LogInfo, err error) {
-	info, err = n.getOwnLog(id)
+	lgs, err := n.store.GetManagedLogs(id)
```
sanderpick, line 1045: I thought we still have a direct `GetOwnLog` for this reason?

net/net.go
sanderpick: Could we also check that the logID doesn't already exist? (That would apply to pubkey-only logs)

net/net.go
sanderpick: What about only adding new keys if they don't already exists? So, modifying `logstore.AddThread` to not blindly overwrite?

carsonfarmer: So silently ignore if new/mismatched keys are used? That's an option for sure, would it lead to unexpected results?

sanderpick: Maybe just error if something would be overwritten?

carsonfarmer: Yeh that's what I had, but we have some existing tests that rely on being able to overwrite. Hence my comment above... unless I'm missing something here?

carsonfarmer: In particular, this test: https://github.com/textileio/go-threads/blob/master/db/db_test.go#L94

carsonfarmer: We could modify `NewDB` to allow passing in ThreadOptions, so that you could specify the same keys again, and if you didn't, return an error.

net/net.go
carsonfarmer: yup :p

carsonfarmer: was just cumbersome to use in this case, but should consolidate usage.

net/net.go
carsonfarmer: That's what I went with.

carsonfarmer:
Ok, @sanderpick, this ended up touching a lot more places than I had initially anticipated, but it should now cover all our bases. Checks for all keys types have been added, and errors thrown for any mismatches. One 'own' log per thread is enforced, but multiple 'managed' logs are possible. There are also tests for the expected behavior, and peers can even replicate managed logs which is important for our JS use-cases.

db/options.go
```diff
@@ -68,6 +69,13 @@ func WithNewToken(t thread.Token) NewOption {
 	}
 }
 
+// WithNewThreadKey provides control over thread keys to use with a db.
+func WithNewThreadKey(key thread.Key) NewOption {
```
carsonfarmer, line 73: This was required to be able to get some tests passing again, but is also useful/required.

logstore/logstore.go
```diff
@@ -89,13 +92,35 @@ func (ls *logstore) AddThread(info thread.Info) error {
 	if info.Key.Service() == nil {
 		return fmt.Errorf("a service-key is required to add a thread")
 	}
-	if err := ls.AddServiceKey(info.ID, info.Key.Service()); err != nil {
+	sk, err := ls.ServiceKey(info.ID)
```
carsonfarmer, line 95: Keeping these checks in the high-level AddThread function rather than in the individual stores because it might still be useful to override this behavior in certain cases. So seems like the most common operation should fail, but if you're digging into internals, its your fault if something get's messed up.

net/net.go
```diff
@@ -234,12 +234,37 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
 	return n.getThreadWithAddrs(id)
 }
 
-func (n *net) ensureUnique(id thread.ID) error {
-	_, err := n.store.GetThread(id)
-	if err == nil {
-		return lstore.ErrThreadExists
+func (n *net) ensureUnique(id thread.ID, key crypto.Key) error {
```
carsonfarmer, line 237: This is now a lot more verbose, as there are multiple different possible scenarios here, but I think the early outs help.

net/net.go
```diff
@@ -234,12 +234,37 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
...
+	// Check if thread exists.
+	thrd, err := n.store.GetThread(id)
+	if errors.Is(err, lstore.ErrThreadNotFound) {
+		return nil
 	}
-	if !errors.Is(err, lstore.ErrThreadNotFound) {
+	if err != nil {
+		return err
+	}
+	// Early out if no log key required.
```
carsonfarmer, line 246: If we don't care about log keys, then we'll end up creating a random one downstream, with negligible chance of a collision.

net/net.go
```diff
@@ -234,12 +234,37 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
...
+	if key == nil {
+		return nil
+	}
+	// Ensure we don't already have this log in our store.
+	if _, ok := key.(crypto.PrivKey); ok {
```
carsonfarmer, line 251: If this is a private key, and we don't already have a private key for this thread, we're good to go, otherwise, we fail 'early' here.

net/net.go
```diff
@@ -234,12 +234,37 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
...
+		// Ensure we don't already have our 'own' log.
+		if thrd.GetOwnLog().PrivKey != nil {
+			return lstore.ErrThreadExists
+		}
+		return nil
+	}
+	pubKey, ok := key.(crypto.PubKey)
```
carsonfarmer, line 258: If this is a public key, just make sure we don't already have a log for this key.

net/net_test.go
```diff
@@ -228,6 +304,59 @@ func TestNet_AddReplicator(t *testing.T) {
...
+	info, err := n1.CreateThread(ctx, tid, core.WithLogKey(pk))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	addr, err := ma.NewMultiaddr("/p2p/" + n2.Host().ID().String())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, err = n1.AddReplicator(ctx, info.ID, addr); err != nil {
```
sanderpick, line 333: Sweet

carsonfarmer:
Thank _you_ @sanderpick!

PR #378, opened 2020-06-22
Let peers add threads from self + replicate

carsonfarmer:
Avoids extraneous dials when the data can assume to be already available locally (self dials). This is useful/required for multi-tenant setups where a single peer may be managing multiple threads/logs on behalf of other (remote) peers. For instance, we don't want clients to 'care' if their remote host is already be replicating a thread they are involved in.

Ideally, there would be some logic in here to catch things like this (or other combinations of local vs remote addrs):

```
import manet "github.com/multiformats/go-multiaddr-net"
...
// Early out if we aren't going to be able to dial this peer
if !manet.IsPublicAddr(addr) {
	err = fmt.Errorf("addr must be publicly routable address")
	return
}
```

But I don't think we can assume, based on address alone, that we can filter something out. We might be,
* A local peer dialing another peer on the same machine
* A local peer dialing another peer on another machine on the same local network
* A local peer dialing another peer on another machine on the internet
* A client peer controlling a remote peer to do any of the above

So I'm going to leave most of the 'filtering' logic to clients. For now, the Go client will _not_ do any filtering, because likely it is operating via a localhost, but the Javascript clients should filter out local host addrs by default.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -293,6 +293,27 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
 		return
 	}
 
+	threadComp, err := ma.NewComponent(thread.Name, id.String())
+	if err != nil {
+		return
+	}
+	peerAddr := addr.Decapsulate(threadComp)
+	addri, err := peer.AddrInfoFromP2pAddr(peerAddr)
+	if err != nil {
+		return
+	}
+
+	// Check if we're dialing ourselves (regardless of addr)
+	addFromSelf := addri.ID.Pretty() == n.host.ID().Pretty()
+	if addFromSelf {
+		// Error if we don't have the thread locally
+		if _, err = n.store.GetThread(id); errors.Is(err, lstore.ErrThreadNotFound) {
+			err = fmt.Errorf("cannot retrive thread from self: %s", err)
+			return
+		}
+	}
+
+	// Even if we already have the thread locally, we might still need to add a new log
 	if err = n.store.AddThread(thread.Info{
 		ID:  id,
 		Key: args.ThreadKey,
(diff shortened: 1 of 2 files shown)


net/net.go
```diff
@@ -293,6 +293,27 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
 		return
 	}
 
+	threadComp, err := ma.NewComponent(thread.Name, id.String())
```
carsonfarmer, line 296: This is just moved up from below.

net/net.go
```diff
@@ -293,6 +293,27 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
...
+		return
+	}
+	peerAddr := addr.Decapsulate(threadComp)
+	addri, err := peer.AddrInfoFromP2pAddr(peerAddr)
+	if err != nil {
+		return
+	}
+
+	// Check if we're trying to dial ourselves (regardless of addr)
+	addFromSelf := addri.ID.Pretty() == n.host.ID().Pretty()
```
carsonfarmer, line 307: Here we're checking the ID, rather than the address. Since we can't dial ourself anyway, this seems like a broader check. Regardless of address, if we're trying to dial ourself, don't do it.

net/net.go
```diff
@@ -293,6 +293,27 @@ func (n *net) AddThread(ctx context.Context, addr ma.Multiaddr, opts ...core.New
...
+	if addFromSelf {
+		// Error if we don't have the thread locally
+		if _, err = n.store.GetThread(id); errors.Is(err, lstore.ErrThreadNotFound) {
```
carsonfarmer, line 310: If we are dialing ourselves, but we don't have the thread, that's no good. Throw error.

net/net.go
```diff
@@ -576,27 +592,31 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
...
 
-	// Send all logs to the new replicator
-	for _, l := range info.Logs {
-		if err = n.server.pushLog(ctx, info.ID, l, pid, info.Key.Service(), nil); err != nil {
-			for _, lg := range managedLogs {
-				// Rollback this log only and then bail
-				if lg.ID.Pretty() == l.ID.Pretty() {
-					if err := n.store.SetAddrs(info.ID, lg.ID, lg.Addrs, pstore.PermanentAddrTTL); err != nil {
-						log.Errorf("error rolling back log address change: %s", err)
+		// Send all logs to the new replicator
```
carsonfarmer, line 606: Only push to the replicator if that replicator isn't _us_.

carsonfarmer: It doesn't _really_ make sense to add a replicator that is "us"... but in a multi-tenant situation, the peer calling `AddReplicator` shouldn't know or care if a given peer (that they'd like to add as replicator) is on the same machine.

net/net.go
jsign: Nit: I think you can avoid the `.Pretty()` calls, since `ID` is already a `string` (binary).

net/net.go
```diff
@@ -576,27 +592,31 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
...
 
-	// Update peerstore address
-	dialable, err := getDialable(paddr)
-	if err == nil {
-		n.host.Peerstore().AddAddr(pid, dialable, pstore.PermanentAddrTTL)
-	} else {
-		log.Warnf("peer %s address requires a DHT lookup", pid)
-	}
+	// Check if we're dialing ourselves (regardless of addr)
+	if pid.Pretty() != n.host.ID().Pretty() {
```
jsign, line 596: Nit: Similarly about `.Pretty()`.

net/net.go
```diff
@@ -576,27 +592,31 @@ func (n *net) AddReplicator(ctx context.Context, id thread.ID, paddr ma.Multiadd
...
-				// Rollback this log only and then bail
-				if lg.ID.Pretty() == l.ID.Pretty() {
-					if err := n.store.SetAddrs(info.ID, lg.ID, lg.Addrs, pstore.PermanentAddrTTL); err != nil {
-						log.Errorf("error rolling back log address change: %s", err)
+		// Send all logs to the new replicator
+		for _, l := range info.Logs {
+			if err = n.server.pushLog(ctx, info.ID, l, pid, info.Key.Service(), nil); err != nil {
+				for _, lg := range managedLogs {
+					// Rollback this log only and then bail
+					if lg.ID.Pretty() == l.ID.Pretty() {
```
jsign, line 611: Nit: .Pretty.

carsonfarmer:
Cheers!

PR #387, opened 2020-06-26
Don't panic on empty schemas when looking for _id

### LINKED ISSUE #369: getSchemaTypeProperties panics on empty schema
This was discovered during a recent test where an input schema of `{}` (which is a valid json-schema.org schema) was used in updateCollection.

carsonfarmer:
Also adds tests for this edge case to prevent regressions.

Fixes #369 

Note that this still will cause an error if the schema _doesn't_ have an `_id` field. This makes things safer on the Go side. Calling clients should ensure on the client side that this is the case _before_ calling out to the remote Go peer.

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -245,7 +245,7 @@ func (t *Txn) Create(new ...[]byte) ([]core.InstanceID, error) {
 	return results, nil
 }
 
-// Save saves an instance changes to be commited when the current transaction commits.
+// Save saves an instance changes to be committed when the current transaction commits.
 func (t *Txn) Save(updated ...[]byte) error {
 	for i := range updated {
 		if t.readonly {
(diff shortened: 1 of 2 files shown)


db/collection.go
```diff
@@ -401,7 +404,12 @@ func getSchemaTypeAtPath(schema *jsonschema.Schema, pth string) (*jsonschema.Typ
 	return jt, nil
 }
 
+// getSchemaTypeProperties extracts a map of schema properties from a given input schema.
+// If there are no available properties, it will return an empty (nil) map
```
jsign, line 408: What's your opinion about returning an empty map (`make(map[string]*jsonschema.Type`)?
Returning `error = nil`, and a `nil` map may be buying a future panic on someone trying to use the map.

carsonfarmer: Ah yeh, that's not a bad idea, so calls like map["something"] will always return empty, but at least not panic. Good idea. Will push that change and then merge.

jsign: Just in case, in Go you can do:
```go
var foo map[string]string // So, nil map
fmt.Println(foo["hello"))
```
and that won't panic.

The problem with nil in maps is writing (e.g: `foo["bad"]="idea"`), you'll get a panic.
So the comment was more about just in case covering that scenario, which most prob the caller shouldn't do (but just in case).

PR #388, opened 2020-06-29
crypto/asymmetric: use alternative curve transformation logic

### LINKED ISSUE #383: cannot find package /extra25519
image
 this is the issue : https://github.com/agl/ed25519/issues/27
mb the solution : https://godoc.org/golang.org/x/crypto/ed25519 this dont work.
you can swap agl with decred : extra "github.com/decred/ed25519/extra25519"

image
after the fix

jsign:
Closes #383

### DIFF
--- a/crypto/asymmetric/asymmetric_old/asymmetric_old_test.go
+++ b/crypto/asymmetric/asymmetric_old/asymmetric_old_test.go
@@ -1,17 +1,119 @@
-package asymmetric
+package asymmetric_old
 
 import (
-	"crypto/ed25519"
 	"crypto/rand"
-	"crypto/sha512"
+	"encoding/hex"
 	"fmt"
-	"math/big"
+	"testing"
+
+	extra "github.com/agl/ed25519/extra25519"
+	"github.com/textileio/go-threads/crypto/asymmetric"
 
 	"github.com/libp2p/go-libp2p-core/crypto"
-	"golang.org/x/crypto/curve25519"
 	"golang.org/x/crypto/nacl/box"
 )
 
+func TestCompatEncryptDecrypt(t *testing.T) {
+	plaintext := "Hello, this is a compatibility test!!"
+	priv, pub, err := crypto.GenerateKeyPair(crypto.Ed25519, 0)
+	if err != nil {
+		t.Error(err)
+	}
+
+	t.Run("Encrypt New -> Decrypt Old", func(t *testing.T) {
+		ek, err := asymmetric.FromPubKey(pub)
+		if err != nil {
+			t.Error(err)
+		}
+		ciphertext, err := ek.Encrypt([]byte(plaintext))
+		if err != nil {
+			t.Error(err)
+			return
+		}
+
+		dk, err := FromPrivKey(priv)
+		if err != nil {
+			t.Error(err)
+		}
+
+		decryptedPlaintext, err := dk.Decrypt(ciphertext)
+		if err != nil {
+			t.Error(err)
+			return
+		}
+		if string(decryptedPlaintext) != plaintext {
+			t.Error("result plaintext doesn't match original plaintext")
+		}
+	})
+	t.Run("Encrypt Old -> Decrypt New", func(t *testing.T) {
+		ek, err := FromPubKey(pub)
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/AndreasBriese/bbloom v0.0.0-20190823232136-616930265c33 // indirect
+	// agl/ed25519 only used in tests for backward compatibility, *do not* use in production code.
```
jsign, line 7: I added this note to be clear that this dependency is only used in tests. 
Eventually, I guess we could stop caring about older implementations and directly remove the compatibility test thus this dependency. I think won't be a problem leaving this anyway since it would be weird that something use this dependency in production since is too crypto-focused.

crypto/asymmetric/asymmetric_old/asymmetric_old_test.go
```diff
@@ -0,0 +1,284 @@
+package asymmetric_old
```
jsign, line 1: The `asymmetric_test.go` still lives unchanged, so the new implementation continues to verify the test we had.

In this new `asymetric_old_test.go`, I added a complete copy-paste of *all* the previous implementation. (not just the changed part, so we can be sure about complete compatibility).

This tests do some bi-directional tests between encryption/decryption using new and old, old and new implementations. The idea is to be sure that old and new world still works correctly toegheter.

crypto/asymmetric/asymmetric.go
```diff
@@ -174,6 +177,76 @@ func privateToCurve25519(k *crypto.Ed25519PrivateKey) (*[EphemeralPublicKeyBytes
...
-	extra.PrivateKeyToCurve25519(&cs, &sk)
+	buf := ed25519PrivateKeyToCurve25519(ed25519.PrivateKey(sk[:]))
+	copy(cs[:], buf)
 	return &cs, nil
 }
+
+// The below code is copied exactly as found in:
+// https://github.com/FiloSottile/age/blob/c9a35c072716b5ac6cd815366999c9e189b0c317/internal/agessh/agessh.go#L179
+// Note: We should switch to x/crypto/ed25519 curve transformations when
+// the following issue closes: https://github.com/golang/go/issues/20504.
```
jsign, line 188: ^

crypto/asymmetric/asymmetric.go
```diff
@@ -97,10 +100,10 @@ func publicToCurve25519(k *crypto.Ed25519PublicKey) (*[EphemeralPublicKeyBytes]b
 		return nil, err
 	}
 	copy(pk[:], r)
-	success := extra.PublicKeyToCurve25519(&cp, &pk)
-	if !success {
-		return nil, fmt.Errorf("error converting ed25519 pubkey to curve25519 pubkey")
-	}
+
+	buf := ed25519PublicKeyToCurve25519(ed25519.PublicKey(pk[:]))
```
carsonfarmer, line 104: Excellent

crypto/asymmetric/asymmetric.go
carsonfarmer: Bummer, but good call!

crypto/asymmetric/asymmetric_old/asymmetric_old_test.go
carsonfarmer: Nice!

crypto/asymmetric/asymmetric_test.go
```diff
@@ -84,6 +83,6 @@ func TestDecrypt(t *testing.T) {
 	}
 	_, err = dk.Decrypt(cipherTextInvalidBytes)
 	if err != BoxDecryptionError {
-		t.Error("failed to catch curve25519 drcyption error")
+		t.Error("failed to catch curve25519 decryption error")
```
carsonfarmer, line 86: :)

go.mod
carsonfarmer: Will this ever get overwritten?

jsign: I force-updated some dependency, and seems `go mod tidy` doesn't overwrite this, so will stay.

PR #392, opened 2020-07-08
Enable keepalive pinging thanks to @asutula

carsonfarmer:
Signed-off-by: Carson Farmer <carson.farmer@gmail.com>

This is a direct copy of the solution added here: https://github.com/textileio/textile/pull/263

### DIFF
--- a/threadsd/main.go
+++ b/threadsd/main.go
@@ -118,6 +118,7 @@ func main() {
 			return true
 		}),
 		grpcweb.WithWebsockets(true),
+		grpcweb.WithWebsocketPingInterval(time.Second*5),
 		grpcweb.WithWebsocketOriginFunc(func(req *http.Request) bool {
 			return true
 		}))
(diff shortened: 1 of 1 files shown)


threadsd/main.go
```diff
@@ -118,6 +118,7 @@ func main() {
 			return true
 		}),
 		grpcweb.WithWebsockets(true),
+		grpcweb.WithWebsocketPingInterval(time.Second*5),
```
sanderpick, line 121: Can you make this a flag? Should be configurable by the user

carsonfarmer: Sure

threadsd/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 	connLowWater := fs.Int("connLowWater", 100, "Low watermark of libp2p connections that'll be maintained")
 	connHighWater := fs.Int("connHighWater", 400, "High watermark of libp2p connections that'll be maintained")
 	connGracePeriod := fs.Duration("connGracePeriod", time.Second*20, "Duration a newly opened connection is given before it becomes subject to pruning")
+	keepAliveInterval := fs.Duration("keepAliveInterval", time.Second*5, "Enables websocket keepalive pinging with the specified timeout interval. Configured interval must be >= 1s.")
```
carsonfarmer, line 38: Leaving this defaulting to 5 seconds, as the Listen API is much more useful when the socket is kept alive (in the case of slow updates in realistic scenarios).

threadsd/main.go
```diff
@@ -118,6 +119,7 @@ func main() {
 			return true
 		}),
 		grpcweb.WithWebsockets(true),
+		grpcweb.WithWebsocketPingInterval(*keepAliveInterval),
```
carsonfarmer, line 122: This is the same thing that the improbable-eng lib does. They pass along keepAliveInterval regardless of value, and if it is less than 1 second, it is ignored.

PR #404, opened 2020-07-15
Protect db threads

### LINKED ISSUE #400: Protect threads that are backing a db from direct deletion
Otherwise, on re-load, this happens:
```
textile_1  | 2020-07-14T01:23:39.912Z	FATAL	db	db/db.go:173	unable to connect app: error getting thread bafk5omesrcb4nmnclmrqzhufzmz6manujcpuvpnei7jlstfnhcd7djq: thread not found
```

sanderpick:
Closes #400

- Adds a thread maps `apps` to `net`, which is used to determine if a thread is connected to an app and protect it from deletion 
- Skips the fatal log when there's an error connecting an app to a thread and deletes the failing db (needed because production has some orphaned dbs)
- Forces a db to have a name, since without a name, there won't be any keys in the datastore to reload the db with after a manager restart... but the underlying thread will still be there :/


.github/workflows/test.yml
```diff
@@ -57,33 +57,3 @@ jobs:
           go get -v -t -d ./...
       - name: Test
         run: go test ./integrationtests/foldersync/
-
-  android-client:
```
sanderpick, line 61: Dropping these tests... we don't have official support anymore

db/manager_test.go
```diff
@@ -172,12 +172,35 @@ func TestManager_GetDB(t *testing.T) {
...
+		checkErr(t, err)
+
 		time.Sleep(time.Second)
 
 		err = man.Close()
 		checkErr(t, err)
 		err = n.Close()
 		checkErr(t, err)
+
+		t.Run("test get deleted db after restart", func(t *testing.T) {
```
sanderpick, line 186: This is just double double checking that a db is actually deleted after deleting it.

db/manager.go
```diff
@@ -83,16 +84,28 @@ func NewManager(network app.Net, opts ...NewOption) (*Manager, error) {
...
+			continue
+		}
 		opts, err := getDBOptions(id, m.opts, "")
 		if err != nil {
 			return nil, err
 		}
 		s, err := newDB(m.network, id, opts)
 		if err != nil {
-			return nil, err
+			log.Errorf("unable to reload db %s: %s (marked for deletion)", id, err)
```
carsonfarmer, line 96: 👍

core/app/app.go
```diff
@@ -169,6 +172,11 @@ func (c *Connector) ThreadID() thread.ID {
 	return c.threadID
 }
 
+// Token returns the net token.
+func (c *Connector) Token() net.Token {
```
sanderpick, line 176: Adding the concept of a token to a connector

core/app/app.go
```diff
@@ -212,7 +220,7 @@ func (c *Connector) appToThread(wg *sync.WaitGroup) {
 				return
 			}
 			ctx, cancel := context.WithTimeout(context.Background(), addRecordTimeout)
-			if _, err := c.Net.CreateRecord(ctx, c.threadID, event.Node, net.WithThreadToken(event.Token)); err != nil {
+			if _, err := c.Net.CreateRecord(ctx, c.threadID, event.Node, net.WithThreadToken(event.Token), net.WithAPIToken(c.token)); err != nil {
```
sanderpick, line 223: Apps (dbs) can supply their token to the net api to authenticate

core/net/net.go
```diff
@@ -69,3 +70,12 @@ type API interface {
 	// Subscribe returns a read-only channel of records.
 	Subscribe(ctx context.Context, opts ...SubOption) (<-chan ThreadRecord, error)
 }
+
+// Token is used to restrict network APIs to a single app.App.
+// In other words, a net token protects against writes and deletions
+// external to an app.
+type Token []byte
```
sanderpick, line 77: See docstring

PR #416, opened 2020-08-06
Validator Functions

### LINKED ISSUE #406: Enable errors to flow back to user when creating records from an app
The current design will just panic if there's an error when an app creates new thread records.

### LINKED ISSUE #418: Collection-level validator functions
I've already been POC-ing this idea, but creating an issue now for tracking. Essentially, allow the user to write a function that validates a update event.

sanderpick:
Closes #418 
Closes #406

### Net
- Allows a single thread host to maintain multiple logs for a given thread authored by different identities. 
- Provides a mechanism to allow an `App` to validate a `Record` before it's committed to a log.

### DB
- Collections can now take a JS function that validates collection writes. This opens the door to a lot of complex use cases, e.g., Buckets uses this feature to implement file-level access control.
- Collections can now take a JS function that filters collection reads. This opens the door to a lot of complex use cases, e.g., Buckets uses this feature to implement file-level access control.

### Misc.
- Updates `go-libp2p` to `v0.10.3`
- Removes the replace directives in `go.mod`, which are cumbersome for downstream projects.


core/app/app.go
```diff
@@ -177,53 +147,23 @@ func (c *Connector) Token() net.Token {
...
-				log.Debug("notification channel closed, not listening to external changes anymore")
-				return
-			}
-			if err = c.app.HandleNetRecord(rec, c.threadKey, c.logID, fetchEventTimeout); err != nil {
-				log.Fatal(err)
-			}
-		}
-	}
+// CreateNetRecord calls net.CreateRecord while supplying thread ID and API token.
+func (c *Connector) CreateNetRecord(ctx context.Context, body format.Node, token thread.Token) (net.ThreadRecord, error) {
```
sanderpick, line 151: These connector methods are mostly just helpers that add connector params like thread ID and API token to the underlying methods.

core/db/db.go
```diff
@@ -36,6 +36,8 @@ type Event interface {
 	InstanceID() InstanceID
 	// Collection is the associated instance's collection name.
 	Collection() string
+	// Marshal the event to JSON.
+	Marshal() ([]byte, error)
```
sanderpick, line 40: This is used to marshal the event to something that can be leveraged in a JS validator function.

core/logstore/logstore.go
```diff
@@ -84,6 +84,9 @@ type ThreadMetadata interface {
 
 	// PutBytes stores a byte value under key.
 	PutBytes(t thread.ID, key string, val []byte) error
+
+	// ClearMetadata clears all metadata under a thread.
+	ClearMetadata(t thread.ID) error
```
sanderpick, line 89: This was a missing method but unnoticed because until now we weren't using the metadata API.

logstore/logstore.go
```diff
@@ -191,6 +191,10 @@ func (ls *logstore) DeleteThread(id thread.ID) error {
 		return err
 	}
 
+	if err := ls.ClearMetadata(id); err != nil {
```
sanderpick, line 194: Clean up the metadata on thread delete.

jsonpatcher/jsonpatcher.go
```diff
@@ -249,4 +261,34 @@ func (je patchEvent) Collection() string {
...
+	Patch          operationJson `json:"patch"`
+}
+
+type operationJson struct {
+	Type       string      `json:"type"`
+	InstanceID string      `json:"instance_id"`
+	JSONPatch  interface{} `json:"json_patch"`
+}
+
+func (je patchEvent) Marshal() ([]byte, error) {
```
sanderpick, line 277: This uses some custom types that have `json` decorators

jsonpatcher/jsonpatcher.go
```diff
@@ -22,9 +23,21 @@ type operationType int
 const (
 	create operationType = iota
 	save
-	delete
+	del
```
sanderpick, line 26: Lint fix

jsonpatcher/jsonpatcher.go
```diff
@@ -47,7 +60,6 @@ func init() {
 	cbornode.RegisterCborType(patchEvent{})
 	cbornode.RegisterCborType(recordEvents{})
 	cbornode.RegisterCborType(operation{})
-	cbornode.RegisterCborType(time.Time{})
```
sanderpick, line 50: Turns out `time.Time` was not properly being cbor encoded, so I switched the handling to unix nanos.

go.mod
```diff
@@ -15,46 +11,49 @@ require (
...
 	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
 	github.com/ipfs/go-ipfs-files v0.0.4 // indirect
 	github.com/ipfs/go-ipld-cbor v0.0.4
 	github.com/ipfs/go-ipld-format v0.2.0
 	github.com/ipfs/go-log v1.0.4
 	github.com/ipfs/go-merkledag v0.3.2
-	github.com/libp2p/go-libp2p v0.8.2
-	github.com/libp2p/go-libp2p-connmgr v0.2.1
-	github.com/libp2p/go-libp2p-core v0.5.2
+	github.com/libp2p/go-libp2p v0.10.3
```
sanderpick, line 41: Finally bumped `libp2p`

go.mod
```diff
@@ -2,10 +2,6 @@ module github.com/textileio/go-threads
 
 go 1.14
 
-replace github.com/ipfs/go-datastore v0.4.4 => github.com/textileio/go-datastore v0.4.5-0.20200728205504-ffeb3591b248
```
sanderpick, line 5: No longer need these... `net` still needs `ipfs/go-datastore` since it interacts with things like `Blockstore`, which depend on `ipfs/go-datastore`, but `db` is using `textileio/go-datastore`.

db/query.go
```diff
@@ -302,13 +305,23 @@ func (t *Txn) Find(q *Query) ([][]byte, error) {
...
+		return nil, err
+	}
 	var values []MarshaledResult
 	for {
 		res, ok := iter.NextSync()
 		if !ok {
 			break
 		}
-		values = append(values, res)
+		res.Value, err = t.collection.filterRead(pk, res.Value)
```
sanderpick, line 318: Each result gets passes through the `ReadFilter` if present.

db/options.go
```diff
@@ -184,6 +186,23 @@ func WithNewManagedToken(t thread.Token) NewManagedOption {
 	}
 }
 
+// WithNewManagedThreadKey provides control over thread keys to use with a managed db.
+func WithNewManagedThreadKey(key thread.Key) NewManagedOption {
```
sanderpick, line 190: Just adding these for consistency

db/listeners.go
```diff
@@ -29,10 +29,6 @@ func (d *DB) Listen(los ...ListenOption) (Listener, error) {
 	return sl, nil
 }
 
-func (d *DB) LocalEventListen() *app.LocalEventListener {
```
sanderpick, line 32: No longer needed

db/db.go
```diff
@@ -491,12 +539,29 @@ func defaultIndexFunc(d *DB) func(collection string, key ds.Key, oldData, newDat
 	}
 }
 
-func (d *DB) HandleNetRecord(rec net.ThreadRecord, key thread.Key, lid peer.ID, timeout time.Duration) error {
-	if rec.LogID() == lid {
-		return nil // Ignore our own events since DB already dispatches to DB reducers
+func (d *DB) ValidateNetRecordBody(_ context.Context, body format.Node, identity thread.PubKey) error {
```
sanderpick, line 542: `Record`s that arrive over the `net` layer are presented to the connected app for validation against the collection's `WriteValidator`

db/collection_test.go
```diff
@@ -102,6 +102,48 @@ func TestNewCollection(t *testing.T) {
...
+		checkErr(t, err)
+	})
+	t.Run("WithReadFilter", func(t *testing.T) {
+		t.Parallel()
+		db, clean := createTestDB(t)
+		defer clean()
+		c, err := db.NewCollection(CollectionConfig{
+			Name:   "Dog",
+			Schema: util.SchemaFromInstance(&Dog{}, false),
+			ReadFilter: `
```
sanderpick, line 130: Here's an example of using a read filter to transform the instance. Buckets uses this to filter the paths shown in the bucket based on the bucket ACL.

db/collection_test.go
```diff
@@ -102,6 +102,48 @@ func TestNewCollection(t *testing.T) {
...
 		}
 	})
+	t.Run("WithWriteValidator", func(t *testing.T) {
+		t.Parallel()
+		db, clean := createTestDB(t)
+		defer clean()
+		c, err := db.NewCollection(CollectionConfig{
+			Name:           "Dog",
+			Schema:         util.SchemaFromInstance(&Dog{}, false),
+			WriteValidator: `return true`,
```
sanderpick, line 112: The most basic validator is just a pass through.

db/collection.go
```diff
@@ -341,13 +431,32 @@ func (t *Txn) Delete(ids ...core.InstanceID) error {
...
+	}
 	for i := range ids {
 		key := baseKey.ChildString(t.collection.name).ChildString(ids[i].String())
 		exists, err := t.collection.db.datastore.Has(key)
 		if err != nil {
 			return false, err
 		}
-		if !exists {
+		if exists {
+			bytes, err := t.collection.db.datastore.Get(key)
```
sanderpick, line 448: If has would return true, we're forced to load the full instance and pass it through the read filter. If the result is nil, we return false (not exist).

db/collection.go
```diff
@@ -196,6 +211,101 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	if err != nil && !errors.Is(err, ds.ErrNotFound) {
+		return err
+	}
+	if val != nil {
+		inv, err = parseJSON(vm, val)
+		if err != nil {
+			return err
+		}
+	}
+	res, err := validate(nil, writer, event, inv)
```
sanderpick, line 263: The user-defined method is given the writer's identity, the event, and the current instance.

db/collection.go
```diff
@@ -196,6 +211,101 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	vm := goja.New()
+	filter, reader, err := getJSFunc(vm, c.readFilter, identity)
+	if err != nil {
+		return nil, err
+	}
+	inv, err := parseJSON(vm, instance)
+	if err != nil {
+		return nil, err
+	}
+	res, err := filter(nil, reader, inv)
```
sanderpick, line 296: The user-defined read filter method is given the writer's identity and the current instance.

db/collection.go
```diff
@@ -341,13 +431,32 @@ func (t *Txn) Delete(ids ...core.InstanceID) error {
 
 // Has returns true if all IDs exists in the collection, false otherwise.
 func (t *Txn) Has(ids ...core.InstanceID) (bool, error) {
+	if err := t.collection.db.connector.Validate(t.token, true); err != nil {
```
sanderpick, line 434: Here's an example of calling down into the `net` layer to check a `thread.Token`. The bool here means the access is read-only.

core/thread/identity.go
```diff
@@ -167,8 +187,30 @@ func NewToken(issuer crypto.PrivKey, key PubKey) (tok Token, err error) {
 	return Token(str), nil
 }
 
-// Validate returns non-nil if token was issued by issuer.
-// If token is present and valid, the embedded public key is returned.
+// PubKey returns the public key encoded in the token.
+// Note: This does NOT verify the token.
+func (t Token) PubKey() (PubKey, error) {
```
sanderpick, line 192: This is used to pluck the pub key from a `thread.Token` to be used in the collection `WriteValidator` and `ReadFilter` functions.

core/thread/id.go
```diff
@@ -353,9 +353,10 @@ type Info struct {
 	Addrs []ma.Multiaddr
 }
 
-// GetOwnLog returns the first log found with a private key.
+// GetFirstPrivKeyLog returns the first log found with a private key.
 // This is a strict owership check, vs returning all directly 'managed' logs.
-func (i Info) GetOwnLog() *LogInfo {
+// Deprecated: This is no longer safe to use.
+func (i Info) GetFirstPrivKeyLog() *LogInfo {
```
sanderpick, line 359: Renaming this to reflect more specifically what it does. This is now used only to support legacy logs that do not have an associated author (see `net/net.go` for more comments on this).

net/net.go
```diff
@@ -240,62 +235,27 @@ func (n *net) CreateThread(_ context.Context, id thread.ID, opts ...core.NewThre
...
 	if err != nil {
 		return
 	}
-	if err = id.Validate(); err != nil {
+	identity, err := n.Validate(id, args.Token, false)
+	if err != nil {
 		return
 	}
-	if err = n.ensureUnique(id, args.LogKey); err != nil {
+	if identity != nil {
```
sanderpick, line 252: If an identity is not given via a `thread.Token`, we fall back to using the host's identity.

net/net.go
```diff
@@ -674,58 +617,55 @@ func getDialable(addr ma.Multiaddr) (ma.Multiaddr, error) {
...
 	}
-	if !n.validateAPIToken(id, args.APIToken) {
+	if identity == nil {
+		identity = thread.NewLibp2pPubKey(n.getPrivKey().GetPublic())
+	}
+	con, ok := n.getConnector(id, args.APIToken)
+	if !ok {
 		return nil, fmt.Errorf("cannot create record: %w", app.ErrThreadInUse)
+	} else if con != nil {
+		if err = con.ValidateNetRecordBody(ctx, body, identity); err != nil {
```
sanderpick, line 636: If this thread is owned by an app, allow it to validate the record.

net/net.go
```diff
@@ -866,41 +800,54 @@ func (n *net) ConnectApp(a app.App, id thread.ID) (*app.Connector, error) {
...
+func (n *net) Validate(id thread.ID, token thread.Token, readOnly bool) (thread.PubKey, error) {
+	if err := id.Validate(); err != nil {
+		return nil, err
+	}
+	return token.Validate(n.getPrivKey())
+}
+
+// getConnector returns the connector tied to the thread if it exists
+// and whether or not the token is valid.
+func (n *net) getConnector(id thread.ID, token core.Token) (*app.Connector, bool) {
```
sanderpick, line 821: Connectors proof ownership of a thread by passing a unique `core.Token` (random bytes).

net/net.go
```diff
@@ -954,20 +902,42 @@ func (n *net) putRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
...
+			if key != nil {
+				dbody, err := event.GetBody(ctx, n, key)
+				if err != nil {
+					return err
+				}
+				identity := &thread.Libp2pPubKey{}
+				if err = identity.UnmarshalBinary(r.PubKey()); err != nil {
+					return err
+				}
+				if err = con.ValidateNetRecordBody(ctx, dbody, identity); err != nil {
```
sanderpick, line 920: In order to pass this record up the the app, we have to decrypt the body.

net/net.go
```diff
@@ -1112,31 +1078,78 @@ func (n *net) startPulling() {
...
-	return
+	if err = n.store.PutBytes(id, identity.String(), lidb); err != nil {
+		return info, err
+	}
+	return info, nil
+}
+
+// getOrCreateLog returns a log for identity under the given thread.
+// If no log exists, a new one is created.
+func (n *net) getOrCreateLog(id thread.ID, identity thread.PubKey) (info thread.LogInfo, err error) {
```
sanderpick, line 1122: This is a fairly significant change: A host can have multiple logs for the same thread that are being authored by different identities. In order to do this, we have to be able to look up a log by an identity. Remember, the first records in a log refers to the author (identity), which allows us to prove to ourselves that it is single-writer.

I'm tracking the relationship between an identity and it's logs in the metadata store.

net/net.go
```diff
@@ -1159,6 +1172,61 @@ func (n *net) createExternalLogIfNotExist(tid thread.ID, lid peer.ID, pubKey cry
 	return nil
 }
 
+// ensureUniqueLog returns a non-nil error if a log with key already exists,
+// or if a log for identity already exists for the given thread.
+func (n *net) ensureUniqueLog(id thread.ID, key crypto.Key, identity thread.PubKey) (err error) {
```
sanderpick, line 1177: Related to the above mentioned change, ensuring a log is unique is a bit complex. Logs can be created by specifying the the public/private key, and/or specifying the identity. This method walks through those cases.

db/collection.go
```diff
@@ -196,13 +211,108 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	out := res.Export()
+	switch out.(type) {
+	case bool:
+		if out.(bool) {
+			return nil
+		} else {
+			return app.ErrInvalidNetRecordBody
+		}
+	case nil:
+		return app.ErrInvalidNetRecordBody
```
jsign, line 276: Q: This `nil` case indicates some error executing the validator func?
If that's the case, I guess the logic is leaning to the safer option to leave out the event from the log.
Is that right?

db/collection.go
```diff
@@ -196,13 +211,108 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	default:
+		return fmt.Errorf("%w: %v", app.ErrInvalidNetRecordBody, out)
```
jsign, line 278: Maybe could be merged with the `nil` case? 
(Maybe not if `nil` case describes some other case I'm missing)

db/collection.go
```diff
@@ -196,13 +211,108 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	res, err := filter(nil, reader, inv)
+	if err != nil {
+		return nil, fmt.Errorf("error running read filter func: %v", err)
+	}
+	out := res.Export()
+	switch out.(type) {
+	case nil:
+		return nil, nil
+	default:
+		return json.Marshal(out)
```
jsign, line 305: Should have a safer type checking as done in the other function?
(i.e: a `default` case that errors, and only unmarshaling in the correct output type)

db/collection.go
```diff
@@ -341,13 +431,35 @@ func (t *Txn) Delete(ids ...core.InstanceID) error {
 
 // Has returns true if all IDs exists in the collection, false otherwise.
 func (t *Txn) Has(ids ...core.InstanceID) (bool, error) {
+	if err := t.collection.db.connector.Validate(t.token, true); err != nil {
+		return false, err
+	}
+	pk, err := t.token.PubKey()
```
jsign, line 437: I see now why the clarification in the exported comment that this doesn't check is verified. 
The previous `Validate` is solving that first.

db/collection.go
```diff
@@ -341,13 +431,35 @@ func (t *Txn) Delete(ids ...core.InstanceID) error {
...
+			bytes, err = t.collection.filterRead(pk, bytes)
+			if err != nil {
+				return false, err
+			}
+			if bytes == nil { // Access denied
+				return false, nil
+			}
+		} else {
 			return false, nil
 		}
```
jsign, line 464: Some general comment now that I see this `Has` method using the read filter.
I wonder if an extra comment in the method might help a bit to understand what this is doing now with the existence of this _optional_ filter.

The `Has` function will return _false_ even if that key exist in the ThreadDB but you're not allowed to see it. So feels like operations are more like _your view about the ThreadsDB_ rather than the DB.
I guess that's the new way of interpreting methods in ThreadsDB, so maybe that comment clarification is unnecessary. Sharing thoughts.

We aren't now in perf considerations, but maybe making read filters accept a batch of instances might alleviate a bit computational things. Also, I'm not sure if JS funcs can be cached to avoid som VM-JS compilation? (not sure if done, just a feeling) So caching+batching _could_ help a bit in the future. (e.g: in this case, grouping `Has` ids in 10, and validating groups). Anyway, out of scope now... brainstorming.

db/collection.go
```diff
@@ -453,3 +585,37 @@ func setNewInstanceID(t []byte) (core.InstanceID, []byte) {
...
 	return newID, patchedValue
 }
+
+func compileJSFunc(v []byte, args ...string) ([]byte, error) {
+	if v != nil && len(v) == 0 {
+		v = nil
+	}
+	if v == nil {
+		return nil, nil
+	}
```
jsign, line 595: This could be simplified a bit, [since `len(nilSlice)==0`](https://play.golang.org/p/TFp5pb2ta_2).

db/collection.go
```diff
@@ -453,3 +585,37 @@ func setNewInstanceID(t []byte) (core.InstanceID, []byte) {
...
+		return nil, nil, fmt.Errorf("object is not a function: %s", string(obj))
+	}
+	var idv goja.Value
+	if id != nil {
+		idv, err = vm.RunString(fmt.Sprintf("'%s'", id.String()))
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	return fn, idv, nil
```
jsign, line 620: Nice. So we want to call a func, all parameters should be evaluated and passed as "object" rather than values.

db/collection_test.go
jsign: Oh, it can also mutate the instance. Scary but powerful :P
Now feel `ReadFilter` might be a bit short in semantics since isn't only deciding to "show/not_show" but possibly changing. I don't have a good candidate name though.

db/db.go
jsign: Nice. With this write validators, we should stress more than ever the single-writer approach.

That is, is very important that write validations relies on the assumption on getting writes in a single log where records are always presented in order. If some validation depends on two logs, depending on the order of processing you might have unwanted consequences.  Sharing thoughts to validate mentality.

Maybe under more advanced validators, multi-writers might make sense... but thinking about all possible out of order arrivals to validate or not sounds hard.

jsonpatcher/jsonpatcher.go
jsign: Is this breaking change? Saying: changing the json format in the records?
Thinking about if older code can unmarshal correctly the new json tags? Sorry if confused!

sanderpick: I created these new `Json` flavored structs (`patchEventJson` and `operationJson`) to avoid that problem. `Marshal` first unmarshals the patch to an empty interface, which is used to manually create a `patchEventJson` object, which is finally marshaled to JSON.

db/collection_test.go
sanderpick: Yep haha... I had a similar thought. Would `ReadTransform` or `ReadTransformer` be better?

db/db.go
sanderpick: Agree! Maybe I can add something to the readme that stresses this fact. I think a well written validator in many cases can avoid this problem when combined with the lower lever thread ACL... say, only the thread starter can alter some paths. Of course, this requires the user to think deeply about this stuff, which isn't ideal.

db/collection_test.go
jsign: Uhm, yeah, but not sure how to conciliate both since "Transformer" doesn't sound like it could filter.
`ReadFilterTransformer` might be too much?

db/db.go
jsign: Totally agree.

db/collection.go
sanderpick: Yep, function params have to be evaluated as `goja.Value` before they can be understood internally.

db/bench_test.go
```diff
@@ -127,7 +127,7 @@ func BenchmarkNoIndexSave(b *testing.B) {
 
 	b.ResetTimer()
 
-	for i := 0; i < b.N; i++ {
+	for i := 0; i < len(res); i++ {
```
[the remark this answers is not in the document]
sanderpick: Oh, so nest the loops. I'm not familiar with the bench setups

db/collection.go
sanderpick: At this point, there was no error running the function... we got a return value, it was just `nil` (`undefined` in JS land). I'm just taking that to mean a falsey value. That's all to say, the function must explicitly `return true` to let the record pass.

db/collection.go
sanderpick: I kept them separate to avoid having to do some other checking to construct the error message. If the function returned an object of any kind, a `number`, or a `string`, that will be used as the error message. If they returned `false` or `undefined`, they just get the default error message (no wrapping).

db/collection.go
sanderpick: We could check for whether or not `out` is _not_ a bool, string, int, etc... since that would be weird to have a primitive _as_ the instance, but I that feels a little heavy handed given that it is valid JSON still. What do you think?

db/collection.go
sanderpick: Nice, I'll create a couple issues for this... I definitely should have cached the `goja.Runtime` and the compiled `goja.Callable` object on `Collection`.

db/bench_test.go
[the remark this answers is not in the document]
jsign: Yep. Depending on how much time it takes to benchmark the action, the Go tooling might call it again with a greater `b.N`, so it can calculate a significant per/op time doing `totaltime/b.N`.

db/collection.go
jsign: Ah, ok, so `false` and `undefined` are considered `false`. 👍🏼

db/collection.go
jsign: Cool, same "undefined" thing as above.

db/collection.go
sanderpick: Yep, JS is [pretty wild](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

db/collection.go
jsign: Dismiss my comment... afterward, when I see that filter can mutate things, I got a better sense of what this return is. Looks good to me as is!

db/collection.go
jsign: Nice, sounds good.

jsonpatcher/jsonpatcher.go
```diff
@@ -220,21 +232,21 @@ func saveEvent(id core.InstanceID, prev []byte, curr []byte) (*operation, error)
...
-		Type:       delete,
+		Type:       del,
 		InstanceID: id,
 		JSONPatch:  nil,
 	}, nil
 }
 
 type patchEvent struct {
-	Timestamp      time.Time
+	Timestamp      int64
```
requilence, line 242: @sanderpick FYI this change breaks the backward compatibility with existing logs/dbs...
Not sure if you have an intent to promise this by the moment. 

Let me know, I can propose PR to support the old format

sanderpick: Hrm, okay. Let's see... what do you propose? To add some context, `time.Time` isn't cbor-serializable since it doesn't export any fields.

requilence: You can try to cbor-serialize the old-format patchEvent(with time.Time) and then try to deserialize it into the new format(with int64) or vice versa. You will get unmarshal error.
Here is the proof to test: 
https://gist.github.com/requilence/d80e0eb1e346a8b40403e8593163aabf

sanderpick: Added a comment to your gist: https://gist.github.com/requilence/d80e0eb1e346a8b40403e8593163aabf#gistcomment-3438857

requilence: @sanderpick thanks, this solution you've proposed seems ok, btw older clients will not be able to read the data created with the newer version.

Personally, I don't like the way these cbor-serialized structs become no-upgradable at all. In case of any struct amend, you need to preserve the old struct and still break the non-upgraded clients that read the same data.

On the other hand, PB-encoded ipld nodes look more solid as it provides the way to add new fields and remove the old ones as far as you don't reuse the same indexes. It also ignores the extra fields, which means compatible with older clients. Maybe you can provide some context why did you prefer cbor?

Anyway, it may be also useful to add some comments on all cbor-encoded structs that it shouldn't be changed

sanderpick: Sorry for the delay. Coming up for air :) Agreed, the upgrade-ability is poor. IMO, CBOR is much nicer to work with compared to proto nodes in every other way. In any case, it would be an undertaking to migrate to proto nodes. Ideally, we could use any kind of `format.Node` at the `net` layer.

We could introduce a `Version` to the `core.Event` to make the handling a little cleaner. Still, older nodes are out of luck on this one.

sanderpick: I ended up going with a simpler solution: https://github.com/textileio/go-threads/pull/439

PR #430, opened 2020-09-01
revert patchEvent field change

requilence:
the change merged in the https://github.com/textileio/go-threads/pull/416#discussion_r481080485, breaks the backward compatibility with the existing db's records.

This PR aims to revert this until we have consensus with the Textile team


requilence:
I've just opened it by mistake in the textile repo instead of the forked one

PR #432, opened 2020-09-09
Token validate fatal error fix

requilence:
We've faced the problem. In case you have some `net.Validate()` requests in the background and do `netBoostrapper.Close()` this leads to a fatal error because datastore is closed and it is not possible to retrieve the identity.

Looks like it safe to have the nil check before

### DIFF
--- a/core/thread/identity.go
+++ b/core/thread/identity.go
@@ -214,13 +214,14 @@ func (t Token) PubKey() (PubKey, error) {
 // If token is not present, both the returned public key and error will be nil.
 func (t Token) Validate(issuer crypto.PrivKey) (PubKey, error) {
 	var ok bool
+	if t == "" {
+		return nil, nil
+	}
 	issuer, ok = issuer.(*crypto.Ed25519PrivateKey)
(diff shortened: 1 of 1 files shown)


jsign:
I'm don't have this code that fresh in my mind, but leaving a point here for you @requilence  and @sanderpick .

Looks to me the assertion fails because the result is a `nil` value which isn't the expected type?
Might be better to check `if issuer == nil` and handle this case that way?

This PR  handles the `issuer == nil` being an unexpected `nil` type by checking if `t == ""`. Which implies a relationship between `t == ""` and `issuer` being in some border case. Maybe that's the correct thing... but looks weird.

sanderpick:
How about returning an error if issuer is `nil`? That seems like a more straightforward take on the situation.

sanderpick:
Hey @requilence, I snuck in an `issuer` check [here](https://github.com/textileio/go-threads/blob/master/core/thread/identity.go#L216). I'll close this now, but feel free to reopen if that solution doesn't work for you.

PR #435, opened 2020-09-12
Eliminate semaphore contention

dgtony:
Using go-threads on a high-degree nodes we detected there is a scalability problem with a current approach of protecting thread operations with semaphores. Currently semaphore remains acquired by an active process during entire flow, including cyclic fetches of record blocks from the network, app-level validation etc. Sometimes it may result in a long time holding the semaphore. Unfortunately several processes trying to acquire the same semaphore become naturally queued even further amplifying waiting times.

To give an idea on how bad it can be, here are the quantiles of per-thread semaphore acquiring time measured in a sliding window during the day and a week respectively:
acquire_duration_day
acquire_duration_week

In the end it prevents normal system operation and leads to a lot of requests failed due to exceeded deadlines.

Here we propose a solution to the problem by decoupling long-running operations and synchronized sections. Network operations and record validation are performed without acquiring the thread semaphore, but concurrent log pullings are still avoided with per-thread-per-log semaphores. Per-thread semaphores now protect only local thread state mutations, record processing and internal bus broadcasting to preserve order of the records.


sanderpick:
@dgtony, this is great! Thanks for diving so deep. I am back at the desk this week. I gave it a skim but need to look deeper tomorrow.

net/net.go
```diff
@@ -826,11 +824,25 @@ func (n *net) Validate(id thread.ID, token thread.Token, readOnly bool) (thread.
 	return token.Validate(n.getPrivKey())
 }
 
-// getConnector returns the connector tied to the thread if it exists
+func (n *net) addConnector(id thread.ID, conn *app.Connector) {
```
sanderpick, line 827: Ooph, good catch on the unprotected map access 🤦

net/net.go
```diff
@@ -848,110 +860,183 @@ func (n *net) PutRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
...
+
+		if appConnected {
+			if err := connector.HandleNetRecord(ctx, record); err != nil {
+				// Future improvement notes.
+				// If record handling fails there are two options available:
+				// 1. Just interrupt and return error (current behaviour). Log head remains moved and some events
+				//    from the record possibly won't reach reducers/listeners or even get dispatched.
+				// 2. Rollback log head to the previous record. In this case record handling will be retried until
+				//    success, but reducers must guarantee its idempotence and there is a chance of getting stuck
+				//    with bad event and not making any progress at all.
```
sanderpick, line 899: Good notes! The original thinking here is inline with your warning in (2)... if it failed the first time, chances are it will fail again.

net/net.go
```diff
@@ -848,110 +860,183 @@ func (n *net) PutRecord(ctx context.Context, id thread.ID, lid peer.ID, rec core
...
+				return fmt.Errorf("handling record failed: %w", err)
+			}
+		}
+
+		// Generally broadcasting should not block for too long, i.e. we have to run it
+		// under the semaphore to ensure consistent order seen by the listeners. Record
+		// bursts could be overcome by adjusting listener buffers (EventBusCapacity).
```
sanderpick, line 906: 👍

net/net.go
```diff
@@ -426,29 +416,40 @@ func (n *net) pullThreadUnsafe(ctx context.Context, id thread.ID) error {
...
 			}
+
 			lock.Lock()
 			fetchedRcs = append(fetchedRcs, recs)
 			lock.Unlock()
 		}(lg)
 	}
 	wg.Wait()
+
+	// maybe we should preliminary deduplicate records?
```
sanderpick, line 448: As in, check for duplicates in `fetchedRcs`?

sanderpick (approved):
Thanks, @dgtony, this is a big improvement. I left a couple questions, but LGTM!

sanderpick:
Oop, I left you with conflict... I think just from this tiny change: https://github.com/textileio/go-threads/commit/f4a9ab7691f98685b58a610a64dbfc85ccadb0d6#diff-2fef49f8571a7dfa93a40dec78adc852

dgtony:
Here is a little update - take a look, please.

We've been testing this branch during the last week and discovered that in a presence of large amount of threads, semaphore protection performs better on a thread-level rather than on a log-level. So there are two types of semaphores introduced: first protects against concurrent pulls of the same thread, and another one guards updates of local thread state.

net/net.go
dgtony: yes, in order to remove redundant processing of already seen records

net/client.go
```diff
@@ -201,81 +215,89 @@ func (s *server) getRecords(ctx context.Context, id thread.ID, lid peer.ID, offs
...
+				return nil
 			}
+
 			for _, l := range reply.Logs {
-				log.Debugf("received %d records in log %s from %s", len(l.Records), l.LogID.ID, p)
+				var logID = l.LogID.ID
+				log.Debugf("received %d records in log %s from %s", len(l.Records), logID, pid)
+
+				if l.Log != nil && len(l.Log.Addrs) > 0 {
+					if err = s.net.store.AddAddrs(tid, logID, addrsFromProto(l.Log.Addrs), pstore.PermanentAddrTTL); err != nil {
```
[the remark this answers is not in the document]
dgtony: Mostly to avoid moving the head of the log. Method `getRecords` now just fetches the records and except adding new addresses and keys is stateless, real processing and state update happens in `putRecord`.

net/client.go
```diff
@@ -315,49 +337,43 @@ func (s *server) pushRecord(ctx context.Context, id thread.ID, lid peer.ID, rec
 		Body: body,
 	}
 
+	logErr := func(addr ma.Multiaddr, f func(addr ma.Multiaddr) error) {
```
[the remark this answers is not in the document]
dgtony: sure

sanderpick:
Cool, looks reasonable. @jsign, since you worked on the initial concurrency model, could you give this a skim?

jsign:
Well, a lot of things happened since that so I'm not very gymnastic with all the changed code since then. 
But sure, I'll give a quick look to today if I see any catches.

sanderpick:
Cool, makes sense. Even just a 2 minute skim as a sanity check

jsign (approved):
Looks good.

Every time I see the `TryAcquire` defaulting to do nothing if the lock is already taken, making that behavior indistinguishable from a real update seems a bit weird. (btw, was always that way).

Maybe something for the future is to return some extra bool to indicate if real work was done or the call was ignored, since the caller can't really know after having a successful return if something _really_ happened or maybe got unlucky and should try again _soon_.

Anyway, looks like if this hasn't happened yet is because callers doesn't want to have that guarantee. Or maybe that's something that callers doesn't have that clear, since in general when you're calling a method you wouldn't expect that "it might do nothing" and you should retry later (without any way to identify that). Looks like this magical "ignore" style might be error-prone.

Just sharing some thoughts here. Not fully related with this PR change really.

sanderpick:
Great, good stuff to think about 👍 

@dgtony, are you still testing or do you consider this mergeable?

dgtony:
We've done with testing, latest version looks stable.

Regarding silent `TryAcquire`. In a first draft implementaton methods returned specific error in case of failed semaphore acquisition, to let caller know about skipped processing. But later I removed it, because in fact `TryAcquire` is currently used only to ensure there is no more than a single process pulling given thread from the other nodes. So if you come and see semaphore is already busy, that's fine because no matter what process will pull the thread, and error checks just cluttered the code.

But as you said, problem may arise in a future if some caller will assume synchronism, e.g. that thread was definitely updated after return from the `pullThread`. There seems to be no such assumptions currently in the code, but I can bring the error back, to make it explicit, if you think it's worth it.

jsign:
> But as you said, problem may arise in a future if some caller will assume synchronism, e.g. that thread was definitely updated after return from the `pullThread`. There seems to be no such assumptions currently in the code, but I can bring the error back, to make it explicit, if you think it's worth it.

Yes, until there's some good argument to include a signal that the call was skipped I think it's ok as it is. 👍🏼

PR #436, opened 2020-09-15
DB: Adds a verify save method, filter read before creating save actions

sanderpick:
- When generating patches, the current value must be filtered for the given identity otherwise weird deletions can happen (and they will get permission denied).
- Adds a `Verify` method that function much like `Save` but does _not_ commit the generated actions. This is useful in a transaction to know if a future action will be permitted.


db/collection.go
```diff
@@ -404,42 +419,86 @@ func (t *Txn) Create(new ...[]byte) ([]core.InstanceID, error) {
 	return results, nil
 }
 
+// Verify verifies updated instances but does not save them.
+func (t *Txn) Verify(updated ...[]byte) error {
```
sanderpick, line 423: This is the bulk of the verify features (most of the related changes for the API+client)

db/collection.go
```diff
@@ -404,42 +419,86 @@ func (t *Txn) Create(new ...[]byte) ([]core.InstanceID, error) {
...
 			InstanceID:     id,
 			CollectionName: t.collection.name,
-			Previous:       beforeBytes,
-			Current:        item,
+			Previous:       previous,
+			Current:        next,
 		})
 	}
-	return nil
+	return actions, nil
```
sanderpick, line 501: `Save` will add these actions to the transaction's pending `actions`, `Verify` just uses them to validate events.

db/collection_test.go
```diff
@@ -107,18 +107,39 @@ func TestNewCollection(t *testing.T) {
 		db, clean := createTestDB(t)
 		defer clean()
 		c, err := db.NewCollection(CollectionConfig{
-			Name:           "Dog",
-			Schema:         util.SchemaFromInstance(&Dog{}, false),
-			WriteValidator: `return true`,
+			Name:   "Dog",
+			Schema: util.SchemaFromInstance(&Dog{}, false),
+			WriteValidator: `
```
sanderpick, line 112: A more complete test for `WriteValidator`...

db/collection.go
```diff
@@ -404,42 +419,86 @@ func (t *Txn) Create(new ...[]byte) ([]core.InstanceID, error) {
...
-		t.actions = append(t.actions, core.Action{
+		actions = append(actions, core.Action{
 			Type:           core.Save,
 			InstanceID:     id,
 			CollectionName: t.collection.name,
-			Previous:       beforeBytes,
-			Current:        item,
+			Previous:       previous,
+			Current:        next,
 		})
```
jsign, line 499: I don't feel much confident in having the `go-threads` logic in my head anymore, most prob since I only do big-chunk reviews every some weeks, so a bit hard to follow up. Sorry for, probably, repeated/dumb questions.

So here the patch will be calculated from the filtered/transformed instance regarding the identity. 
Can this break in an unwanted way what other identity with more or less privileges about that same instance when applying this patch?
Saying it differently: are patches generated from different identities with different permissions compatible resulting in wanted changes for all users in the same thread?

jsign (approved):
LGTM!
Left some question

db/collection.go
sanderpick: Yep, well... that's what this change prevents. An identity can only create a patch against its own "view" of the instance, which means that differences are _zeroed_ in the patch.

api/client/write.go
```diff
@@ -201,5 +231,11 @@ func (t *WriteTransaction) Delete(instanceIDs ...string) error {
 
 // end ends the active transaction.
 func (t *WriteTransaction) end() error {
-	return t.client.CloseSend()
+	if err := t.client.CloseSend(); err != nil {
+		return err
+	}
+	if _, err := t.client.Recv(); err != nil && err != io.EOF {
+		return err
```
sanderpick, line 238: Buckets is now using transactions... I noticed that after ending the transaction from the client, server errors are swallowed 😨 . cc @asutula

db/collection.go
jsign: Ah I see. Doing the diff on the original state of the change, not the full one since might be interpreted as wiping hidden attributes

sanderpick: Yep, exactly

PR #437, opened 2020-09-16
db: add txn discard to api

sanderpick:
Signed-off-by: Sander Pick <sanderpick@gmail.com>


api/client/client_test.go
```diff
@@ -748,6 +748,49 @@ func TestClient_WriteTransaction(t *testing.T) {
...
+		}()
+
+		existingPerson.Age = 38
+
+		err = txn.Discard()
+		if err != nil {
+			t.Fatalf("failed to write txn discard: %v", err)
+		}
+
+		err = txn.Save(existingPerson)
```
carsonfarmer, line 789: To be clear, this method discards the changes, but it _doesn't_ abort the transaction itself, correct? To abort the transaction, it looks like we need to close the context, is that right?

carsonfarmer (approved):

api/client/client_test.go
carsonfarmer: I ask because from a remote client perspective, this doesn't appear to be possible?

sanderpick: Yep, exactly, it's still "alive" but will error when it's committed using `End`

carsonfarmer: Perfect!

api/client/client_test.go
```diff
@@ -748,6 +748,49 @@ func TestClient_WriteTransaction(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("failed to create write txn: %v", err)
+		}
+
+		end, err := txn.Start()
+		if err != nil {
+			t.Fatalf("failed to start write txn: %v", err)
+		}
+		defer func() {
+			err = end()
```
sanderpick, line 776: @carsonfarmer this is calling the func returned by the Go client that actually just sends a `CloseSend` to the server, which makes it release the transaction... that means it will be committed, but will error if it has been marked as "discarded"

sanderpick: You can see how the Go client "ends" the transaction here: https://github.com/textileio/go-threads/blob/master/api/client/write.go#L233 (note the extra error catch...)

carsonfarmer: Yeh this is exactly what I needed. Perfect. One thing I noticed though, in the js client, we're not properly getting the final response as in here: https://github.com/textileio/go-threads/blob/master/api/client/write.go#L237 which we need to ensure the transaction was successful or not.

carsonfarmer: So this has been a useful thread/discussion. Cheers!

sanderpick: Yeah totally, I ran into that yesterday: https://github.com/textileio/go-threads/pull/436#discussion_r489016111

PR #438, opened 2020-09-17
db: adds mid txn errors

sanderpick:
Right now, when you call methods on a transaction like `Save`, `Delete`, etc., an error will end the txn, which is cumbersome in cases where you want to handle that error and retry (w/o recreating the whole transaction).

Easiest solution: Add an error to all the reply objects (CreateReply, etc.)


net/net.go
```diff
@@ -386,7 +386,6 @@ func (n *net) PullThread(ctx context.Context, id thread.ID, opts ...core.ThreadO
 }
 
 func (n *net) pullThread(ctx context.Context, id thread.ID) error {
-	log.Debugf("pulling thread %s...", id)
```
sanderpick, line 389: This creates too much output, to the point where you can't easily make sense of the logs.

db/collection.go
```diff
@@ -619,6 +619,17 @@ func (t *Txn) Discard() {
 	t.discarded = true
 }
 
+// RefreshCollection updates the transaction's collection reference from the master db map,
+// which may have received updates while the transaction is open.
+func (t *Txn) RefreshCollection() error {
```
sanderpick, line 624: This is needed to pick up collection changes during a transaction.

sanderpick:
Oh, note that the new errors are only used during a transaction... normal non-txn calls return the error as before.

db/collection.go
carsonfarmer: This seems odd given we're using a transaction here :P But I guess this actually makes sense, given our transactions provide atomicity but not isolation guarantees!

carsonfarmer (approved):
This is great @sanderpick, exactly what we need for remote offline-frist updates tbh.

db/collection.go
sanderpick: Yeah, it's kinda weird :/ Though, the transaction changes themselves and the underlying state are still "locked" during the transaction, as you stated differently above. Really I just needed the ability to update the collection after catching a schema mismatch error _during the txn_.

sanderpick:
Thanks for the late night look!

bitcard:
mark

PR #439, opened 2020-09-18
jsonpatcher: empty interface timestamp

sanderpick:
See https://github.com/textileio/go-threads/pull/416#discussion_r481080485

After trying a few different ways to fix this, I landed on just using `interface{}` for `Timestamp`, which allows us to continue bulk-decoding events with potentially mixed `Timestamp` types.

### DIFF
--- a/jsonpatcher/jsonpatcher.go
+++ b/jsonpatcher/jsonpatcher.go
@@ -51,8 +51,7 @@ type operation struct {
 	JSONPatch  []byte
 }
 
-type jsonPatcher struct {
-}
+type jsonPatcher struct{}
 
 var _ core.EventCodec = (*jsonPatcher)(nil)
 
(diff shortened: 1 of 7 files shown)


jsonpatcher/jsonpatcher.go
```diff
@@ -239,20 +238,36 @@ func deleteEvent(id core.InstanceID) (*operation, error) {
...
-	_ = binary.Write(buf, binary.BigEndian, t)
+	_ = binary.Write(buf, binary.BigEndian, nanos)
 	return buf.Bytes()
 }
 
+func (je patchEvent) time() (t time.Time) {
+	switch ts := je.Timestamp.(type) {
+	case time.Time:
+		t = ts
+	case int:
```
jsign, line 265: Should `s/int/int64` considering symmetry with `Time()` switch case?

jsign (approved):
LGTM! with a comment.

jsonpatcher/jsonpatcher.go
sanderpick: I did that initially, but the type actually comes back as `int` when it's set on an empty interface value.

requilence:
@sanderpick I've just tested this fix and and now I got the gob encode error for old-format time in the `dispatcher.Dispatch()` : ```gob: type not registered for interface: map[string]interface {}```

So looks like empty time.Time decoded into map[string]interface which is not auto-registred in the gob

PR #444, opened 2020-09-22
jsonpatcher: fix gob-encode of old ts events

### LINKED ISSUE #443: dispatcher broken for old-style event with time.Time
I've just tested https://github.com/textileio/go-threads/pull/439 and got the gob encode error in the dispatcher for old-format event [here](https://github.com/textileio/go-threads/blob/master/db/dispatcher.go#L80)
```gob: type not registered for interface: map[string]interface {}```

So looks like an empty time.Time decoded into map[string]interface which is not auto-registred in the gob

### PR DESCRIPTION
**Pull request by requilence**
fixes #443

### DIFF
--- a/jsonpatcher/jsonpatcher_test.go
+++ b/jsonpatcher/jsonpatcher_test.go
@@ -71,7 +71,7 @@ func TestJsonPatcher_Migration(t *testing.T) {
 
 	b.Reset()
 	e = gob.NewEncoder(&b)
-	if err := e.Encode(tsout); err != nil {
+	if err := e.Encode(out); err != nil {
 		t.Errorf("failed to gob encode event with int64 timestamp: %v", err)
 	}
 
--- a/jsonpatcher/jsonpatcher_test.go
+++ b/jsonpatcher/jsonpatcher_test.go
@@ -1,6 +1,8 @@
 package jsonpatcher
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by requilence** [comment]
I've added gob.Encode checks into the existing test in `jsonpatcher_test.go` instead of separate test in the `db` pkg. This seems not very correct from, because this way we also need to do gob.Register in test so we don't actually test the `db` pkg. Let me know if you are ok with it, or it's better to do a separate test

**Review by sanderpick — APPROVED** [review]
Sweet, thanks!

**Comment by sanderpick** [comment]
I'm okay with this test... thanks 🙏

File: jsonpatcher/jsonpatcher_test.go [hunk 1]
```diff
@@ -35,18 +44,37 @@ func TestJsonPatcher_Migration(t *testing.T) {
...
+
 	err = cbornode.DecodeInto(node2.RawData(), &out)
 	if err != nil {
 		t.Errorf("failed to decode event with int64 timestamp: %v", err)
 	}
 	tsout := out.time()
+
+	b.Reset()
+	e = gob.NewEncoder(&b)
+	if err := e.Encode(tsout); err != nil {
```
sanderpick, line 74: Did you mean for this to be `e.Encode(out)` as above?

requilence: yeap, fixed

Issue #441, opened 2020-09-18
Should calling delete on a non-existent instance just be silently ignored?


carsonfarmer:
Right now, this throws an error, but should it? For remote clients that don't know ahead of time if an instance has already been deleted, it would be useful if this just failed silently. Either way the instance is no longer in the collection. Are there unexpected consequences to this that we need to consider? Now that we have errors coming back from transactions, maybe this isn't such a problem?

PR #450, opened 2020-10-08
Local first optimizations

carsonfarmer:
This PR contains a few small nits and optimizations that make building on offline-first client for ThreadDB easier. The main changes are that a call to Save can now also lead to a Create action being generated if and only if the call to Save includes an existing Instance ID as a simple safety check. This allows callers to write or overwrite an existing Instance, without having to check first if it already exists. An additional change is that Delete now silently ignores missing Instances. This means a client can call delete on an Instance that might have already been deleted while it was offline or similar, without throwing an error.

Finally, the more important addition is the "special" _mod field. This is simple a "readonly" timestamp that gets added to all Instances. Any changes to this field will be ignored by the peer, and will be automatically updated on Save (and initialized on Create). This doesn't really have any effect on Delete, which is partly why the ModifiedSince experimental API was also created.

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -210,28 +210,17 @@ func (c *Collection) DeleteMany(ids []core.InstanceID, opts ...TxnOption) error 
 }
 
 // Save saves changes of an instance in the collection.
-func (c *Collection) Save(v []byte, opts ...TxnOption) (id core.InstanceID, err error) {
-	err = c.WriteTxn(func(txn *Txn) error {
-		var ids []core.InstanceID
-		ids, err = txn.Save(v)
-		if err != nil {
-			return err
-		}
-		if len(ids) > 0 {
-			id = ids[0]
-		}
-		return nil
+func (c *Collection) Save(v []byte, opts ...TxnOption) error {
+	return c.WriteTxn(func(txn *Txn) error {
+		return txn.Save(v)
 	}, opts...)
-	return
 }
 
 // SaveMany saves changes of multiple instances in the collection.
-func (c *Collection) SaveMany(vs [][]byte, opts ...TxnOption) (ids []core.InstanceID, err error) {
-	err = c.WriteTxn(func(txn *Txn) error {
-		ids, err = txn.Save(vs...)
-		return err
+func (c *Collection) SaveMany(vs [][]byte, opts ...TxnOption) error {
+	return c.WriteTxn(func(txn *Txn) error {
+		return txn.Save(vs...)
 	}, opts...)
-	return
 }
 
 // Verify verifies changes of an instance in the collection.
@@ -524,22 +513,18 @@ func (t *Txn) Verify(updated ...[]byte) error {
 }
 
 // Save saves an instance changes to be committed when the current transaction commits.
-func (t *Txn) Save(updated ...[]byte) ([]core.InstanceID, error) {
+func (t *Txn) Save(updated ...[]byte) error {
 	identity, err := t.token.PubKey()
 	if err != nil {
-		return nil, err
+		return err
 	}
-	results := make([]core.InstanceID, len(updated))
 	actions, err := t.createSaveActions(identity, updated...)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	t.actions = append(t.actions, actions...)
-	for i, a := range actions {
-		results[i] = a.InstanceID
-	}
-	return results, nil
+	return nil
 }
 
 func (t *Txn) createSaveActions(identity thread.PubKey, updated ...[]byte) ([]core.Action, error) {
--- a/db/collection.go
+++ b/db/collection.go
@@ -5,14 +5,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/alecthomas/jsonschema"
 	"github.com/dop251/goja"
 	jsonpatch "github.com/evanphx/json-patch"
-	"github.com/ipfs/go-ipld-format"
+	format "github.com/ipfs/go-ipld-format"
 	ds "github.com/textileio/go-datastore"
+	"github.com/textileio/go-datastore/query"
 	"github.com/textileio/go-threads/core/app"
 	core "github.com/textileio/go-threads/core/db"
 	"github.com/textileio/go-threads/core/thread"
@@ -36,9 +39,9 @@ var (
 	ErrInvalidSchemaInstance = errors.New("instance doesn't correspond to schema")
 
 	errMissingInstanceID           = errors.New("invalid instance: missing _id attribute")
+	errMissingModTag               = errors.New("invalid instance: missing _mod attribute")
 	errAlreadyDiscardedCommitedTxn = errors.New("can't commit discarded/committed txn")
 	errCantCreateExistingInstance  = errors.New("can't create already existing instance")
-	errCantSaveNonExistentInstance = errors.New("can't save unkown instance")
 
 	baseKey = dsPrefix.ChildString("collection")
 )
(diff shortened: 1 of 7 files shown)


db/collection.go
```diff
@@ -473,14 +541,28 @@ func (t *Txn) createSaveActions(identity thread.PubKey, updated ...[]byte) ([]co
...
+		// it has to have a valid _id ahead of time.
 		id, err := getInstanceID(next)
 		if err != nil {
 			return nil, err
 		}
 		key := baseKey.ChildString(t.collection.name).ChildString(id.String())
 		previous, err := t.collection.db.datastore.Get(key)
 		if err == ds.ErrNotFound {
-			return nil, errCantSaveNonExistentInstance
+			a := core.Action{
```
carsonfarmer, line 556: This means we can "Save" an instance even if we haven't seem it before, so long as it already has a valid _id.

db/collection.go
```diff
@@ -656,11 +739,11 @@ func getSchemaTypeProperties(jt *jsonschema.Type, defs jsonschema.Definitions) (
 	if jt.Ref != "" {
 		parts := strings.Split(jt.Ref, "/")
 		if len(parts) < 1 {
-			return nil, ErrInvalidCollectionSchema
+			return make(map[string]*jsonschema.Type), nil
```
carsonfarmer, line 742: This should still be safe, because we _always_ check and add the _id as needed. So even if the schema doesn't call for it, all ThreadDB instances _will_ get an _id field. This makes it easier to support "empty" schemas.

db/collection.go
```diff
@@ -689,6 +772,28 @@ func setNewInstanceID(t []byte) (core.InstanceID, []byte) {
 	return newID, patchedValue
 }
 
+func getModifiedTag(t []byte) (int64, error) {
```
carsonfarmer, line 775: Just a copy-pasta pretty much from the _id equivalent.

db/collection_test.go
```diff
@@ -20,12 +21,14 @@ const (
 
 type Person struct {
 	ID   core.InstanceID `json:"_id"`
+	Mod  int64           `json:"_mod"`
```
carsonfarmer, line 24: Added for tests, this _isn't_ needed by end-users. They can leave this off and things will work just fine.

db/collection_test.go
```diff
@@ -959,8 +1070,83 @@ func TestSaveInstance(t *testing.T) {
 		checkErr(t, err)
 
 		p := util.JSONFromInstance(Person{Name: "Alice", Age: 42})
-		if err := m.Save(p); !errors.Is(err, errCantSaveNonExistentInstance) {
-			t.Fatal("shouldn't save non-existent instasnce")
+		if err := m.Save(p); err != nil {
```
carsonfarmer, line 1073: Adjusting to allow save without prior instance.

db/collection_test.go
```diff
@@ -995,8 +1181,8 @@ func TestDeleteInstance(t *testing.T) {
 	}
 
 	// Try to delete again
-	if err = c.Delete(res[0]); err != ErrInstanceNotFound {
-		t.Fatalf("cant't delete non-existent instance")
+	if err = c.Delete(res[0]); err != nil {
```
carsonfarmer, line 1184: This is up for discussion! I needed to be able to delete an instance that _maybe_ exists. In some cases, a remote client that has local changes but hasn't pushed them all yet might try to delete an instance that doesn't yet exist on the remote. This essentially just silently ignores this case. It seems reasonable: you wanted the instance gone, and it is, it just wasn't deleted by you right now ;)

db/db.go
```diff
@@ -60,7 +61,7 @@ var (
 func init() {
 	nameRx = regexp.MustCompile(`^[A-Za-z0-9]+(?:[-][A-Za-z0-9]+)*$`)
 	// register empty map in order to gob-encode old-format events with non-encodable time.Time which cbor decodes as map[string]interface{}
-	gob.Register(map[string]interface {}{})
+	gob.Register(map[string]interface{}{})
```
carsonfarmer, line 64: linter I guess?

db/db_test.go
```diff
@@ -390,7 +392,8 @@ func runListenersComplexUseCase(t *testing.T, los ...ListenOption) []Action {
 	// Collection2 Save j1
 	j1 = util.SetJSONProperty("Counter", -1, j1)
 	j1 = util.SetJSONProperty("Name", "Textile33", j1)
-	checkErr(t, c2.Save(j1))
+	err = c2.Save(j1)
```
carsonfarmer, line 395: These are leftover from some earlier changes. I can revert them if we want.

db/dispatcher.go
```diff
@@ -43,10 +43,14 @@ func newDispatcher(store datastore.TxnDatastore) *dispatcher {
 }
 
 // Store returns the internal event store.
-func (d *dispatcher) Store() datastore.Datastore {
+func (d *dispatcher) Store() datastore.TxnDatastore {
```
carsonfarmer, line 46: Always has been a txndatastore in practice, now I want to make this explicit.

db/dispatcher.go
```diff
@@ -43,10 +43,14 @@ func newDispatcher(store datastore.TxnDatastore) *dispatcher {
...
 	return d.store
 }
 
+func (d *dispatcher) Lock() *sync.RWMutex {
```
carsonfarmer, line 50: Just handy

db/dispatcher.go
```diff
@@ -122,7 +126,8 @@ func getKey(event core.Event) (key datastore.Key, err error) {
 	}
 	time := strconv.FormatInt(unix, 10)
 	key = dsDispatcherPrefix.ChildString(time).
-		ChildString(event.InstanceID().String()).
-		ChildString(event.Collection())
+		ChildString(event.Collection()).
```
carsonfarmer, line 129: Up for discussion, but it defs makes queries for specific collection types within a db easier, and a lot more efficient.

carsonfarmer (commented):
Some comments for when this thing is no longer WIP.

db/collection.go
```diff
@@ -263,6 +266,68 @@ func (c *Collection) Find(q *Query, opts ...TxnOption) (instances [][]byte, err
...
+	key := ds.NewKey(e.Key)
+	base := key.Parent().BaseNamespace()
+	number, _ := strconv.Atoi(base)
+	kind := key.Type()
+	return kind == f.Collection && number > f.Time
+
+}
+
+// ModifiedSince returns a list of all instances that have been modified (and/or touched) since `time`.
+func (c *Collection) ModifiedSince(time int64, opts ...TxnOption) (modified []core.InstanceID, err error) {
```
carsonfarmer, line 285: Note that I have _not_ yet exposed this via any gRPC APIs/updates. Partly because I can work around having this for now using less efficient means, and partly because it should be reviewed before I bother exposing it.

db/collection.go
carsonfarmer: This is useful for reasons I've outlined elsewhere (see PR description), but additionally, it circumvents a restriction on current Hub. With that in mind, we should _think through_ the implications of this a little bit. Seems safe on the outset.

db/collection_test.go
carsonfarmer: Context: #441

db/dispatcher.go
sanderpick: Nice, good call

db/collection.go
sanderpick: Would it make sense to combine this with `Find`, possibly as a new kind of query / condition? Then you could combine this functionality with other queries / conditions.

sanderpick (approved):
Awesome, LGTM 💪

db/collection.go
carsonfarmer: Hmmm, interesting. That would be ideal (no new APIs). I'm not sure off the top of my head how this might work given we are querying the dispatcher directly... but as you suggest, some sort of additional query condition/operator might do the trick. Will look into this. In that case, I might mark that API as experimental, not expose it via gRPC for now, and explore this alternative in a follow on PR. Seem reasonable?

sanderpick: SGTM!

PR #451, opened 2020-10-14
Balanced thread pulling

dgtony:
Here is a small improvement essential for the ThreadsDB nodes serving a lot of threads. Currently threads are pulled as follows: once in a period goroutines with `pullThread` procedure for all known threads start simultaneously, and it leads to substantial spikes in CPU/memory/network usage. We can pretty easily get rid of the spikes and equalize resource consumption by allocating pull operations evenly during the specified period.
balanced-pulling


sanderpick (approved):
Nice, LGTM! Thanks for the great contributions.

PR #460, opened 2020-10-28
Pass thread key and log key through client when calling NewDB

dmerrill6:
# Change log

- Add `threadKey`, `logKey` and `block` params to `NewDB` at threads protobuf.

- Add `threadKey` and `logKey` params to `NewDBFromAddr` at threads protobuf.

- Handle those params and pass them through to `Manager`.

# Notes

Regenerating the pb file using `make` made a lot of changes, maybe because my local version of `protoc` is newer. I tested the changes by running the client tests and manually by calling the function from the Space Daemon and all seems to be working, but if you need me to downgrade my protoc binary and regenerate let me know.

Also please let me know if this is missing anything before being ready to merge.


sanderpick (approved):
LGTM! Thanks for jumping in 🙏

PR #462, opened 2020-10-29
Better behavior when saving non-existent instance

carsonfarmer:
This PR fixes a "regression" introduced with the new Save behavior introduced with #450. It also adds a test that should catch this in the future. The issue is that our transactions don't currently provide isolation guarantees, which is particularly problematic when doing remote updates in batches. To get around this, we provide default behavior for "Saving" a non-existent instance. However, we still had an additional check on the patcher that wasn't allowing this. This now allows saves on non-existent instances, but also simplifies the logic on the collection so we aren't special casing this anymore.


sanderpick (approved):
LGTM!

PR #463, opened 2020-10-30
Hybrid logstore

dgtony:
Currently logstore requests are skewed towards the reads, while some retrievals (notably `GetThread`) are pretty expensive, requiring multiple subrequests of logstore components that goes through the series of mutexes preserving data consistency.
So we implemented a new hybrid model of logstore, which is both persistent and enables fast reads. Basically it's just a write-through in-memory cache backed by a datastore for persistency. We combined two existing implementations (lstoreds + lstoremem), such that all reads goes into in-memory store only, and writes are performed in both.
Resulting hybrid model makes request processing order of magnitude faster with a tradeoff of slower writes and increased memory consumption as we're keeping all the data in memory. Given that such a setting is not universal but would be useful for higher-rank nodes, it can be enabled with an option and lstoreds remains the default one.
As a useful side effect, now we also get export/import capabilities for all the different logstore implementations. It makes implementation of components and data exchange easier, enables dumping data for analysis etc.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -92,7 +92,7 @@ func DefaultNetwork(repoPath string, opts ...NetOption) (NetBoostrapper, error) 
 		return nil, fin.Cleanup(err)
 	}
 
-	tstore, err := buildLogstore(ctx, config.LogstoreKind, repoPath, fin)
+	tstore, err := buildLogstore(ctx, config.LSType, repoPath, fin)
 	if err != nil {
 		return nil, fin.Cleanup(err)
 	}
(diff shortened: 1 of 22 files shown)


sanderpick:
Hey @dgtony, looking good! I started looking though it. Will take another look this evening. Happy to see the importer / exporter functionality. That was actually on my to-do list 🎉

util/finalizer.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"context"
+	"io"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+// Finalizer collects resources for convenient cleanup.
+func NewFinalizer() *Finalizer {
```
sanderpick, line 11: Nice, clever.

common/common.go
```diff
@@ -123,33 +103,89 @@ func DefaultNetwork(repoPath string, opts ...NetOption) (NetBoostrapper, error)
...
+	}
+
+	if len(config.LogstoreKind) == 0 {
+		config.LogstoreKind = LogstorePersistent
+	}
+
+	return nil
+}
+
+type LogstoreKind string
```
sanderpick, line 173: How about `LogstoreType`?

sanderpick (approved):
Good idea! I didn't dive super deep into the dump/restore logic for all the books, but will be playing around with that functionality soon and will look closer.

Any ideas about the test failures?

common/common.go
dgtony: renamed

dgtony:
Regarding test failures. This one and some other random failures start periodically emerge on the master. Looks like a timing issue, e.g. introducing the 20 milliseconds sleep before checking test results (db/collection_test.go:931) seems to fix it.

PR #468, opened 2020-12-04
go-ds-mongo

sanderpick:
Allows the use of [`go-ds-mongo`](https://github.com/textileio/go-ds-mongo) for all needed persistence.


db/options.go
```diff
@@ -115,10 +76,10 @@ func WithNewEventCodec(ec core.EventCodec) NewOption {
 	}
 }
 
-// WithNewLowMem specifies whether or not to use low memory settings.
-func WithNewLowMem(low bool) NewOption {
+// WithNewToken provides authorization for interacting with a db.
+func WithNewToken(t thread.Token) NewOption {
```
jsign, line 80: Maybe _New_ can be avoided and was biased by the deleted option? i.e: `WithToken`?

docker-compose-dev.yml
```diff
@@ -12,8 +12,25 @@ services:
...
+      - "27017"
+    ports:
+      - "127.0.0.1:27017:27017"
+    command:
+      - /bin/bash
+      - -c
+      - |
+        /usr/bin/mongod --fork --logpath /var/log/mongod.log --bind_ip_all --replSet rs0
+        mongo --eval 'rs.initiate({_id: "rs0", version: 1, members: [{ _id: 0, host: "mongo:27017" }]})'
+        tail -f /var/log/mongod.log
```
jsign, line 36: Cool, what we chated in the other PR. 👍🏼

threadsd/main.go
```diff
@@ -28,7 +31,7 @@ var log = logging.Logger("threadsd")
 func main() {
 	fs := flag.NewFlagSetWithEnvPrefix(os.Args[0], "THRDS", 0)
 
-	repo := fs.String("repo", ".threads", "Repo location")
+	repo := fs.String("repo", ".threadsd", "Repo location")
```
jsign, line 34: Isn't this a breaking change? An existing `threadsd` that uses the previous default value won't find its data and start from scratch. Was this intentional?

threadsd/main.go
```diff
@@ -37,6 +40,9 @@ func main() {
 	connGracePeriod := fs.Duration("connGracePeriod", time.Second*20, "Duration a new opened connection is not subject to pruning")
 	keepAliveInterval := fs.Duration("keepAliveInterval", time.Second*5, "Websocket keepalive interval (must be >= 1s)")
 	enableNetPubsub := fs.Bool("enableNetPubsub", false, "Enables thread networking over libp2p pubsub")
+	mongoUri := fs.String("mongoUri", "", "MongoDB URI (if not provided, an embedded Badger datastore will be used)")
+	mongoDatabase := fs.String("mongoDatabase", "threaddb", "MongoDB database name")
```
jsign, line 44: I think is safer to not provide a default value, and error on startup if this isn't set explicitly through flag/env.
I fear that not providing a value might overlap with multiple instances running in the same cluster, which might be dangerous.

That's what I did in Powergate simply to avoid a misconfiguration (or simply doing nothing using defaults) that might cause a really bad situation.

threadsd/main.go
```diff
@@ -73,23 +79,40 @@ func main() {
 	log.Debugf("connGracePeriod: %v", *connGracePeriod)
 	log.Debugf("keepAliveInterval: %v", *keepAliveInterval)
 	log.Debugf("enableNetPubsub: %v", *enableNetPubsub)
+	if *mongoUri != "" {
+		log.Debugf("mongoUri: %v", *mongoUri)
```
jsign, line 83: This might be a bit of a security issue since the URI contains the password and that can end up living in some log file or external logging systems.

jsign (approved):
LGTM. Left some comments to think about.

threadsd/main.go
sanderpick: Ah yeah, reverted!

docker-compose-dev.yml
sanderpick: Yep, I added a little test util package in `go-ds-mongo`... makes it easy to start the replica set for tests: https://github.com/textileio/go-ds-mongo/blob/master/mongo_test.go#L20

threadsd/main.go
sanderpick: Switched to using the `url` package to parse the URL and then display a redacted string like ` mongoUri: mongodb://boom:xxxxx@127.0.0.1:27017`

threadsd/main.go
sanderpick: Cool, default removed

db/options.go
sanderpick: `WithNewToken` is used with the `NewOptions` when creating a new database. `WithToken` is used with normal `Options`... open to naming suggestions there but that was the best I came up with.

PR #469, opened 2020-12-08
Datastore copy util

sanderpick:
Adds a small program for copying a badger/mongo datastore to another badger/mongo datastore.

### DIFF
--- a/util/datastore/dscopy/main.go
+++ b/util/datastore/dscopy/main.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"context"
+	"net/url"
+	"os"
+
+	ds "github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/namsral/flag"
+	badger "github.com/textileio/go-ds-badger"
+	mongods "github.com/textileio/go-ds-mongo"
+)
+
+var log = logging.Logger("ds-copy")
+
+func main() {
+	fs := flag.NewFlagSet(os.Args[0], 0)
+
+	fromBadgerRepo := fs.String("fromBadgerRepo", "", "Source badger repo path")
+	toBadgerRepo := fs.String("toBadgerRepo", "", "Destination badger repo path")
+
+	fromMongoUri := fs.String("fromMongoUri", "", "Source MongoDB URI")
+	fromMongoDatabase := fs.String("fromMongoDatabase", "", "Source MongoDB database")
+	fromMongoCollection := fs.String("fromMongoCollection", "", "Source MongoDB collection")
+	toMongoUri := fs.String("toMongoUri", "", "Destination MongoDB URI")
+	toMongoDatabase := fs.String("toMongoDatabase", "", "Destination MongoDB database")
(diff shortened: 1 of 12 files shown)


util/datastore/dscopy/main.go
```diff
@@ -0,0 +1,127 @@
...
+		log.Infof("connected to mongo destination: %s", uri.Redacted())
+	}
+
+	results, err := from.Query(query.Query{})
+	if err != nil {
+		log.Fatalf("querying source: %v", err)
+	}
+	defer results.Close()
+	for r := range results.Next() {
+		if err := to.Put(ds.NewKey(r.Key), r.Value); err != nil {
```
jsign, line 118: Between these two lines we need to add:
```
if r.Error != nil {
   log.Fatal(...)
}
```

util/datastore/dscopy/main.go
```diff
@@ -0,0 +1,127 @@
...
+			log.Fatalf("copying %s: %v", r.Key, err)
+		}
```
jsign, line 120: I discovered while working and testing Powergate migrations that this is too slow for copying (at least for my taste).

[Here](https://github.com/textileio/powergate/blob/600eb5aba73f82f39c809d61f34fed0d4226dab3/migration/migration1_joblogger.go#L24) is what I did for migrations; which reduced times by some orders of magnitude. Nothing crazy, ignoring the migration logic, basically some rate-limiting to Read-and-Put up to 1000 in concurrently instead of serially.

jsign (approved):
LGTM! I left two somewhat important comments to consider.

util/datastore/dscopy/main.go
sanderpick: Ah yeah, thanks!

util/datastore/dscopy/main.go
sanderpick: Nice, copied your approach there

sanderpick:
Sorry the big code update post review here, but I realized using this tool would be much easier if it wasn't a generic datastore copier, but was specific to the structure of `go-threads`'s datastores.

util/dscopy/main.go
```diff
@@ -0,0 +1,253 @@
...
+var stores = []store{
+	{
+		name: "eventstore",
+	},
+	{
+		name: "logstore",
+	},
+	{
+		name:  "ipfslite",
+		files: []string{"key"},
```
sanderpick, line 37: Move the key as well...

util/dscopy/main.go
```diff
@@ -0,0 +1,253 @@
...
+				lock.Unlock()
+				return
+			}
+			if verbose {
+				log.Infof("copied %s", r.Key)
+			}
+			lock.Lock()
+			count++
+			lock.Unlock()
+			if count%parallel == 0 {
```
jsign, line 196: Nit: move `lock.Unlock()` to include the if case since `count` is being read without lock and might be a data-race with other goroutine owning the lock. (tongue twister)

jsign (approved):
LGTM! Left a comment.

util/dscopy/main.go
sanderpick: Nice, done

PR #478, opened 2021-01-08
Track real counts for limit and skip

### LINKED ISSUE #458: limit and skip operation is applied before conditions
Given a set of 20 instances with a schema of :

```json
{
	"properties": {
		"_id": {
			"type": "string"
		},
                "index": {
                       "type": "number"
                }
	},
	"type": "object"
}
```
with index property values that range from 1 - 20.

If I do a JSON query on the result set with a query:

```json
{
     "ands": [
         {
              "fieldPath": "index",
              "operation": 0, //Eq
              "value": {
                  "float": 15
              }
         }
     ],
     "limit": 10,
     "skip": 0,
}
```

I expect to get a list of a single result of the instance with a property of `index = 15`.

I actually get an empty list.

The limit and skip operation is applied before the conditions so if my item is not within the limit and skip operation it won't find it.
I would expect that it would search through the whole result set for the item and then apply the limit and skip operation to the result set.

carsonfarmer:
Applies Limit and Skip _after_ other conditions _and_ read filter have been applied. This leads to potentially slightly less efficient queries (because the child query engine can't handle the limits and skips directly), but provides more consistent and accurate/expected results. A test has been added that captures the issue presented in #458, plus a "regression" test was added to ensure we don't over-compensate.

Fixes #458.

Technically this is a bug fix, but it also changes how results are returned when applying limit and skip, so this might need to come with an appropriate version bump?

### DIFF
--- a/db/index.go
+++ b/db/index.go
@@ -290,8 +290,10 @@ func newIterator(txn dse.TxnExt, baseKey ds.Key, q *Query) *iterator {
 	dsq := dse.QueryExt{
 		Query: query.Query{
 			Prefix: prefix.String(),
-			Limit:  q.Limit,
-			Offset: q.Skip,
+			// Pull out Skip and Limit here because we won't know ahead of time
+			// (due to readFilters) how many to skip/limit
+			// Limit:  q.Limit,
+			// Offset: q.Skip,
 		},
 	}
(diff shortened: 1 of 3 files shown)


sanderpick (approved):
LGTM! Thanks for jumping in there 💪

PR #480, opened 2021-01-15
Fixes error when chaining skip and limit

carsonfarmer:
Fixes a simple error when skip and limit are chained, leading to what is essentially an offset. This also adds a test for this case. In addition to fixing an error, this PR makes the query more strict about what it "counts". It no longer counts instances that are set to nil by the read filter.


sanderpick (approved):
Nice!

PR #481, opened 2021-02-01
Total records limit in the GetRecords reply

dgtony:
Currently number of records returned in the `GetRecordsReply` is limited by a constant value for every log. However, if some thread contains large number of logs we can still hit on a gRPC's message size limit. Instead we can limit total number of records for all logs in the thread.


sanderpick (approved):
Nice, good catch!

PR #495, opened 2021-03-08
Exchange edges

dgtony:
Follow-up to #477.

Here is a rework of internal node communication patterns. It should significantly reduce traffic and CPU consumption of the nodes in a steady state and make it scale better. Surely, the PR is huge, but unfortunately it turned out to be impractical to divide it into the smaller ones, as all the features play well just being implemented together.

Here is the outline:
* Internal communication protocol for ThreadsDB nodes extended with a new call `ExchangeEdges`. New method enables efficient checks if thread information is in sync between any two nodes, using heads/address edges (digest), and is a replacement for continuous pulling with `GetRecords`. It also helps to reduce communication overhead with both synced and inactive peers.
* Now all calls to `GetLogs`/`GetRecords` are grouped by peers and performed through the corresponding queues, thus deduplicating and distributing peer calls over the time.
* Address edge computation implemented - deterministic digest of all peer addresses involved into the thread (similar to heads edge).
* Multiple minor fixes and improvements.

New communication logic is fully compatible with an older nodes, as it fallbacks to the `GetRecords`-based pulling if peer is lacking support of `ExchangeEdges`.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -59,7 +59,6 @@ func DefaultNetwork(opts ...NetOption) (NetBoostrapper, error) {
 	if err != nil {
 		return nil, fin.Cleanup(err)
 	}
-	fin.Add(litestore)
 
 	pstore, err := pstoreds.NewPeerstore(ctx, litestore, pstoreds.DefaultOpts())
 	if err != nil {
--- a/logstore/lstoreds/headbook.go
+++ b/logstore/lstoreds/headbook.go
@@ -2,8 +2,12 @@ package lstoreds
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
+	"math/rand"
+	"time"
 
+	"github.com/dgraph-io/badger"
 	"github.com/gogo/protobuf/proto"
 	"github.com/ipfs/go-cid"
 	ds "github.com/ipfs/go-datastore"
@@ -157,20 +161,28 @@ func (hb *dsHeadBook) ClearHeads(t thread.ID, p peer.ID) error {
 }
 
 func (hb *dsHeadBook) HeadsEdge(tid thread.ID) (uint64, error) {
+	var key = dsThreadKey(tid, hbEdge)
+	// computing and writing back previously invalidated
+	// edge frequently results in transaction conflicts
+	for attempt := 1; attempt <= 3; attempt++ {
+		edge, err := hb.getEdge(tid, key)
+		if err == nil {
+			return edge, nil
+		} else if !errors.Is(err, badger.ErrConflict) {
+			return 0, err
+		}
+		time.Sleep(time.Duration(50*attempt+rand.Intn(30)) * time.Millisecond)
+	}
+	return 0, core.ErrEdgeUnavailable
+}
+
+func (hb *dsHeadBook) getEdge(tid thread.ID, key ds.Key) (uint64, error) {
 	txn, err := hb.ds.NewTransaction(false)
 	if err != nil {
 		return 0, fmt.Errorf("error when creating txn in datastore: %w", err)
 	}
 	defer txn.Discard()
-	edge, err := hb.getEdge(txn, tid)
-	if err != nil {
-		return 0, err
-	}
-	return edge, txn.Commit()
-}
 
-func (hb *dsHeadBook) getEdge(txn ds.Txn, tid thread.ID) (uint64, error) {
-	var key = dsThreadKey(tid, hbEdge)
 	if v, err := txn.Get(key); err == nil {
 		return binary.BigEndian.Uint64(v), nil
 	} else if err != ds.ErrNotFound {
@@ -199,11 +211,15 @@ func (hb *dsHeadBook) getEdge(txn ds.Txn, tid thread.ID) (uint64, error) {
 	}
 
 	var (
-		buff [8]byte
 		edge = util.ComputeHeadsEdge(hs)
+		buff [8]byte
 	)
+
 	binary.BigEndian.PutUint64(buff[:], edge)
-	return edge, txn.Put(key, buff[:])
+	if err := txn.Put(key, buff[:]); err != nil {
+		return 0, err
+	}
+	return edge, txn.Commit()
 }
 
 func (hb *dsHeadBook) invalidateEdge(txn ds.Txn, tid thread.ID) error {
--- a/logstore/lstoreds/headbook.go
+++ b/logstore/lstoreds/headbook.go
@@ -282,7 +282,7 @@ func (hb *dsHeadBook) decodeHeadEntry(
 	withHeads bool,
 ) (tid thread.ID, lid peer.ID, heads []cid.Cid, err error) {
 	kns := ds.RawKey(entry.Key).Namespaces()
-	if len(kns) < 4 {
+	if len(kns) < 3 {
 		err = fmt.Errorf("bad headbook key detected: %s", entry.Key)
 		return
 	}
--- a/logstore/lstoremem/addr_book.go
+++ b/logstore/lstoremem/addr_book.go
@@ -4,17 +4,22 @@ import (
 	"context"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
+	"github.com/dgtony/collections/bitset"
 	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-peerstore/addr"
 	ma "github.com/multiformats/go-multiaddr"
 	core "github.com/textileio/go-threads/core/logstore"
 	"github.com/textileio/go-threads/core/thread"
+	"github.com/textileio/go-threads/util"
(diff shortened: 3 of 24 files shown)


sanderpick:
Really nice @dgtony! I gave it a skim, but would you mind adding some annotations here and there throughout the code to help me follow the big pieces? Easiest way to do that is by doing a self-review w/comments.

net/server.go
```diff
@@ -267,6 +289,80 @@ func (s *server) PushRecord(ctx context.Context, req *pb.PushRecordRequest) (*pb
...
+				ThreadID:    &pb.ProtoThreadID{ID: tid},
+				Exists:      true,
+				AddressEdge: addrsEdgeLocal,
+				HeadsEdge:   headsEdgeLocal,
+			})
+
+		case errNoAddrsEdge:
+			// requested thread doesn't exist locally
+			log.Warnf("addresses for requested thread %s not found", tid)
+			s.net.queueGetLogs.Schedule(
```
dgtony, line 331: we don't know about the thread - let's create it (later)

logstore/lstoreds/headbook.go
```diff
@@ -157,20 +161,28 @@ func (hb *dsHeadBook) ClearHeads(t thread.ID, p peer.ID) error {
 }
 
 func (hb *dsHeadBook) HeadsEdge(tid thread.ID) (uint64, error) {
+	var key = dsThreadKey(tid, hbEdge)
+	// computing and writing back previously invalidated
+	// edge frequently results in transaction conflicts
+	for attempt := 1; attempt <= 3; attempt++ {
```
dgtony, line 167: large instances would better use hybrid logstore with edges fetched from the in-memory storage to avoid performance issues

logstore/lstoremem/addr_book.go
```diff
@@ -173,40 +340,35 @@ func (mab *memoryAddrBook) AddAddrs(t thread.ID, p peer.ID, addrs []ma.Multiaddr
...
+					x.TTL = ttl
+				}
+				if exp.After(x.Expires) {
+					x.Expires = exp
+				}
 			}
 		}
+	})
+	if addrChanged {
+		go mab.segments.updateEdge(t)
```
dgtony, line 371: similar to `HeadsEdges` approach - edges in the in-memory implementation are recomputed on the thread's info change as its pretty cheap operation here

net/client.go
```diff
@@ -122,18 +122,61 @@ func (s *server) pushLog(ctx context.Context, id thread.ID, lg thread.LogInfo, p
...
+	)
+
+	// Pull from every peer
+	for _, p := range peers {
+		wg.Add(1)
+
+		go withErrLog(p, func(pid peer.ID) error {
+			defer wg.Done()
+
+			return s.net.queueGetRecords.Call(pid, tid, func(ctx context.Context, pid peer.ID, tid thread.ID) error {
```
dgtony, line 149: synchronous call

net/queue/ff.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (q *peerQueue) Size() int {
+	return len(q.index)
+}
+
+/* Call queue implementation */
+
+var _ CallQueue = (*ffQueue)(nil)
+
+type ffQueue struct {
```
dgtony, line 122: receiving, keeping and executing calls evenly and in time for the multiple peers

net/server.go
```diff
@@ -200,30 +209,43 @@ func (s *server) GetRecords(ctx context.Context, req *pb.GetRecordsRequest) (*pb
...
+			var prs = make([]*pb.Log_Record, 0, len(recs))
+			for _, r := range recs {
+				pr, err := cbor.RecordToProto(ctx, s.net, r)
+				if err != nil {
+					log.Errorf("constructing proto-record %s (thread %s, log %s): %v", r.Cid(), tid, lid, err)
+					break
+				}
+				prs = append(prs, pr)
+			}
+			if pblg == nil && len(prs) == 0 {
```
dgtony, line 231: save traffic and CPU if log is empty

net/net.go
```diff
@@ -1178,6 +1223,10 @@ func (n *net) startPulling() {
 	// it will be redefined on the next iteration
 	var interval = InitialPullInterval
 
+	// group threads by peers and exchange edges efficiently
+	var compressor = queue.NewThreadPacker(n.ctx, MaxThreadsExchanged, ExchangeCompressionTimeout)
+	go n.startExchange(compressor)
+
 PullCycle:
 	for {
 		ts, err := n.store.Threads()
```
sanderpick, line 1232: It would be great to distribute this over some time interval by paginating over the threads in an efficient manner. When there are 10s of 1000s of threads, this is a pretty heavy call.

sanderpick (approved):
Hey @dgtony, apologies for the delay here. This is a huge contribution... thank you! I left a couple comments and questions, but I'm happy to merge as is. @carsonfarmer can you give this a skim when you have a moment just so we're all on the same page about the new functionality? 

Also, I'd love to hear where the scaling bottlenecks are for your application. I have some ideas to improve things, but your usage feedback would be really helpful in deciding what direction to go. Hopefully we can have the team call that was mentioned over email soon.

common/common.go
```diff
@@ -59,7 +59,6 @@ func DefaultNetwork(opts ...NetOption) (NetBoostrapper, error) {
 	if err != nil {
 		return nil, fin.Cleanup(err)
 	}
-	fin.Add(litestore)
```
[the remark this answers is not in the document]
dgtony: It's already added in function `persistentStore`

net/queue/ff.go
```diff
@@ -0,0 +1,256 @@
...
+
+					// clear in-flight status
+					q.mx.Lock()
+					delete(q.inflight, h)
+					q.mx.Unlock()
+				}()
+
+				// spawn all overdue calls and a few ones with coming deadline
+				if remainIters := int(float64(created-deadline) / q.poll.Seconds()); remainIters > 0 &&
+					rand.Float64() > math.Sqrt(3*float64(waiting))/float64(remainIters) {
```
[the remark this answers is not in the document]
dgtony: As the reason of the issue is contention on the thread's head-keys, I believe that optimal solution depends on the intensity of writes to the logs. While contention is relatively low, I'd rather use optimistic approach with retried transactions. In case of unusually intensive writes and high contention pessimistic locking/semaphores would perform better.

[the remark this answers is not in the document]
dgtony: Regarding queue polling, that's a tricky part )
Ok, so whats going on. Here we have an underlying FIFO-queue containing every scheduled call associated with the moment it was added in the linked list. There is a predefined parameter `timeout` - every scheduled call should be spawned during such timeout, though it's not a hard deadline, just a desired property. And moreover, we want it to be as uniform over time as possible (in the presense of non-stationary process of new call arrivals).
On every polling iteration we repeatedly make a decision: pop and spawn next call from the queue or wait until the next polling iteraton. It'd be too expensive to obtain precise deadline distribution from the large queue, but since it's FIFO we know at least they are in the interval `[last_popped_deadline; timeout]`. And, of course, number of calls waiting in the queue is known as well.
Not much information, but we can use some randomized model for polling decisions and it may help to avoid synchronization with other queues. So here is the equation! It doesn't have solid theoretical underpinning, rather it's more an empirical one, reflecting two basic assumptions: probability of taking a call should be: 1) inversely proportional to the worst case deadline (when all calls in the queue share the same deadline) and 2) sublinear of a queue size.
So far its the best model I've been able to come up with in terms of smooth operation and meeting deadlines according to both our simulations and real app metrics. But I'll be happy to replace it with a more solid-ground algorithm, if any.

logstore/lstoreds/headbook.go
dgtony: As the reason of the issue is contention on the thread's head-keys, I believe that optimal solution depends on the intensity of writes to the logs. While contention is relatively low, I'd rather use optimistic approach with retried transactions. In case of unusually intensive writes and high contention pessimistic locking/semaphores would perform better.

net/queue/ff.go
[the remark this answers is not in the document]
sanderpick: Thanks for the explanation! Happy to defer to you here. We can play with it during Testground testing.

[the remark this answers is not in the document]
eightysteele: (👋🏻 @dgtony really great context here—small suggestion would be to include it in a comment for future readers!)

carsonfarmer:
Ok folks, I've given this a thrice-over, but to be honest, I didn't do a deep dive on all the updates here. I'm also happy to merge mostly as is, though I do really like @eightysteele's suggestion of some doc strings that outline the algorithm used in net/queue/ff.go, so I'd like to see that updated before we merge. Otherwise, I'm keen to just play around with this and see how it behaves.

net/queue/ff.go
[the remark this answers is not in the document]
dgtony: added comment and renamed vars for clarity

PR #508, opened 2021-05-05
Remove unnecessary signature from on-the-wire messages

merlinran:
This is the first attempt for my very first task! Passed tests under `/net` locally.


net/pb/net.proto
```diff
@@ -6,12 +6,13 @@ import "github.com/gogo/protobuf/gogoproto/gogo.proto";
...
-// Header holds a key and signature for a request.
+// Header is the message header, for now it just holds the public key of the author
 message Header {
     // pubKey is the author's public key.
     bytes pubKey = 1 [(gogoproto.customtype) = "ProtoPubKey"];
-    // signature is the message signature.
-    bytes signature = 2;
+    // signature is the message signature. Removed as these extra bytes on wire are not needed for integrity.
+    // bytes signature = 2;
+    reserved 2;
```
sanderpick, line 15: 👍

net/client.go
```diff
@@ -194,16 +191,15 @@ func (s *server) buildGetRecordsRequest(
 		Logs:       pblgs,
 	}
 
-	sig, key, err := s.signRequestBody(body)
+	key, err := s.getPubKey()
 	if err != nil {
 		err = fmt.Errorf("signing GetRecords request: %w", err)
```
sanderpick, line 196: nit: error message -> "getting public key for request: %w"

net/client.go
```diff
@@ -520,22 +513,14 @@ func (s *server) getLibp2pDialer() grpc.DialOption {
...
-func (s *server) signRequestBody(msg proto.Marshaler) (sig []byte, pk crypto.PubKey, err error) {
-	payload, err := msg.Marshal()
-	if err != nil {
-		return
-	}
+// getPubKey gets the public key associated with the server.
+func (s *server) getPubKey() (pk crypto.PubKey, err error) {
 	sk := s.net.getPrivKey()
 	if sk == nil {
 		err = fmt.Errorf("private key for host not found")
```
sanderpick, line 520: oop, old nit here: can you do `fmt.Errorf` -> `errors.New`?

net/pb/lstore.pb.go
```diff
@@ -20,7 +21,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.GoGoProtoPackageIsVersion2 // please upgrade the proto package
+const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
```
sanderpick, line 24: Is this from using a newer gogo proto package?

net/server.go
```diff
@@ -427,16 +427,6 @@ func verifyRequest(header *pb.Header, body proto.Marshaler) (pid peer.ID, err er
 		err = status.Error(codes.InvalidArgument, "bad request")
 		return
 	}
-	payload, err := body.Marshal()
-	if err != nil {
-		err = status.Error(codes.Internal, err.Error())
-		return
-	}
-	ok, err := header.PubKey.Verify(payload, header.Signature)
```
sanderpick, line 435: Sweet! I wonder if we can pull out the peer ID of the caller from the gRPC connection. The underlying transport is libp2p. That way, the caller doesn't have to send their public key at all. The issue is that without the signature we don't have a way to verify they are giving us a public key they own.

sanderpick (commented):
Looks great! Just left a couple comments to consider.

sanderpick:
cc @dgtony @requilence: Here we're removing the payload signatures in the p2p API.

net/pb/lstore.pb.go
merlinran: Oops I overlooked this one. Not sure actually. I'll dig into it

net/server.go
merlinran: Sounds like a good idea! Without signature the whole `verifyRequest()` call is meaningless it seems. I guess it's time for me to step back and have a grasp of the whole picture.

sanderpick: Yep, without the signature, we're just using `verifyRequest()` to get the caller's `peer.ID`. Here's some more context:

The peer API in threads is used to exchange thread logs and records between peers. It's a gRPC API, but the underlying protocol is libp2p. We set that up here: https://github.com/textileio/go-threads/blob/master/net/net.go#L161 (the libp2p host is passed to a handy `gostream` package.

Basically, we want to know the `peer.ID` of the caller in all the peer API server methods.

When a peer dials another peer, they setup the gRPC dial context to use their host: https://github.com/textileio/go-threads/blob/master/net/server.go#L48. That `getLibp2pDialer()` method is what leverages the _caller's_ libp2p host. 

The inner closure of `getLibp2pDialer()` has access to the host (and thus the host's `peer.ID`: `s.net.host.ID()`), but I'm not sure how that info can be extracted on the receiver's end. It might be worth inspecting the gRPC context in the server-side methods (where `verifyRequest()` is currently being called) to see if `peer.ID` gets added there. If not, we could consider PRing `gostream`.

net/pb/lstore.pb.go
merlinran: It is because I'm using the latest version of `protoc-gen-gogofaster` to generate this piece of code. The purpose here is to raise a compilation error when `github.com/gogo/protobuf/proto` package is too old and doesn't have `GoGoProtoPackageIsVersion3` defined. I think we are good.

net/server.go
merlinran: It seems that other than logging, the peer ID is used only for `CallQueue.Schedule`, specifically [`ffQueue.Schedule`](https://github.com/textileio/go-threads/blob/master/net/queue/ff.go#L150), as the index to the set of peers and inflight calls (together with thread ID). If we can somehow find another way to uniquely identify peers within the process, we should be able to get rid of the public key. Or if having bogus peer ID doesn't cause security concerns we can just use it for now, or even exchange peer IDs directly rather than public keys.

Without having a deep understanding of the code base, I have a hunch that not having assumption of the underlying protocol (libp2p) may be useful to, for example, test the logic on mocked `net.Conn`s, but I could be very naive.

sanderpick: Agreed! Spoofing the `peer.ID` (which is currently derived from the public key of the caller) _could_ be an attack vector if the caller wanted to mess up the receivers batching strategy. I looked at the incoming context to see what info is getting added there. Looks like the incoming context has a `peerKey` (gRPC peer, not libp2p peer), which via `gostream` _is actually_ the libp2p `peer.ID`. Handy! So, we can just replace calls to `verifyRequest` with something like this:

```
import (
	...
	grpcpeer "google.golang.org/grpc/peer"
)

...

func peerIDFromCtx(ctx context.Context) (peer.ID, error) {
	ctxPeer, ok := grpcpeer.FromContext(ctx)
	if !ok {
		return "", errors.New("unable to identify stream peer")
	}
	pid, err := peer.Decode(ctxPeer.Addr.String())
	if err != nil {
		return "", fmt.Errorf("parsing stream peer id: %v", err)
	}
	return pid, nil
}
```

merlinran: Wow thanks for digging into that for me! I should've inspected the context in the first place. Now that `Header` will have nothing, guess I can remove it entirely in the protobuf definition.

net/pb/net.proto
```diff
@@ -40,8 +32,8 @@ message Log {
 
 // GetLogsRequest is used to request thread logs.
 message GetLogsRequest {
-    // header is the message header.
-    Header header = 1;
+    // this was the message header.
+    reserved 1;
```
merlinran, line 36: For clarity it's better to just move the fields from the `body` field to top level, but I'm not sure if we ever want to allow old and new code to co-exist for a while.

net/pubsub.go
```diff
@@ -115,7 +115,7 @@ func (s *PubSub) Publish(ctx context.Context, id thread.ID, req *pb.PushRecordRe
 	defer s.RUnlock()
 	topic, ok := s.m[id]
 	if !ok {
-		return fmt.Errorf("thread topic not found")
+		return errors.New("thread topic not found")
```
merlinran, line 118: drive-by

net/pubsub.go
```diff
@@ -175,23 +175,15 @@ func (s *PubSub) subscribe(ctx context.Context, id thread.ID, topic *topic) {
 func (s *PubSub) handleMsg(m *pubsub.Message) (from peer.ID, rec *pb.PushRecordRequest, err error) {
 	from, err = peer.IDFromBytes(m.From)
 	if err != nil {
-		return
+		return "", nil, err
 	}
 	if from.String() == s.host.String() {
-		return
+		return "", nil, errors.New("pubsub message can not be from the host itself")
```
merlinran, line 181: From my memory, it's a good practice to return zero value when error happens, just in case if the caller doesn't check error.

net/pubsub.go
```diff
@@ -175,23 +175,15 @@ func (s *PubSub) subscribe(ctx context.Context, id thread.ID, topic *topic) {
...
 	req := new(pb.PushRecordRequest)
 	if err = proto.Unmarshal(m.Data, req); err != nil {
-		return
-	}
-	pid, err := peer.IDFromPublicKey(req.Header.PubKey)
-	if err != nil {
-		return
-	}
-	if from.String() != pid.String() {
-		log.Warnf("multicast sender does not match record header (%s != %s)", from, pid)
```
merlinran, line 193: As the peer is derived from transport, no need to check consistency now.

net/server.go
```diff
@@ -421,26 +421,15 @@ func (s *server) localEdges(tid thread.ID) (addrsEdge, headsEdge uint64, err err
 	return
 }
 
-// verifyRequest verifies that the signature associated with a request is valid.
-func verifyRequest(header *pb.Header, body proto.Marshaler) (pid peer.ID, err error) {
-	if header == nil || body == nil {
-		err = status.Error(codes.InvalidArgument, "bad request")
-		return
+// peerIDFromContext returns peer ID from the GRPC context
+func peerIDFromContext(ctx context.Context) (peer.ID, error) {
```
merlinran, line 425: This is exactly copied from your snippet @sanderpick .

net/pb/net.proto
sanderpick: Agreed! Let's just leave the `Body` field for now. We can remove it when we do a [protocol bump](https://github.com/textileio/go-threads/blob/master/core/thread/protocol.go#L17). The protocol tag is used by libp2p to differentiate p2p services.

net/server.go
sanderpick: No problem! This felt like an additional thing beyond the initial task... so didn't want to bog you down!

sanderpick (approved):
LGTM! 💥

sanderpick:
@merlinran before merging, can you remove the "[wip]" from the PR title? We have started using the draft mode of GH PRs to make WIP status more clear.

merlinran:
Thanks @sanderpick ! btw I don't seem to have the permission to merge.

sanderpick:
Oop! I think you should have access now.

PR #516, opened 2021-05-15
Exchange edges improvements

mcrakhman:
In Anytype we discovered that current synchronisation mechanism of `exchangeEdges` doesn’t provide for cases when we have threads empty logs (i.e. their heads are empty). For example in these cases we exclude such thread from exchange process and thus prevent other node to discover new logs. 

To remedy that I created the same PR as we have in our fork. The idea is that we still exchange information about such threads and send default values for respective hashes. We also made things backward compatible (because essentially previously we also sent empty values as hashes when we responded to `exchangeEdges` request, because 0s are used by default).

### DIFF
--- a/broadcast/broadcast_test.go
+++ b/broadcast/broadcast_test.go
@@ -11,7 +11,7 @@ import (
 const (
 	N       = 3
 	testStr = "Test"
-	timeout = time.Second * 2
+	timeout = time.Second * 4
 )
 
 type ListenFunc func(int, *Broadcaster, *sync.WaitGroup)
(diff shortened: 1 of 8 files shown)


sanderpick (approved):
Nice, LGTM 💪

mcrakhman:
> Nice, LGTM 💪

Thanks, I see that the tests are failing. I ran them locally and they are fine, except that sometimes `TestListenerClose` test fails. I increased the timeout, can you please approve running the tests?

merlinran:
Okay CI passes. As the PR was approved, I'll take the liberty to merge it.

PR #521, opened 2021-05-30
Fix record app validation when processing chain

sanderpick:
Also allows DBs to be loaded in parallel when the manager starts. More comments inline.

### DIFF
--- a/db/manager.go
+++ b/db/manager.go
@@ -3,8 +3,10 @@ package db
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	ds "github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/keytransform"
@@ -16,6 +18,7 @@ import (
 	"github.com/textileio/go-threads/core/thread"
 	kt "github.com/textileio/go-threads/db/keytransform"
 	"github.com/textileio/go-threads/util"
+	"golang.org/x/sync/errgroup"
 )
 
 var (
@@ -23,6 +26,9 @@ var (
 	ErrDBNotFound = errors.New("db not found")
 	// ErrDBExists indicates that the specified db alrady exists in the manager.
 	ErrDBExists = errors.New("db already exists")
+
+	// MaxLoadConcurrency is the max number of dbs that will be concurrently loaded when the manager starts.
+	MaxLoadConcurrency = 100
 
 	dsManagerBaseKey = ds.NewKey("/manager")
 )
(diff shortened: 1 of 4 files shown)


db/manager.go
```diff
@@ -67,44 +73,69 @@ func NewManager(store kt.TxnDatastoreExtended, network app.Net, opts ...NewOptio
...
 	}
 	defer results.Close()
-	invalids := make(map[thread.ID]struct{})
+
+	log.Info("loading dbs")
+	eg, gctx := errgroup.WithContext(context.Background())
+	loaded := make(map[thread.ID]struct{})
+	lim := make(chan struct{}, MaxLoadConcurrency)
+	var lk sync.Mutex
+	var i int
```
sanderpick, line 82: A bit sloppy, but gets the job done.

net/net.go
```diff
@@ -1051,21 +1096,6 @@ func (n *net) loadRecords(
...
-			dbody, err := event.GetBody(ctx, n, readKey)
-			if err != nil {
-				return nil, head, err
-			}
-
-			if err = identity.UnmarshalBinary(r.PubKey()); err != nil {
-				return nil, head, err
-			}
-
-			if err = connector.ValidateNetRecordBody(ctx, dbody, identity); err != nil {
```
sanderpick, line 1064: @dgtony curious what you think about this problem and solution. We can't validate the record here because it might rely on older DB instances. E.g., if we have these events in the chain: `create` <- `save`, `save` write validation requires the instance to already be created.

net/net.go
```diff
@@ -913,8 +914,68 @@ func (n *net) putRecords(ctx context.Context, tid thread.ID, lid peer.ID, recs [
...
+			dbody, err := event.GetBody(ctx, n, readKey)
+			if err != nil {
+				return err
+			}
+
+			if err = identity.UnmarshalBinary(record.Value().PubKey()); err != nil {
+				return err
+			}
+
+			if err = connector.ValidateNetRecordBody(ctx, dbody, identity); err != nil {
```
sanderpick, line 957: Move validation to here, before setting head. If validation fails, we can un-store the internal blocks that were added during chain traversal.

merlinran: IIUC this change is to remove the invalid record but still return error, so the next call to it may succeed, but may still not if there are more records being invalid? 

In a general sense, this logic seems to be a bit unusual. Alternative ways could be either cleaning up all invalid records unconditionally and succeed, or returning a list of invalid records with errors so the caller can decide if it's okay to call some other method to "fix" them, but I could totally miss the context here.

merlinran (commented):
Done my first round. Mostly questions.

net/net.go
dgtony: As far as I understand the problem it may arise if your validation requires access to the previous state of the thread. I didn't actually realize that record validation can be "impure", i.e. depend on some information external to the record itself.
In this case there is no option besides validating along with applying. Failed record autoremoving, as Merlin pointed out isn't very clear, but imho it's fine. Record in the blockstore is in fact a part of internal state and may affect processing (e.g. chain loading), so user shouldn't directly manipulate it while system rollbacks to the clean state.

PR #522, opened 2021-05-30
db: manager race fix

jsign:
Just a v1.1.0-rc1 fix. Won't be merged to master.

### DIFF
--- a/db/manager.go
+++ b/db/manager.go
@@ -68,6 +68,7 @@ func NewManager(store kt.TxnDatastoreExtended, network app.Net, opts ...NewOptio
 	}
 	defer results.Close()
 	invalids := make(map[thread.ID]struct{})
+	var i int
 	for res := range results.Next() {
 		if res.Error != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


db/manager.go
```diff
@@ -115,9 +118,13 @@ func (m *Manager) GetToken(ctx context.Context, identity thread.Identity) (threa
 
 // NewDB creates a new db and prefixes its datastore with base key.
 func (m *Manager) NewDB(ctx context.Context, id thread.ID, opts ...NewManagedOption) (*DB, error) {
```
jsign, line 120: Going the route of tight locking instead of `defer` since the middle-body logic might be long-running and that would be too much locking.

db/manager.go
```diff
@@ -34,7 +35,9 @@ type Manager struct {
 
 	store   kt.TxnDatastoreExtended
 	network app.Net
-	dbs     map[thread.ID]*DB
+
+	lock sync.Mutex
```
merlinran, line 39: Maybe worth a `RWMutex` since most operations to `dbs` are reads?

jsign: Yeah, was mostly a quick mutex for the hotfix.
This should be done in some other PR on the week targeting `master`, not this `v1.1.0-rc1` tag anyway.

jsign: I won't merge this in master. I'll just use the commit hash to update the Hub to use this. (I'm not sure if `master` is really safe to use)

merlinran (approved):
LGTM. Great spot and fix!

merlinran:
Oh wait this PR is targeting `master`

jsign:
Thanks @merlinran . As said in the PR description, I'll leave this PR open and won't merge.
We should upstream this fix in `master` and maybe provide some `rc2`... for now, just using this PR commit hash to update `textile`.

PR #523, opened 2021-05-31
various fixes to the testground tests

merlinran:
Signed-off-by: Merlin Ran <merlinran@gmail.com>

### DIFF
--- a/integrationtests/testground/main.go
+++ b/integrationtests/testground/main.go
@@ -26,7 +26,6 @@ import (
 	ipldcbor "github.com/ipfs/go-ipld-cbor"
 	logging "github.com/ipfs/go-log/v2"
 	crypto "github.com/libp2p/go-libp2p-core/crypto"
-	"github.com/libp2p/go-libp2p-peer"
 	"github.com/multiformats/go-multiaddr"
 	"github.com/multiformats/go-multihash"
 	"github.com/testground/sdk-go/network"
@@ -35,7 +34,6 @@ import (
 	sync "github.com/testground/sdk-go/sync"
 	"google.golang.org/grpc"
 
-	"github.com/textileio/go-threads/cbor"
 	corenet "github.com/textileio/go-threads/core/net"
 	"github.com/textileio/go-threads/core/thread"
 	"github.com/textileio/go-threads/net/api"
@@ -262,19 +260,42 @@ func testRound(ctx context.Context, env *runtime.RunEnv, ic *run.InitContext, ro
 	msg("Created %d records", recordsToSend)
 	ic.SyncClient.MustPublishSubscribe(ctx, topic, thr.logHead, chCurrentHead)
 
-	logsGot := 0
+	expectedHeads := make(map[cid.Cid]bool)
+	for i := 0; i < env.TestInstanceCount; i++ {
+		expectedHeads[<-chCurrentHead] = true
+	}
+	var heads []cid.Cid
+	tk := time.NewTicker(10 * time.Millisecond)
+	defer tk.Stop()
+	for {
+		<-tk.C
+		var err error
+		heads, err = thr.GetHeads(ctx)
+		if err != nil {
+			return err
+		}
+		if len(heads) < len(expectedHeads) {
+			debug("expect %d logs, got %d, continuing", len(expectedHeads), len(heads))
+			continue
+		}
+		for _, head := range heads {
+			if !expectedHeads[head] {
+				debug("log not up-to-date, continuing")
+				continue
+			}
+		}
+		break
+	}
+
+	// now all logs are up-to-date, read records from all
 	totalRecords := 0
-	for head := range chCurrentHead {
+	for _, head := range heads {
 		records, err := thr.GetRecords(ctx, head)
 		if err != nil {
 			return fmt.Errorf("Error getting records: %w", err)
 		}
-		logsGot++
 		debug("the log with head %v has %d records", head, len(records))
 		totalRecords = totalRecords + len(records)
-		if logsGot >= env.TestInstanceCount {
-			break
-		}
 	}
 
 	env.R().RecordPoint("round-"+round+"-phase-2-elapsed-seconds", time.Since(start).Seconds())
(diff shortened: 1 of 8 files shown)


integrationtests/testground/main.go
```diff
@@ -109,14 +109,20 @@ func testSyncThreads(env *runtime.RunEnv, ic *run.InitContext) (err error) {
...
-				Enable:        true,
-				Default:       pair.shape,
-				CallbackState: sync.State("network-configured-" + round),
-				RoutingPolicy: network.AllowAll,
+			ic.NetClient.MustConfigureNetwork(ctx, &network.Config{
+				Network:        "default",
+				Enable:         true,
+				Default:        pair.shape,
+				CallbackState:  sync.State("network-configured-" + round),
+				CallbackTarget: 1,
```
merlinran, line 124: This was what caused the test never succeeded on docker/k8s. By default it waits for `env.TestInstanceCount` replies, but we only call this in one instance.

integrationtests/testground/main.go
```diff
@@ -198,39 +201,40 @@ func testRound(env *runtime.RunEnv, ic *run.InitContext, round string, desiredAd
...
+			ic.SyncClient.MustSignalAndWait(ctx, donePhase1, livePeers)
 		}()
 
-		for i, prev := 0, cid.Undef; i < myRecords; i++ {
-			rec, err := thr.AddRecord(ctx, prev)
-			if err != nil {
-				return err
-			}
-			prev = rec.Cid()
+		if err := thr.CreateRecords(ctx, recordsToSend); err != nil {
```
merlinran, line 214: moved the loop into `CreateRecords`, as it doesn't (incorrectly) manipulate `Prev` anymore.

integrationtests/testground/main.go
```diff
@@ -198,39 +201,40 @@ func testRound(env *runtime.RunEnv, ic *run.InitContext, round string, desiredAd
...
 		meLive = 0
-		myRecords = 0
+		recordsToSend = 0
 	}
-	livePeers = publishAndGetTotal(sync.NewTopic("livePeers-"+round+"-step-2", 0), meLive)
-	expectedRecords = publishAndGetTotal(sync.NewTopic("expectedRecords-"+round+"-step-2", 0), myRecords)
+	livePeers = publishAndGetTotal(sync.NewTopic("livePeers-"+round+"-phase-2", 0), meLive)
+	recordsToReceive = publishAndGetTotal(sync.NewTopic("recordsToReceive-"+round+"-phase-2", 0), recordsToSend) + recordsToReceive
+	chCurrentHead := make(chan cid.Cid, 1)
+	topic = sync.NewTopic("current-head-"+round, cid.Cid{})
```
merlinran, line 231: This is to publish the log head so other instances can `GetRecord` and traverse the whole log

integrationtests/testground/main.go
```diff
@@ -364,7 +375,11 @@ func createThread(ctx context.Context, cli *client.Client) (thr *threadWithKeys,
 	if err != nil {
 		return nil, err
 	}
-	return &threadWithKeys{info, identity, logSk, logPk, cli}, nil
+	ch, err := cli.Subscribe(context.Background(), corenet.WithSubFilter(info.ID))
```
merlinran, line 378: Subscribe early to avoid race condition

integrationtests/testground/main.go
```diff
@@ -403,49 +422,68 @@ func (t *threadWithKeys) Sharable() *SharedInfo {
...
-		return
-	}
-	debug("Subscribed to thread")
-	total := 0
-	for record := range ch {
-		total++
+	msg("Waiting for %d unique records", nRecords)
+	for record := range t.subscribeCh {
+		rec := record.Value()
+		if _, exists := t.seenRecords[rec.Cid()]; exists {
```
merlinran, line 428: This was meant to deduplicate records but there's no longer duplication with the fix

merlinran (commented):
@sanderpick this finally allows running tests successfully in docker and k8s.

integrationtests/testground/main.go
sanderpick: 👍

integrationtests/testground/main.go
sanderpick: Is this using the testground tooling to send records vs. letting the peers get and push records over the p2p lib?

integrationtests/testground/main.go
```diff
@@ -325,18 +334,22 @@ func startClient(desiredAddr string, env *runtime.RunEnv, ic *run.InitContext) (
 	}, nil
 }
 
-// info shared among peers via testground pubsub
+// SharedInfo includes info shared among peers via testground pubsub
```
sanderpick, line 337: Nice, so this is used by peers to join a thread created by some other peer.

integrationtests/testground/main.go
```diff
@@ -403,49 +424,68 @@ func (t *threadWithKeys) Sharable() *SharedInfo {
...
-	if err != nil {
-		return nil, err
-	}
-	logID, err := peer.IDFromPublicKey(t.logPk)
-	if err != nil {
-		return nil, err
+func (t *threadWithKeys) CreateRecords(ctx context.Context, num int) error {
+	for i := 0; i < num; i++ {
+		obj := make(map[string][]byte)
+		fuzzer.Fuzz(&obj)
```
sanderpick, line 448: Slick

integrationtests/testground/main.go
sanderpick: We could also use the raw `CreateRecord` API: https://github.com/textileio/go-threads/blob/master/core/net/net.go#L60 Or maybe that's a different/future test.

sanderpick (approved):
LGTM! I left a couple comments / questions.

integrationtests/testground/main.go
merlinran: Oh right. will do

integrationtests/testground/main.go
merlinran: It's just to publish the CIDs of the logs head so each peer can traverse. Is there another way to get them without subscribing?

sanderpick: When a peer writes to their log, it will push the new records to the other peers in thread using [this method](https://github.com/textileio/go-threads/blob/master/net/client.go#L251) on the p2p API. So, after thread setup, each peer should have the thread (either by creating it or adding it via your share thread method). After that point, the sharing should happen automatically.

merlinran: Yeah I get it but even though the records are pushed to other peers, how can the caller of go-threads on other peers access them without knowing the head CID?

sanderpick: Oh I see, so you want to be able to walk the records in the test to verify everything was received. 

You could do that with the `net/api` `Client` (which is the user API) using the `GetThread` [method](https://github.com/textileio/go-threads/blob/master/net/api/pb/threadsnet.proto#L128), which returns all thread info, including the head, and then use `GetRecord` [method](https://github.com/textileio/go-threads/blob/master/net/api/pb/threadsnet.proto#L134) to walk back with `Prev`.

merlinran: Aha that's much easier! I noticed the `Logs` field of `thread.Info` before but totally forgot that! Thanks!

merlinran: So I'd imagine code would typical either subscribe the thread to get updates, or periodically call `GetThread` to get a full picture at that point. But then the test code needs to wait long enough for the records to be actually synced to other peers. The wait time is tricky to decide. Maybe I need to keep `GetThread` and check if the heads of the logs are all up-to-date?

merlinran: This test does record the time taken to get all records, so I guess it doesn't hurt to do the periodical polling as you mentioned.

merlinran: Done https://github.com/textileio/go-threads/pull/523/commits/4c9ab518efccaccbdae274bb280f2ae7a9ba5cf8

integrationtests/testground/main.go
merlinran: Done https://github.com/textileio/go-threads/pull/523/commits/4c9ab518efccaccbdae274bb280f2ae7a9ba5cf8

PR #528, opened 2021-06-11
add basic resources usage measurements

merlinran:
This is the first step toward resources optimization. Upon successful running of the tests, the code prints nice line graphs like below to show resources usage, with the data points being sent to InfluxDB as well, if configured in Testground. 

image


integrationtests/testground/main.go
```diff
@@ -105,40 +110,56 @@ func testSyncThreads(env *runtime.RunEnv, ic *run.InitContext) (err error) {
...
-				CallbackTarget: 1,
-				RoutingPolicy:  network.AllowAll,
-			})
-			msg("Done configuring %s network", round)
-		}
+	successState := sync.State("testSyncThreads-success")
+	failState := sync.State("testSyncThreads-fail")
+	barrierSuccess := ic.SyncClient.MustBarrier(context.Background(), successState, env.TestInstanceCount)
+	barrierFail := ic.SyncClient.MustBarrier(context.Background(), failState, 1)
+	go func() {
```
merlinran, line 117: code is moved into goroutine unchanged. The purpose of this change is to either wait for all instances to succeed, so the test results are printed at the same place, or exit the whole test when any of the instance fails.

integrationtests/testground/main.go
```diff
@@ -105,40 +110,56 @@ func testSyncThreads(env *runtime.RunEnv, ic *run.InitContext) (err error) {
...
-		if err := testRound(ctx, env, ic, round, desiredAddr); err != nil {
-			msg("################### Peer #%d with %s network failed: %v ###################", ic.GlobalSeq, round, err)
+			desiredAddr := ""
+			if env.TestSidecar {
+				// listen on non-local interface when running in container, so the peers can communicate with each other
+				ip := ic.NetClient.MustGetDataNetworkIP().String()
+				desiredAddr = fmt.Sprintf("/ip4/%s/tcp/%d", ip, 3000+i)
+			}
+			if err = testRound(ctx, env, ic, round, desiredAddr); err != nil {
+				msg("################### Peer #%d with %s network failed: %v ###################", ic.GlobalSeq, round, err)
```
merlinran, line 145: actual changes resume here

merlinran:
Oh wait I think I should use https://github.com/google/cadvisor to monitor container resources usage. it looks like node_exporter measures at the host level.

merlinran:
Per https://www.bwplotka.dev/2019/golang-memory-monitoring/ at least memory usage collected at the container level can be more than the application actually uses. we can simply switch back to `runtime.ReadMemStats()`. But things become tricky on CPU and bandwidth...

integrationtests/testground/measurement.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	p.lastTransmit = total
+}
+
+func (p *measure) collectBytes(metrics []*dto.Metric) float64 {
+	var total, exclude float64
+	for _, m := range metrics {
+		total += *m.Counter.Value
+		for _, label := range m.Label {
+			if *label.Name == "device" && *label.Value == "lo0" {
```
merlinran, line 121: Only the network interfaces inside container are visible here, so unlike CPU and memory, network traffic can be measured accurately.

merlinran (commented):
okay @sanderpick this is the phased result before I change my focus. Ended up tracking number of goroutines, Go heap usage and bandwidth usage, which seem to be good factors to consider at this point.

sanderpick (approved):
Very nice @merlinran ! Love the graphs :) Let's checkin about next steps on this stuff tomorrow.

PR #529, opened 2021-06-16
README: fix NewClient() calls

bbigras:
Without this change I get:

```
panic: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: address /ip4/127.0.0.1/tcp/6006: missing port in address"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ As described in the [paper](https://docsend.com/view/gu3ywqi), ThreadDB's networ
 import "github.com/textileio/go-threads/api/client"
 ...
 
-db, err := client.NewClient("/ip4/127.0.0.1/tcp/6006", grpc.WithInsecure())
+db, err := client.NewClient("127.0.0.1:6006", grpc.WithInsecure())
 ```
 
 #### Getting a thread token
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -447,7 +447,7 @@ The full API spec is available [here](https://pkg.go.dev/github.com/textileio/go
 import "github.com/textileio/go-threads/net/api/client"
 ...
 
-net, err := client.NewClient("/ip4/127.0.0.1/tcp/6006", grpc.WithInsecure())
```
merlinran, line 450: Good catch. As the client communicates with regular gRPC server, TCP addr should be passed in instead of multiaddr. Correct me if I'm wrong @sanderpick

sanderpick: yep, correct! thanks @bbigras

sanderpick:
@bbigras can you please add a DCO to your commit? In case you're not familiar, here's a guide on how to do that: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md

bbigras:
> @bbigras can you please add a DCO to your commit? In case you're not familiar, here's a guide on how to do that: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md

Done

merlinran (approved):

Issue #530, opened 2021-06-20
Found an interesting bug related to synchronisation of threads


mcrakhman:
Lately we have been experiencing an interesting bug in go-threads.
The problem occurs when we have a lot of records and we are trying to synchornize them.

This happens as follows:
- Device A has thread with 0 records and it tries to get a lot of records from Device B regarding Log B (for example 1000)
- At the same time Device A requests Head of Log B through GetRecord effectively getting it through Bitswap. As a result of this the Head of Log B is added to Blockstore.
- When Device B returns 1000 records to Device A, Device A tries to put them using `putRecords`. So it checks them using `isKnown` which looks into blockstore and sees that we already have the Head of Log B (which we got through bitswap), so we decided not to proceed with saving the records.
- So the Head of Log B in Device A stays as `Undef` as we haven't updated it.
- So `ExchangeEdges` will see that we have different heads and will try to get records again. Thus we will have an infinite loop. 

I tried to reproduce this `isKnown` problem on testground:
- First instance creates a thread and adds 200 records
- Second instance starts after (so it doesn't receive anything through pubsub or push) and pulls the thread thus calling get records.
- At the same time Second instance tries to get the head through bitswap using GetRecord.
- Because Second instance receives head earlier than the batch of 200 records it will effectively skip adding them and thus will not update its head.

The test is not of super good quality, but at least it sometimes works :-) Feel free to ping me in case I missed something.

https://github.com/mcrakhman/go-threads/tree/sync-bug-demonstration

carsonfarmer:
Does this sound familiar, given your recent tests using test ground @merlinran ?

merlinran:
> At the same time Device A requests Head of Log B through GetRecord effectively getting it through Bitswap. As a result of this the Head of Log B is added to Blockstore.

IIUC what triggers this bug is that Device A gets the log head via a 3rd party channel and calls `GetRecord` before the records being synced (I tried the same when drafting the first Testground test, that's why `SharedInfo.LogHead` exists but was not used!). It's probably not an initially envisioned use case of go-threads, but definitely a legit one.

What if there's a new API call so Device A can explicitly tell go-threads that this is the log head so the data structure gets set properly? Far from an elegant solution but it introduces no performance penalty.

mcrakhman:
> What if there's a new API call so Device A can explicitly tell go-threads that this is the log head so the data structure gets set properly? Far from an elegant solution but it introduces no performance penalty.

Can you elaborate please?

[...]

PR #531, opened 2021-06-23
Add head counters 

mcrakhman:
for the context: https://github.com/textileio/go-threads/issues/530


integrationtests/migration/Makefile
```diff
@@ -0,0 +1,12 @@
+.EXPORT_ALL_VARIABLES:
+
+RUN_MIGRATION_TEST="true"
+ENV_FILENAME="thread_ids"
+ENV_NUM_RECORDS="10"
+ENV_NUM_THREADS="10"
+
+test-migration:
```
mcrakhman, line 8: So I've found out that implementing the migration test in testground is actually not convenient. Because you need to have:
- two different docker containers versions pointing to two different repositories
- they should have the same storage (because how would the other one migrate on the results of the first one)
- we need to also note that the head structure is different, so these containers should not use the same test (see `GetHead` method in test or we should create some helper methods outside the test) which is not as easy as it seems

So instead I decided to go with the quick and dirty solution of just generating the records with the old version and the having the test using the other version which seems to work pretty well.

integrationtests/testground/go.mod
```diff
@@ -10,13 +10,15 @@ require (
...
+	github.com/prometheus/client_golang v1.10.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/node_exporter v1.1.2
 	github.com/testground/sdk-go v0.2.7
 	github.com/textileio/go-threads v1.1.0-rc1.0.20210609142634-18b4ea73088f
 	google.golang.org/grpc v1.37.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 )
+
+replace github.com/textileio/go-threads v1.1.0-rc1.0.20210609142634-18b4ea73088f => github.com/mcrakhman/go-threads v1.1.0-rc1.0.20210626214637-1a88e4ef1d3b
```
mcrakhman, line 24: This should be change definitely, we can also get rid of replace, but still to do this we need to have a merged commit in upstream `go-threads` repo. Not sure how to do this right now. Maybe have a follow up PR after we merge this one?

integrationtests/migration/Makefile
merlinran: makes total sense esp. the latter two points, and this test is much easier to run than having to spin up testground

net/record.go
```diff
@@ -38,22 +43,49 @@ func (r *recordCollector) Store(lid peer.ID, rec core.Record) {
...
 
+func (r *recordCollector) UpdateHeadCounter(lid peer.ID, counter int64) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	// we update the counter only if we have some records
+	val, found := r.counters[lid]
+	if !found {
+		r.counters[lid] = counter
+		// if a peer does not support the new logic we cannot rely on counter comparison
```
merlinran, line 54: uber nit: how about moving the comment into the block it refers to? This line is for the `if` below, but looks more closely to the code above it.

net/net.go
```diff
@@ -954,22 +1040,26 @@ func (n *net) loadRecords(
...
-	head, err := n.currentHead(tid, lid)
-	if err != nil {
-		return nil, head, err
+	if counter == thread.CounterUndef {
+		if exist, err := n.isKnown(last.Cid()); err != nil {
+			return nil, thread.HeadUndef, err
+		} else if exist || !last.Cid().Defined() {
+			return nil, thread.HeadUndef, nil
+		}
+	} else if counter <= head.Counter {
```
merlinran, line 1061: nice

merlinran (commented):
Still need to check more details in `net` package. here are some nits so far.

net/server.go
```diff
@@ -198,24 +198,31 @@ func (s *server) GetRecords(ctx context.Context, req *pb.GetRecordsRequest) (*pb
...
 			offset = opts.Offset.Cid
+			counter = opts.Counter
 			limit = minInt(int(opts.Limit), logRecordLimit)
 		} else {
 			offset = cid.Undef
 			limit = logRecordLimit
-			pblg = logToProto(lg)
+			counter = thread.CounterUndef
 		}
+		pblg = logToProto(lg)
```
merlinran, line 215: uber nit: since it's unconditionally initialized, it can be put into the `var()` section above, or to save some CPU cycles, don't create the var until L245, and get the counter from `lg.Head.Counter` in L221.

merlinran (approved):
LGTM. Neat stuff! It would be great if @carsonfarmer can have a look, as I may have missed some key invariants or constraints while focusing on the code level.

net/net.go
```diff
@@ -954,22 +1040,27 @@ func (n *net) loadRecords(
...
-	if exist, err := n.isKnown(last.Cid()); err != nil {
-		return nil, cid.Undef, err
-	} else if exist || !last.Cid().Defined() {
-		return nil, cid.Undef, nil
-	}
-
-	head, err := n.currentHead(tid, lid)
-	if err != nil {
-		return nil, head, err
+	// if we don't have the counter (but have some recs) then we were communicating with old version peer
```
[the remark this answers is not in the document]
mcrakhman: As far as I understand in situation with `AddRecord` we should basically use old checks for `isKnown`, because it is not possible for us to know the counter of record. This is not ideal of course, but we will not be able to fix it, unless the counter would be a part of record.

net/net.go
```diff
@@ -174,6 +179,73 @@ func NewNetwork(
...
+	return counter, nil
+}
+
+func (n *net) migrateHeadsIfNeeded(ctx context.Context, ls lstore.Logstore) (err error) {
+	threadIds, err := ls.Threads()
+	if err != nil {
+		return err
+	}
+
+	log.Info("Starting migrating heads")
```
carsonfarmer, line 209: nit: this is going to log "Starting migration heads" even if we don't end up doing a migration, which might be confusing? What about just a) changing this to "Checking for head migrations" / "Finished checking for head migrations" or b) using a flag outside the loop to check if any migrations are required, and only lazily logging if true?

net/record.go
```diff
@@ -38,22 +43,49 @@ func (r *recordCollector) Store(lid peer.ID, rec core.Record) {
...
 	for id, seq := range r.rs {
 		ordered, ok := seq.List()
 		if !ok {
 			return nil, fmt.Errorf("disjoint record sequence in log %s", id)
 		}
+
+		counter, found := r.counters[id]
+		// this should never happen because we do this for every log
+		if !found {
+			return nil, fmt.Errorf("did not found log counter in log %s", id)
```
carsonfarmer, line 78:
```suggestion
			return nil, fmt.Errorf("did not find log counter in log %s", id)
```

carsonfarmer (approved):
OK, LGTM as well. I have made a few _minor_ suggestions and nits. Nothing blocking. But would be nice to have those addressed. Please also make sure you've dealt with all of @merlinran's suggestions (it looks like you have, maybe one left?).

Note also that we will require a follow up PR to address: https://github.com/textileio/go-threads/pull/531/files#r659274884

net/api/pb/Makefile
```diff
@@ -7,7 +7,7 @@ all: $(GO)
 %.pb.go: %.proto
 	protoc -I=. \
 	--plugin="protoc-gen-ts=${PROTOC_GEN_TS_PATH}" --js_out="import_style=commonjs,binary:javascript/." --ts_out="service=grpc-web:javascript/." \
-	--go_out=\
+	--gogofaster_out=\
```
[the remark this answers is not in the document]
sanderpick: We had been using gogo protos for node-to-node communication, and normal protos for APIs that are open to different language bindings, e.b., JS, python, etc. I'll take a look at the setup (it's been awhile), maybe gogo is fine here.

PR #536, opened 2021-07-04
Fix nil log crash in get records

mcrakhman:
Signed-off-by: mcrakhman <rakhmanov.m@gmail.com>


merlinran (approved):
good catch @mcrakhman !

PR #541, opened 2021-08-04
net: revert gogo usage in api

sanderpick:
To avoid potential compatablity problems with our JS clients, this revert the change to `gogo` protos in the `net/api` layer. I'll do a follow up PR after getting other things sorted out that switches to the bingo+buf style proto generation for the public APIs.


merlinran (approved):
Cool that you can generate the go file.

PR #542, opened 2021-08-04
Clean up dangling PRs

sanderpick:
- Pulls in @jsign's race fix
- Pulls in the concurrent DB loading changes
- Pulls in the record validation fix
- Updates dependencies


net/net.go
```diff
@@ -997,10 +998,70 @@ func (n *net) putRecords(ctx context.Context, tid thread.ID, lid peer.ID, recs [
...
+				body, err := event.GetBody(ctx, n, nil)
+				if err != nil {
+					return err
+				}
+
+				if err := n.RemoveMany(ctx, []cid.Cid{event.Cid(), header.Cid(), body.Cid()}); err != nil {
+					return fmt.Errorf("removing invalid blocks: %w", err)
+				}
+
+				return userErr
```
merlinran, line 1061: See my comment at https://github.com/textileio/go-threads/pull/521/files#r642159397 maybe we can continue here instead of returning error? the caller can do nothing about the error apart from knowing the block being removed.

merlinran (commented):
Left a couple of comments but the only one I want to discuss is the error return after removing invalid blocks.

db/manager.go
```diff
@@ -115,9 +148,13 @@ func (m *Manager) GetToken(ctx context.Context, identity thread.Identity) (threa
 
 // NewDB creates a new db and prefixes its datastore with base key.
 func (m *Manager) NewDB(ctx context.Context, id thread.ID, opts ...NewManagedOption) (*DB, error) {
+	m.lk.RLock()
 	if _, ok := m.dbs[id]; ok {
```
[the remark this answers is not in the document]
sanderpick: nice, changed

net/net.go
sanderpick: Oop, sorry... I forgot about that comment. Copying it here:

> IIUC this change is to remove the invalid record but still return error, so the next call to it may succeed, but may still not if there are more records being invalid?

> In a general sense, this logic seems to be a bit unusual. Alternative ways could be either cleaning up all invalid records unconditionally and succeed, or returning a list of invalid records with errors so the caller can decide if it's okay to call some other method to "fix" them, but I could totally miss the context here.

If we continue instead of erroring, then we may end up with gaps, since a later record might pass the app-specific validation check. The logic here will cleanup the invalid record, and return an error to the caller that comes from the app-specific validation check, e.g., "write validation failed: only dogs are allowed, not cats". 

I could still be missing your point though!

api/pb/javascript/package.json
```diff
@@ -15,11 +15,11 @@
     "threads_pb_service.d.ts"
```
carsonfarmer, line 15: Ah right, so this is going to kick off a new javascript client package. We'll have to make sure we update downstream JS clients as well.

carsonfarmer (approved):
👍

net/net.go
merlinran: Ah okay so this `RemoveMany` is simply to clear the effect of adding this record which failed the app-specific validation check? Makes sense.

merlinran (approved):

net/net.go
sanderpick: Yep, that will remove all the blocks that are making up the single record.

api/pb/javascript/package.json
sanderpick: yep, i have one more change coming (different PR) and then will do some testing with `textile`. if all goes well, we can do a release and that will kick off these low-level js libs.

PR #546, opened 2021-08-05
Control thread pulling

sanderpick:
- Moves thread pulling control to config vars and flags.
- Makes logging level `INFO` by default instead of just `ERROR`.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -99,13 +99,13 @@ func DefaultNetwork(opts ...NetOption) (NetBoostrapper, error) {
 
 	// Build a network
 	api, err := net.NewNetwork(ctx, h, lite.BlockStore(), lite, tstore, net.Config{
-		PullThreadsDisabled:        config.PullThreadsDisabled,
-		PullThreadsLimit:           config.PullThreadsLimit,
-		PullThreadsStartAfter:      config.PullThreadsStartAfter,
-		PullThreadsInitialInterval: config.PullThreadsInitialInterval,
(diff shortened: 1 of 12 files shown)


common/common.go
```diff
@@ -329,6 +347,17 @@ func WithNetLogstore(lt LogstoreType) NetOption {
 	}
 }
 
+func WithNetPulling(disabled bool, threadLimit int, startAfter, initialInterval, interval time.Duration) NetOption {
```
merlinran, line 350: It's a little surprising that `WithNetPulling(true, ...)` actually disables net pulling. I'd even vote for removing the `disabled` parameter entirely, so calling this simply enables pulling with the params.

net/net.go
```diff
@@ -114,8 +104,29 @@ type net struct {
 
 // Config is used to specify thread instance options.
 type Config struct {
-	Debug  bool
-	PubSub bool
+	PullThreadsDisabled        bool
+	PullThreadsLimit           int
```
merlinran, line 108: Maybe make this `uint` to signal that it shouldn't be negative?

common/common.go
merlinran: If pulling should be on by default, we could provide another utility called `NoNetPulling()` to disable it.

merlinran (approved):
LGTM with just some personal opinions

common/common.go
sanderpick: Yeah, that is pretty weird now that you mention it :) I'll go with the `NoNetPulling` option.

PR #548, opened 2021-08-09
pubsub: migrate to rpc lib

sanderpick:
- Migrates pubsub handling to the rpc lib
- Fixes test deadlocking
- Updates JS libs


jsonpatcher/jsonpatcher.go
```diff
@@ -267,6 +267,12 @@ func (je patchEvent) time() (t time.Time) {
 		t = ts
 	case int:
 		t = time.Unix(0, int64(ts))
+	case uint:
+		t = time.Unix(0, int64(ts))
+	case int64:
```
sanderpick, line 272: cbor now does not map int64 to int, meaning we need another case here

net/client.go
```diff
@@ -304,10 +304,8 @@ func (s *server) pushRecord(ctx context.Context, tid thread.ID, lid peer.ID, rec
 	}
 
 	// Finally, publish to the thread's topic
-	if s.ps != nil {
-		if err = s.ps.Publish(ctx, tid, req); err != nil {
-			log.Errorf("error publishing record: %s", err)
-		}
+	if err = s.publishRecord(ctx, tid, req); err != nil {
```
sanderpick, line 307: `s.publishRecord` checks for `s.ps` being nil.

net/client.go
```diff
@@ -454,7 +452,7 @@ func (s *server) dial(peerID peer.ID) (pb.ServiceClient, error) {
 	conn, ok := s.conns[peerID]
 	if ok {
 		if conn.GetState() == connectivity.Shutdown {
-			if err := conn.Close(); err != nil {
+			if err := conn.Close(); err != nil && status.Code(err) != codes.Canceled {
```
sanderpick, line 455: no reason to log this error when it's expected

net/net.go
```diff
@@ -279,7 +284,7 @@ func (n *net) Close() (err error) {
 			log.Errorf("error closing connection: %v", err)
 		}
 	}
-	n.rpc.GracefulStop()
+	tu.StopGRPCServer(n.rpc)
```
sanderpick, line 287: I added a util that kills the server as some time instead of just waiting for it to gracefully close, which can never happen. The tests were hanging once in awhile, I think it was due to this.

net/net.go
```diff
@@ -635,8 +634,6 @@ func (n *net) AddReplicator(
 		dialable, err = getDialable(paddr)
 		if err == nil {
 			n.host.Peerstore().AddAddr(pid, dialable, pstore.PermanentAddrTTL)
-		} else {
-			log.Warnf("peer %s address requires a DHT lookup", pid)
```
sanderpick, line 639: This error can be misleading since the peer will only need a DHT lookup if the user supplied address doesn't include an IP/DNS portion _and_ there are no addresses in the peerstore. We're not checking the latter here.

net/server.go
```diff
@@ -30,19 +33,23 @@ var (
...
 // server implements the net gRPC server.
 type server struct {
-	sync.Mutex
 	net   *net
-	ps    *PubSub
 	opts  []grpc.DialOption
 	conns map[peer.ID]*grpc.ClientConn
+
+	ps     *pubsub.PubSub
+	topics map[thread.ID]*rpc.Topic
```
sanderpick, line 41: Here are some mechanics for managing topics, which we need to do here now. This functionality used to be in a local `pubsub` package.

net/server.go
```diff
@@ -54,22 +61,23 @@ func newServer(n *net, opts ...grpc.DialOption) (*server, error) {
 	s.opts = append(defaultOpts, opts...)
 
 	if n.conf.PubSub {
-		ps, err := pubsub.NewGossipSub(
+		var err error
+		s.ps, err = pubsub.NewGossipSub(
 			n.ctx,
 			n.host,
-			pubsub.WithMessageSigning(false),
-			pubsub.WithStrictSignatureVerification(false))
```
sanderpick, line 61: These are deprecated. Also, we really do want message signing here because we're no longer doing it at the proto level.

net/server.go
```diff
@@ -426,6 +419,116 @@ func (s *server) localEdges(tid thread.ID) (addrsEdge, headsEdge uint64, err err
...
+	if !ok {
+		return fmt.Errorf("publish to unknown thread %s", topic)
+	}
+
+	data, err := req.Marshal()
+	if err != nil {
+		return fmt.Errorf("marshaling pubsub message: %v", err)
+	}
+
+	if _, err := t.Publish(ctx, data, rpc.WithIgnoreResponse(true)); err != nil {
```
merlinran, line 490: With v0.0.4 of the rpc lib, each `Publish` call spawns a goroutine which keeps waiting for response until the context expires, even with this ignore response flag. It can cause too many goroutines if the timeout is long enough and `publishRecord` being called often. That might be an oversight, but depends on the answer to the question: if the caller doesn't expect a response, should `Publish` deliver the message to whichever peers online at that moment, or should it wait and re-publish to new peers until the context ends (same as the case of `WithMultiResponse`)?

util/util.go
```diff
@@ -251,3 +253,18 @@ func ComputeAddrsEdge(as []PeerAddr) uint64 {
...
+func StopGRPCServer(server *grpc.Server) {
+	stopped := make(chan struct{})
+	go func() {
+		server.GracefulStop()
+		close(stopped)
+	}()
+	timer := time.NewTimer(10 * time.Second)
+	select {
+	case <-timer.C:
+		server.Stop()
```
merlinran, line 266: Maybe worth a warning message?

merlinran (approved):
LGTM with one concern that the current rpc lib behaviour may cause too many goroutines.

net/server.go
merlinran: And we can change this behavior in the rpc lib pretty easily.

sanderpick: hmm, good point! what do you think about an option in the rpc lib that is all about republishing: `rpc.WithRepublishing(true)` ?

util/util.go
sanderpick: nice, will do

PR #569, opened 2021-12-19
Fix links in README

sbs2001:
Due to missing scheme in URLs Github seems to interpret them as relative paths. 

Previously the `DB client ` link was pointing to https://github.com/textileio/go-threads/blob/master/github.com/textileio/go-threads/api/client and the ` Network client` was pointing to https://github.com/textileio/go-threads/blob/master/github.com/textileio/go-threads/net/api/client . 

Both of these are now pointed to their pkg go docs.


sanderpick (approved):
Great, thanks!

PR #574, opened 2022-01-07
Announce Address - allows running ThreadsDB behind a NAT

vkost:
This PR lets users assign an AnnounceAddress to libp2p. After you assign the address, you can run ThreadDB behind a NAT, or inside a docker container, and the thread addresses will contain the public IP instead of the local one.

### DIFF
--- a/threadsd/main.go
+++ b/threadsd/main.go
@@ -139,7 +139,7 @@ func main() {
 	} else {
 		opts = append(opts, common.WithNetBadgerPersistence(*repo))
 	}
-	if announceAddrStr != nil {
+	if announceAddr != nil {
 		opts = append(opts, common.WithAnnounceAddr(announceAddr))
 	}
 	n, err := common.DefaultNetwork(opts...)
--- a/threadsd/main.go
+++ b/threadsd/main.go
@@ -34,6 +34,7 @@ func main() {
 
 	repo := fs.String("repo", ".threads", "Repo location")
 	hostAddrStr := fs.String("hostAddr", "/ip4/0.0.0.0/tcp/4006", "Libp2p host bind address")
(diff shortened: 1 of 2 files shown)


threadsd/main.go
```diff
@@ -128,6 +139,9 @@ func main() {
 	} else {
 		opts = append(opts, common.WithNetBadgerPersistence(*repo))
 	}
+	if announceAddrStr != nil {
```
sanderpick, line 142: should be `if announceAddr != nil {` ?

sanderpick (commented):
Looks good! Just one comment

threadsd/main.go
vkost: Thank you - I missed it. GoLand probably hinted the var and I accepted it without paying attention.
I fixed the error!

sanderpick:
Thanks! Looks like you need to modify your commits to include a DCO: https://github.com/textileio/go-threads/pull/574/checks?check_run_id=4763261922

sanderpick (approved):
no worries at all. it's kind of a weird requirement IMO. running tests now and will merge

omar391:
@sanderpick One question: How do I create my "Announce Address"? Are there some public signaling nodes? Any reference to the docs will be very appreciated! Thanks in advance!

vkost:
Hi @omar391 the purpose of the AnnounceAddress parameter is to fix the problem with running libp2p within a docker container. If you run it on a host directly, you don't have to set it. If you run it within a container, it would normally add the docker internal network address to the peer addr, since that's all that executable sees. In this case, you'd want to set the AnnounceAddress to the public IP of the host running docker, and you'll get it to work as expected. 
The parameter should be a multiaddr, like this: 
/ip4/<public_ip_of_the_host>/tcp/<port_on_host>

Vic

omar391:
@vkost Thanks for clarifying it!

PR #578, opened 2022-03-29
Fix panic in ensureUniqueLog - GetFirstPrivKeyLog() can return null s…

vkost:
…o the test for PrivKey afterwards can panic

Signed-off-by: vkost <viktor.kostadinov@gmail.com>


sanderpick (approved):
Thank you ser!

